PR #250, opened 2022-03-31
Fix/cancelled test jobs

nodlesh:
The test run in GHAs in the wallet pipelines have changed. Two things are causing these changes; Our Sauce Labs plan only support one execution at a time, and we are getting cancelled test jobs because of either time or resources.
Changes will include:

- If either of the builds (iOS or Android) fails, no tests will be executed.
- Test runs will run in a matrix, and will run in series not in parallel
- Test Scope will change to run a minimal set of tests, right now it will be one test scenario that goes through the happy path from new user, to getting a cred, then to proof. I think this is fine since we only want smoke tests in this pipeline anyway.


jleach (approved):

jleach:
@nodlesh Not required, just a Pro Tip: You can rebase and "squash" (combine several commits into one) to get rid of some of the "again again again" message (I do it all the time like you did). You can `git log` and find the commit ID just before the ones you want to squash. Then `git rebase -i COMMIT_ID`. They're squashed bottom to top so don't do anything to the one on top. Others just replace `pick` with `fixup` to roll 'em up into the previous. Ones with `pick` will be untouched.

PR #258, opened 2022-04-07
Update/add proof tests ci

nodlesh:
This PR updates the running of the integration smoke tests in the CI with added proof tests. The Scenario added goes through the whole app from new user to getting a proof verified.


nodlesh:
@jleach Do you know why I always get every commit I've ever done in commits for this PR? I just created a new branch changed one thing and still seeing all those commits.

jleach:
Are the other commits merged into `main`? My though is that we see them above because GitHub thinks they are unique to this branch (not on main). @nodlesh

PR #259, opened 2022-04-08
Bifold framework

cvarjao:
This PR follows the refactoring of the upstream project ( [PR-274](https://github.com/hyperledger/aries-mobile-agent-react-native/pull/274) ) which now enables bc wallet to extend the upstream project as opposed to patch it


jleach (approved):
Ran locally, builds fine. Workflow ran as expected also. LGTM.

PR #383, opened 2022-06-01
Open file limit to 1024

jleach:
Increase the open file limit to 1024 to match settings in Bifold repo.


jleach:
Already reviewed and merged into Bifold so I'll do the same here. Merging...

PR #405, opened 2022-06-23
Add AuthProvider

jleach:
Add the AuthProvider to our App so we can use the new uber secure wallet mechanics. Looks like our Prettier or lint rules may be different so it updated `'` to be `"` and added `;`. I'd change it but will do a different PR to align our rules with Bifold for simplicity.

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -1,6 +1,7 @@
 import {
   Agent,
   AgentProvider,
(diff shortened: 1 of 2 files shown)


amanji (approved):
The lint rules are applied across the repo? Otherwise LGTM!

PR #492, opened 2022-08-19
feat: credential branding

amanji:


wadeking98 (approved):
LGTM

swcurran:
Shouldn’t the template code be in Aries Bifold?  I would think this should be a common feature —especially if/when we OCA enable it.

swcurran:
Obviously anything hardcoded about specific BC credentials shouldn’t be in Bifold.

cvarjao:
@swcurran , there are 2 PRs:
hyperledger/aries-mobile-agent-react-native#413 (in Bifold) and this one.
The Bifold one has the underline template/rendering, this one has the resources of a couple of credentials we are branding.

swcurran:
Good stuff, thanks. I looked on my phone so didn’t really look too closely at the code. Clearly. 🤦

PR #505, opened 2022-08-25
feat: deep linking support

amanji:
* Updates git submodule for Bifold
* Update credential branding schema ID for demo student ID


swcurran:
I haven't reviewed the code, but wanted to ask if we are sure the code here is in the right repo. I would think the capability to use deep-links should be in aries-bifold, while the the use of the feature (e.g. bcgov://, etc.) should be in BC Wallet and other derivatives.  

Thanks!

cvarjao:
> I haven't reviewed the code, but wanted to ask if we are sure the code here is in the right repo. I would think the capability to use deep-links should be in aries-bifold, while the the use of the feature (e.g. bcgov://, etc.) should be in BC Wallet and other derivatives.
> 
> Thanks!

@swcurran , it is a bit of both. The boiler plate projects (iOS/Android) needs to also be updated. The code itself will point to core code running from bifold and configurable to each wallet.

amanji:
@swcurran I had thought about this as well. We can support deep links directly in Bifold, the question is which ones? `didcomm` for sure, however BC Wallet also needs to support the `bcwallet` scheme. These need to be added directly to the iOS AppDelegate and AndroidManifest files which are present in both BC Wallet Mobile and Bifold. I opted to leave the schemes out of Bifold pending a larger discussion with the Bifold Community. For now this approach enables deep linking in BC Wallet specifically.

I should also. point out that the code to handle deep links is present in Bifold through hooks. If the schemes are configured, Bifold will attempt to process the deep link, otherwise it silently continues.

cvarjao (approved):

wadeking98 (approved):

PR #511, opened 2022-08-26
fix: debug android manifest to work with metro

wadeking98:
Changed `android:usesCleartextTraffic` to `true` in the manifest. It seems like metro needs the attribute to be set to true. This change only affects the debug/development environment

### DIFF
--- a/app/android/app/src/debug/AndroidManifest.xml
+++ b/app/android/app/src/debug/AndroidManifest.xml
@@ -5,9 +5,10 @@
     <uses-permission android:name="android.permission.SYSTEM_ALERT_WINDOW"/>
 
     <application
-        android:usesCleartextTraffic="false"
+        android:usesCleartextTraffic="true"
         tools:targetApi="28"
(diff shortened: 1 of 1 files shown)


amanji (commented):
There's another one that I had set to false. I changed it because Sonar was complaining. Maybe let's change them both back to the default?

cvarjao:
> There's another one that I had set to false. I changed it because Sonar was complaining. Maybe let's change them both back to the default?

@amanji , it is recommended that in the release build of the app that flag stays off. So, if we can keep it on just for debug/metro purposes that would be great

amanji:
> > There's another one that I had set to false. I changed it because Sonar was complaining. Maybe let's change them both back to the default?
> 
> @amanji , it is recommended that in the release build of the app that flag stays off. So, if we can keep it on just for debug/metro purposes that would be great

If it was set back to false because the deep linking wasn't working id confirm with a release build that we don't run into the same issue. Otherwise I'd take the recommendation to turn it off.

PR #512, opened 2022-08-26
refactor: adds watermarked background for lsbc demo

amanji:


app/src/assets/branding/credential-branding.ts
```diff
@@ -1,7 +1,7 @@
 import { Overlay } from 'aries-bifold/App/types/overlay'
 
 const memberCardOverlay = {
-  imageSource: require('./lsbc-member-card.png'),
```
cvarjao, line 4: The PROD credential needs to NOT have the demo background. We may need 2 `memberCardOverlay` here

cvarjao (approved):

PR #531, opened 2022-08-30
feat: override biometrics screen

jleach:
Override the biometrics screen to match BC designs, update the keychain package, and alight test to match BC.


amanji (approved):
LGTM. Might want to update with main as it looks like branding is not present.

PR #577, opened 2022-09-15
feat: OCA for the translation of labels

jcdrouin21:
## Multilingual display of credential attributes labels with Overlay Capture Architecture (OCA)

Integration of the overlay Capture Architecture (OCA) to allow the translation of the attributes of a credential at the time of display. The oca.js-form-core[https://github.com/THCLab/oca.js-form-core] and oca.js libraries are used to create the OCA structure coming from an OCA bundle (hardcoded at this time). The code of these libraries is imported by copy/paste in OCA-package because the installation via npm does not work (these libraries are not compatible with a React Native project). This code will preferably change for the npm packages as soon as we are able to use them so it does not need an extensive review.

Other PRs will follow to add attribute value formatting as well as to retrieve an OCA bundle associated with the cred_def_id or schema_id of a credential.


jleach:
@amanji Able to check this one out?

jcdrouin21:
We just open the PR in the Bifold. You now have all the necessary part to test the PR. You also can change the oca.json file in the `assets` to use your own OCA bundle corresponding to your credential schema.

amanji:
Im curious whether we can contribute the core OCA code to a repository and add it as a package?

swcurran:
It should go in the "right place".  There have been questions about where that is.  At minimum, it's Bifold -- this isn't BC specific.  But I don't know the technical answer of whether it should AFJ-ext, AFJ, or in an OCA repository.  The scope if interest is at least all of Aries, so all derivative implementations of Aries Bifold will use this.  It could be that anyone using OCA in react-native can use what we are building, so it could go beyond Aries and into OCA Core.

amanji:
For the purpose of testing this are we ok with temporarily housing this in BC Wallet until a permanent location has been decided upon? I just worry that it won't get moved once incorporated.

jleach:
Closing. I think we all collectively decided to go with the implementation in #646.

PR #586, opened 2022-09-19
feat: Automatically handle credential offers from the BCSC / IDIM agent.

jleach:
This PR adds the ability to detect credential offers from the BCSC IDIM agent and automatically handle them. This helps to automate the entire workflow so the use has less to do.


jleach:
Closing. Adding as part of another PR.

PR #592, opened 2022-09-21
feat: use secure browser tab for auth

### LINKED ISSUE #585: Use a secure browser tab for integrating with BCSC web components
When opening BCSC Verified Person credential workflow the BC Wallet uses a WebView. We would like to use a secure browser tab to for a more secure experience, and so that history can be preserved in the local device browser.

### LINKED ISSUE #587: Handle custom URLs so they trigger specific application behaviour
The application should be able to deal with specific URLs in a way where custom behaviour can be associated to them when `window.location` is set to one of them.

- [ ] bcwallet://bcsc/v1/dids/<wallet-did>/success
- [ ] bcwallet://bcsc/v1/dids/<wallet-did>/cancel

jleach:
This PR adds support for a secure authentication tab on iOS and Android API => 31. For Android API 30 or lower it will trigger the built in Web browser to open the URL, once authentication is completed it will hand back to the BC Wallet. The API updates are not common to Bifold so no changes in support levels there.

- feat: use secure browser tab for authentication with BCSC.
- feat: handle callback (redirect) URLs from BCSC.
- chore: update Android SDK to API 31 (was 30).
- chore: updated cicd pipeline to be more simple for Android builds and use API 31.

Fixed #585
Fixed #587

### DIFF
--- a/app/src/components/BCIDView.tsx
+++ b/app/src/components/BCIDView.tsx
@@ -19,17 +19,34 @@ import {
   testIdWithKey,
   HomeContentView,
   BifoldError,
+  Agent,
 } from "aries-bifold";
 import { IDIM_AGENT_INVITE_URL, IDIM_AGENT_INVITE_ID } from "../constants";
 import { useNavigation } from "@react-navigation/core";
 import { Screens } from "aries-bifold";
 import { Config } from "react-native-config";
+import { InAppBrowser, RedirectResult } from "react-native-inappbrowser-reborn";
 
 const legacyDidKey = "_internal/legacyDid"; // TODO:(jl) Waiting for AFJ export of this.
 const trustedInvitationIssueRe =
   /3Lbd5wSSSBv1xtjwsQ36sj:[0-9]{1,1}:CL:[0-9]{5,}:default/i;
 const trustedFoundationCredentialIssuerRe =
   /7xjfawcnyTUcduWVysLww5:[0-9]{1,1}:CL:[0-9]{5,}:Person\s\(SIT\)/i;
+const redirectUrlTemplate = "bcwallet://ServiceCard/v1/dids/<did>";
+
+enum AuthenticationResultType {
+  Success = "success",
+  Fail = "fail",
+  Cancel = "cancel",
+}
+
+enum ErrorCodes {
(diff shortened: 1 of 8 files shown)


app/src/components/BCIDView.tsx
```diff
@@ -19,17 +19,34 @@ import {
...
 import { useNavigation } from "@react-navigation/core";
 import { Screens } from "aries-bifold";
 import { Config } from "react-native-config";
+import { InAppBrowser, RedirectResult } from "react-native-inappbrowser-reborn";
 
 const legacyDidKey = "_internal/legacyDid"; // TODO:(jl) Waiting for AFJ export of this.
 const trustedInvitationIssueRe =
   /3Lbd5wSSSBv1xtjwsQ36sj:[0-9]{1,1}:CL:[0-9]{5,}:default/i;
 const trustedFoundationCredentialIssuerRe =
   /7xjfawcnyTUcduWVysLww5:[0-9]{1,1}:CL:[0-9]{5,}:Person\s\(SIT\)/i;
```
wadeking98: Should probably make these regexes match from the start of the string. I could create a cred def with the tag that matches the regex and then bc wallet would think it's coming from the verified person use case

app/src/components/BCIDView.tsx
```diff
@@ -111,8 +146,82 @@ const BCIDView: React.FC = () => {
...
+    if (navigation.canGoBack()) {
+      navigation.goBack();
+    }
+  };
+
+  const authenticateWithServiceCard = async (did: string): Promise<void> => {
+    try {
+      const url = `${Config.IDIM_PORTAL_URL}/${did}`;
+
+      console.log("target URL = ", url);
```
wadeking98: should we get rid of this console statement here?

app/src/components/BCIDView.tsx
jleach: Good catch. Fixed.

app/src/components/BCIDView.tsx
jleach: Fixed.

wadeking98 (approved):

PR #599, opened 2022-09-26
chore: add style formatter to repo and apply style formatting

amanji:


jleach (changes requested):
Can you remove any of the files from `iOS/*` and/or `Android/*`. xcode will update them and we get into a pin-pong match where xcode and prettier keep chaning files.

PR #600, opened 2022-09-26
refactor!: Initialization screens

amanji:
This PR may introduce a breaking change to app initialization (hence the `!` as a precaution). Initialization was moved from the Root Stack to the Splash screen (although no real change to the authentication was introduced).

This PR should be tested


jleach:
@amanji Able to add a description to describe the breaking `!` part of the change. Ty?

jleach (approved):

PR #623, opened 2022-10-04
chore: update styling to support a11y enabled device features

amanji:


swcurran:
Hi @amanji --- did you find any a11y items that the Issuer would specify?   I can't think of any, but wanted to be sure.  AFAIK, an issuer might provide braille language sets, but those would be needed for each supported language, right?  E.g. braille-english, braille-french.  Obvs, the alt-text can come from the multi-lingual label and information overlays that the issuer already provides.  Anything else from the issuer to help with a11y?

amanji:
In this case, most of the changes were to improve the layout on devices that have features like Zoom enabled. In the case of issuer-specific content, I wonder if this can be captured through OCA? The only place where Issuers can add specifications is at the credential branding.

amanji:
I suppose that connection information itself could be also be specified by the issuer/verifier (for example a user is connected with a service at the BC Government where they provide language specific descriptions)

amanji:
There's also been talk about machine readable governance and I'm not sure where action menus stand but I suppose those would also be places where issuers/verifiers can add specifications for a11y enhancements.

swcurran:
I don't think the Issuer would have any input on the zoom handling of the wallet.   The OCA is pretty much only about the multilingual information about issuer, and various pieces of info about the credential and attributes.  Things like connections, and Action Menu are up to the handling within that protocol -- it wouldn't be handled in OCA.

amanji:
Could issuers provide things like screen reader text and a11y labels that can override the default values that would be applied to credentials? Those are about the only things I can think of where the issuer can customize a11y. We could then apply those at the credential branding and on attributes in the credential detail pages.

swcurran:
I think they already provide that in the OCA Multilingual files -- the "label" overlay (label to use for each attribute) and "information" overlay (description/sentence/help text about each attribute).  Is that sufficient?  Or is there something additional needed for the screen reader?

amanji:
I think that would be sufficient. By default we would just apply those values as the screen reader text.

PR #668, opened 2022-10-24
fix: credentials branding

cvarjao:
- Fix Verified/Unverified Person Mapping to OCA bundle
- Fix Lawyer Member bundle
- Add Test Person Credential background
- Introduce enum for credential keys (schema or cred def)

Signed-off-by: Clécio Varjão <1348549+cvarjao@users.noreply.github.com>


app/src/assets/branding/credential-branding.ts
```diff
@@ -47,116 +57,101 @@ const digitalIdInvitationCardOverlay = {
...
-}
-
-const idCardBundle = {
-  capture_base: {},
-  overlays: [
-    {
+const createPersonCredentialBundle = (backgroundImageSource: any, verified = true) => {
+  const metaOverlays = []
+  if (verified) {
+    metaOverlays.push({
```
jleach, line 63: Don't know if it's more effective or not but you can short hand this to `metaOverlays = [...metaOverlays, a, b]` perhaps a bit more declarative.

```
elcome to Node.js v16.14.0.
Type ".help" for more information.
> a = [1,2,3]
[ 1, 2, 3 ]
> [...a, 4, 5]
[ 1, 2, 3, 4, 5 ]
> 
```

jleach (dismissed):
LGTM. One comment as an FYI but not a blocker.

app/src/assets/branding/credential-branding.ts
cvarjao: Yes. I am aware of the spread operator. I use that syntax for making copies. I would also need to change from `const ` to `let`. I decided to keep it that way for now. This will eventually go away, and get published somewhere else.

cvarjao: But that did got me to take a look and notice that I forgot to actually use the array as part of the return statement.

jleach (approved):

PR #686, opened 2022-11-02
feat: extend state and move constants

jleach:
This PR implements the new extensible state mechanics of Bifold to add our own custom `developer` state. Included are moving some of the constants used by BCIDView such as the URLs and invitation ID. They will be editable in the eminent Developer screen so this is where they should live rather that a constants file.

Depends on [514](https://github.com/hyperledger/aries-mobile-agent-react-native/pull/514).


amanji (dismissed):

cvarjao (commented):
Are those dependency update needed for this change? or left over from something else? I didn't see any code change related to those dependencies

jleach:
> Are those dependency update needed for this change? or left over from something else? I didn't see any code change related to those dependencies

@cvarjao They match up to bifold package versions. For example, bifold uses 8.1.1 of the keychain but our package was 7.x.

cvarjao:
Ok. So not exactly related, but an opportunity to get them back in sync.  LGTM

cvarjao (approved):

PR #690, opened 2022-11-08
chore: update RN to 0.66.5 and latest bifold

jleach:
As noted in [517](https://github.com/hyperledger/aries-mobile-agent-react-native/pull/517) we need to update to RN 0.66.5 to fix an android package issue that breaks the build. This PR makes said change.

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -63,7 +63,7 @@
     "query-string": "7.0.1",
     "react": "17.0.2",
     "react-i18next": "11.13.0",
-    "react-native": "0.66.1",
+    "react-native": "^0.66.5",
     "react-native-animated-pagination-dots": "0.1.72",
     "react-native-argon2": "2.0.1",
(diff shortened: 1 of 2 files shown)


amanji (dismissed):

app/package.json
```diff
@@ -63,7 +63,7 @@
     "query-string": "7.0.1",
     "react": "17.0.2",
     "react-i18next": "11.13.0",
-    "react-native": "0.66.1",
+    "react-native": "^0.66.5",
```
cvarjao: I think the dependency needs to match exactly the one from bifold. If using tilde or caret expression, they could still  diverge. Depending when `npm install` was ran it could end up with different versions.
I had provided a [script](https://github.com/bcgov/bc-wallet-mobile/blob/main/DEVELOPER.md#updating-packagejson) to automate the syncing of the package files ad the dependencies should always be exact the same.

amanji (approved):

PR #719, opened 2022-11-17
chore: update script settings to auto pub to google play

jleach:
Change settings to see about auto publishing to internal test track.


jleach:
@amanji I fixed that issue where. Looks like I branched of the wrong one and had old changes.

amanji (approved):

PR #833, opened 2023-01-18
feat: added custom notifications to bcwallet

wadeking98:
Added custom notification for BCID Person credential:
image
image

### DIFF
--- a/app/src/index.ts
+++ b/app/src/index.ts
@@ -48,7 +48,7 @@ const configuration: ConfigurationContext = {
       if (dispatch) {
         dispatch({
           type: BCDispatchAction.PERSON_CREDENTIAL_OFFER_DISMISSED,
-          payload: [true]
+          payload: [{personCredentialOfferDismissed: true}]
         })
       }
     },
--- a/app/src/index.ts
+++ b/app/src/index.ts
@@ -10,7 +10,6 @@ import { Bundles } from 'aries-bifold/lib/typescript/App/types/oca'
 import merge from 'lodash.merge'
 
 import bundles from './assets/branding/credential-branding'
-import BCIDView from './components/BCIDView'
 import AddCredentialButton from './components/AddCredentialButton'
 import AddCredentialSlider from './components/AddCredentialSlider'
 import EmptyList from './components/EmptyList'
@@ -19,7 +18,11 @@ import Developer from './screens/Developer'
 import { pages } from './screens/OnboardingPages'
 import Splash from './screens/Splash'
 import Terms from './screens/Terms'
+import PersonCredentialScreen from './screens/PersonCredential'
 import { defaultTheme as theme } from './theme'
+import { useNotifications } from './hooks/notifications'
+import { ReducerAction } from 'react'
+import { BCDispatchAction } from './store'
 
 const localization = merge({}, translationResources, {
   en: { translation: en },
@@ -31,7 +34,6 @@ const configuration: ConfigurationContext = {
   pages,
   splash: Splash,
   terms: Terms,
-  homeContentView: BCIDView,
   credentialListHeaderRight: AddCredentialButton,
   credentialListOptions: AddCredentialSlider,
   credentialEmptyList: EmptyList,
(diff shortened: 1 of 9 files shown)


app/src/store.tsx
```diff
@@ -19,9 +20,14 @@ interface AddCredential {
   addCredentialPressed: boolean
 }
 
+interface DismissPersonCredentialOffer {
+  personCredentialOfferDissmissed: boolean
```
bryce-mcmath: This is nitpicky, but I believe this should be `personCredentialOfferDismissed` with one less S.

app/src/index.ts
```diff
@@ -40,6 +42,22 @@ const configuration: ConfigurationContext = {
...
+    component: PersonCredentialScreen,
+    onCloseAction: (dispatch?: React.Dispatch<ReducerAction<any>>) => {
+      if (dispatch) {
+        dispatch({
+          type: BCDispatchAction.PERSON_CREDENTIAL_OFFER_DISMISSED,
+          payload: [true]
+        })
+      }
+    },
+    pageTitle: "PersonCredential.PageTitle",
```
jleach, line 55: Why not use `t('PersonCredential.PageTitle')`?`

app/src/screens/Splash.tsx
```diff
@@ -91,6 +92,21 @@ const Splash: React.FC = () => {
     }
   }
 
+  const loadPersonNotificationDismissed = async (): Promise<void> => {
+    try {
+      const dismissedData = await AsyncStorage.getItem('PersonCredentialOfferDismissed')
+      if(dismissedData){
```
jleach: This line wasn't caught by prettier. `if(x){` should be more spacious `if (x) {` for readability.

app/src/screens/Splash.tsx
```diff
@@ -91,6 +92,21 @@ const Splash: React.FC = () => {
...
+        const dismissed = JSON.parse(dismissedData)
+        dispatch({
+          type: BCDispatchAction.PERSON_CREDENTIAL_OFFER_DISMISSED,
+          payload: [dismissed.personCredentialOfferDissmissed],
+        })
+      }
+    } catch (error) {
```
jleach: Unless catch. May as well ax it since its not doing anything.

https://eslint.org/docs/latest/rules/no-useless-catch

app/src/index.ts
```diff
@@ -40,6 +42,22 @@ const configuration: ConfigurationContext = {
   record: Record,
   indyLedgers: selectedLedgers,
   settings: [],
+  customNotification: {
+    component: PersonCredentialScreen,
+    onCloseAction: (dispatch?: React.Dispatch<ReducerAction<any>>) => {
+      if (dispatch) {
+        dispatch({
+          type: BCDispatchAction.PERSON_CREDENTIAL_OFFER_DISMISSED,
+          payload: [true]
```
jleach: I think its more common in the app to dispatch the object rather than just the value. Able to check. For example:

```
        payload: [{ didDismissSomething: true }],
```

app/src/screens/PersonCredential.tsx
```diff
@@ -0,0 +1,166 @@
...
+        outerHeaderContainer: {
+            flexDirection: 'column',
+            backgroundColor: transparent,
+            height: cardHeaderHeight + borderPadding,
+            borderTopLeftRadius: borderRadius,
+            borderTopRightRadius: borderRadius,
+        },
+        innerHeaderContainer: {
+            flexDirection: 'row',
+            marginLeft: borderPadding,
```
jleach: Does `margin: 8` do the same thing as adding all 4?

jleach (changes requested):

app/src/index.ts
wadeking98: I tried that but useTranslations is a hook and react-native doesn't let you use hooks outside of components

app/src/store.tsx
wadeking98: good catch, resolved

app/src/screens/Splash.tsx
wadeking98: resolved

app/src/screens/Splash.tsx
wadeking98: resolved

app/src/screens/PersonCredential.tsx
wadeking98: resolved

app/src/screens/Splash.tsx
```diff
@@ -90,6 +91,17 @@ const Splash: React.FC = () => {
     }
   }
 
+  const loadPersonNotificationDismissed = async (): Promise<void> => {
+    const dismissedData = await AsyncStorage.getItem('PersonCredentialOfferDismissed')
```
jleach: Best to always wrap await in a try/catch unless it creates a unless try/catch. In this case I might consider putting the getItem and JSON.parse in a separate fn with a try/catch so it returns `undefined` if it throws.

app/src/screens/Splash.tsx
```diff
@@ -100,6 +112,9 @@ const Splash: React.FC = () => {
         // load authentication attempts from storage
         const attemptData = await loadAuthAttempts()
 
+        // load BCID person credential notification dismissed state from storage
+        await loadPersonNotificationDismissed()
```
jleach, line 126: Usually in try/catch unless usles.

jleach (dismissed):
Two comments for consideration. Your call what to do with 'em.

jleach:
@wadeking98 Sorry about the merge conflict. Looks like my previous PR conflicted with this one. Looked kind of big. Might be easier to the the file from `main` and just copy-paste your changes into it.

jleach (approved):

PR #840, opened 2023-01-24
chore: update to use AFJ 0.3.3

jleach:
This PR updates the "bifold" submodule to bring in recent changes including the update to AFJ 0.3.3. This requires some local changes to make the BC Wallet also use AFJ 0.3.0.


jleach:
@amanji @wadeking98 @bryce-mcmath @cvarjao This PR brings 0.3.3 into BC Wallet. Without it updating the "Bifold" submodule will break things. I've fixed the API as needed but need to figure out some custom API calls we're doing for getting the BCSC. That feature won't work just yet. Working on it but merge in this PR if you need to do anything with BC Wallet.

amanji (approved):

PR #852, opened 2023-01-26
fix: add kotlinstdlib dep resolution to build.gradle

### LINKED ISSUE #851: Android build fails because of dependency conflict.
The Android build started failing with the error noted below. Need to find a work around until we understand what caused the problem. Everything seems to be working fine in the Bifold build.

```
* What went wrong:
Execution failed for task ':app:mergeDebugJavaResource'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.MergeJavaResWorkAction
   > 2 files found with path 'META-INF/kotlin-stdlib-jdk7.kotlin_module' from inputs:
      - /home/runner/.gradle/caches/transforms-3/62a06b451886688aece40cfc4a27d866/transformed/jetified-kotlin-stdlib-jdk7-1.5.20.jar
      - /home/runner/.gradle/caches/transforms-3/08622b15bbb1c46621e81d1beaf7316b/transformed/jetified-kotlin-stdlib-1.8.0.jar
     Adding a packagingOptions block may help, please refer to
     https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.PackagingOptions.html
     for more information
```

bryce-mcmath:
Resolves #851 

Not an ideal solution, but seems to work until we can figure out why (without this) bifold will build but bc wallet won't. In future, locking deps like these seems like a good idea

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ node_modules
 .gradle
 lib
 yarn.lock
+local.properties
 
 # aries-bifold should be cloned separately
 aries-bifold/
--- a/app/android/app/build.gradle
+++ b/app/android/app/build.gradle
@@ -77,6 +77,17 @@ import com.android.build.OutputFile
  * ]
  */
 
+// added 1/26/2023 by BM, when our builds suddenly started failing due to conflicting kotlin stdlib versions
+configurations.all {
+    resolutionStrategy {
+        eachDependency {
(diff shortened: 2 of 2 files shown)


jleach (approved):
LGTM. Better to not update the bifold module in a PR (do a separate one) if its not needed for the PR.

bryce-mcmath:
> LGTM. Better to not update the bifold module in a PR (do a separate one) if its not needed for the PR.

Noted for next time, thanks

PR #853, opened 2023-01-27
feat: add progress ui features to initialization screen

### LINKED ISSUE #829: Initialization & loading  flow
- provide some granular status (perhaps starting with 2 steps)
- initializing x of `n` ledgers
- tips and hints while waiting : [see wireframes](https://xd.adobe.com/view/1fa13810-e597-4029-8e89-7643ee880a4b-4af2/)
- better error messages (provide some indication of the cause)

bryce-mcmath:
Resolves #829

Here is the new flow when a user starts the app:
initialization

Here is the flow when an error occurs during initialization (manually thrown in this case):
manual_error

Here is the XD flow for reference:
https://xd.adobe.com/view/1fa13810-e597-4029-8e89-7643ee880a4b-4af2/

Things to note:
- The Splash screen remains mounted before and after a user has to enter their PIN, so the tip carousel continues scrolling in the background while they enter it. Not really a big issue imo but worth noting

### DIFF
--- a/app/src/components/ProgressBar.tsx
+++ b/app/src/components/ProgressBar.tsx
@@ -1,6 +1,6 @@
 import { useTheme } from 'aries-bifold'
 import React, { useEffect, useState } from 'react'
-import { StyleSheet, View, LayoutChangeEvent, Animated } from 'react-native'
+import { StyleSheet, View, Animated, useWindowDimensions } from 'react-native'
 
 export interface ProgressBarProps {
   progressPercent: number
@@ -8,7 +8,7 @@ export interface ProgressBarProps {
 
 const ProgressBar: React.FC<ProgressBarProps> = ({ progressPercent }) => {
   const { ColorPallet } = useTheme()
-  const [progressBarWidth, setProgressBarWidth] = useState(700)
+  const { width: windowWidth } = useWindowDimensions()
   const [progressBarScale] = useState(new Animated.Value(0))
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


bryce-mcmath:
I think the SonarCloud failure can be dismissed, it's just because Math.random was used to randomize the order of the tips

app/src/screens/Splash.tsx
```diff
@@ -205,30 +239,54 @@ const Splash: React.FC = () => {
         newAgent.registerOutboundTransport(wsTransport)
         newAgent.registerOutboundTransport(httpTransport)
 
+        setProgress(t('Init.InitializingAgent'), 55)
```
amanji: Instead of a hardcoded number can we just do some division based on number of steps?

app/src/components/TipCarousel.tsx
```diff
@@ -0,0 +1,151 @@
...
+    arr[j] = temp
+  }
+
+  return arr
+}
+
+// increase / decrease the count in this array if you remove or add tips from the i18n files
+const tipOrder: number[] = shuffleArray([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14])
+
+interface ITip {
```
amanji: I don't know if we're following the convention of prefixing interfaces with 'I'.

app/src/components/ProgressBar.tsx
```diff
@@ -0,0 +1,58 @@
...
+import React, { useEffect, useState } from 'react'
+import { StyleSheet, View, LayoutChangeEvent, Animated } from 'react-native'
+
+export interface ProgressBarProps {
+  progressPercent: number
+}
+
+const ProgressBar: React.FC<ProgressBarProps> = ({ progressPercent }) => {
+  const { ColorPallet } = useTheme()
+  const [progressBarWidth, setProgressBarWidth] = useState(700)
```
wadeking98: From my understanding this is used to position the progress bar animations, right? We might want to use % values or position it relative to the left side of the screen rather than position it relative to the centre. Not sure how it will scale to bigger screens like tablets.

amanji: Concur. Can we use the `width` value from RN dimensions?

app/src/components/TipCarousel.tsx
amanji: Could call this `TipData` or `TipItem` or something like that

app/src/components/ProgressBar.tsx
bryce-mcmath: I had the width being set from the outer containers `onLayout` in case we wanted to use this component somewhere where the width wouldn't be the total screen width, but we probably won't, so I've simplified it to just use the window `width`

app/src/components/TipCarousel.tsx
bryce-mcmath: Good point, changed

app/src/screens/Splash.tsx
bryce-mcmath: Updated

jleach (approved):

PR #892, opened 2023-02-08
added moment to dependancies

wadeking98:

added moment to dependencies to support bifold exact date


bryce-mcmath:
What is the huge change in the root lockfile? There doesn't seem to be any change to the root `package.json` file so I'm wondering why the lockfile changed so much

wadeking98:
> What is the huge change in the root lockfile? There doesn't seem to be any change to the root `package.json` file so I'm wondering why the lockfile changed so much

I think that got updated automatically from running npm install in the root of the project

bryce-mcmath (approved):

PR #907, opened 2023-02-13
feat: separated events from react state

wadeking98:

Moved events from react state to DeviceEventEmitter


app/src/components/AddCredentialButton.tsx
```diff
@@ -1,20 +1,15 @@
...
     const { t } = useTranslation()
     const { ColorPallet } = useTheme()
-    const [, dispatch] = useStore<BCState>()
 
     const activateSlider = useCallback(() => {
-        dispatch({
-            type: BCDispatchAction.ADD_CREDENTIAL_PRESSED,
-            payload: [true],
-        })
+        DeviceEventEmitter.emit("ADD_CREDENTIAL_PRESSED")
```
bryce-mcmath: Could these strings be stored as constants somewhere instead?

app/src/components/AddCredentialButton.tsx
```diff
@@ -1,20 +1,15 @@
-import React, {useCallback} from 'react'
+import React, { useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
-import { StyleSheet, TouchableOpacity } from 'react-native'
+import { StyleSheet, TouchableOpacity, DeviceEventEmitter } from 'react-native'
```
bryce-mcmath, line 4: I'm having trouble finding docs for `DeviceEventEmitter`. Could you link where you found info about this API? Also, does it support both Android and iOS in 0.66.x?

app/src/components/AddCredentialSlider.tsx
```diff
@@ -25,6 +26,13 @@ const AddCredentialSlider: React.FC = () => {
   const navigation = useNavigation()
   const [canUseLSBCredential] = useState<boolean>(true)
 
+  DeviceEventEmitter.addListener("ADD_CREDENTIAL_PRESSED",
```
bryce-mcmath: same as above

app/src/components/AddCredentialSlider.tsx
```diff
@@ -68,10 +76,7 @@ const AddCredentialSlider: React.FC = () => {
   })
 
   const deactivateSlider = useCallback(() => {
-    dispatch({
-      type: BCDispatchAction.ADD_CREDENTIAL_PRESSED,
-      payload: [false],
-    })
+    DeviceEventEmitter.emit("ADD_CREDENTIAL_PRESSED")
```
bryce-mcmath: same as above

app/src/components/EmptyList.tsx
```diff
@@ -1,24 +1,27 @@
...
     message?: string
 }
 
 const EmptyList: React.FC<EmptyListProps> = ({ message }) => {
     const { t } = useTranslation()
     const { ListItems, } = useTheme()
-    const [store, dispatch] = useStore<BCState>()
+    const [addCredentialPressed, setAddCredentialPressed] = useState<boolean>(false)
+
+    DeviceEventEmitter.addListener("ADD_CREDENTIAL_PRESSED",
```
bryce-mcmath: "

app/src/components/EmptyList.tsx
```diff
@@ -1,24 +1,27 @@
...
+            setAddCredentialPressed(newVal)
+        }
+    )
 
     const addCredentialPress = useCallback(() => {
-        dispatch({
-            type: BCDispatchAction.ADD_CREDENTIAL_PRESSED,
-            payload: [true],
-        })
+        DeviceEventEmitter.emit("ADD_CREDENTIAL_PRESSED", !addCredentialPressed)
```
bryce-mcmath: "

app/src/helpers/BCIDHelper.ts
```diff
@@ -232,10 +232,7 @@ export const authenticateWithServiceCard = async (
       code === ErrorCodes.CanceledByUser ? AuthenticationResultType.Cancel : AuthenticationResultType.Fail
     )
 
-    dispatch({
-      type: DispatchAction.ERROR_ADDED,
-      payload: [{ error }],
-    })
+    DeviceEventEmitter.emit("ERROR_ADDED", error)
```
bryce-mcmath: "

app/src/helpers/BCIDHelper.ts
```diff
@@ -268,9 +265,6 @@ export const startFlow = async (
   } catch (error: unknown) {
     setWorkflowInFlight(false)
 
-    dispatch({
-      type: DispatchAction.ERROR_ADDED,
-      payload: [{ error }],
-    })
+    DeviceEventEmitter.emit("ERROR_ADDED", error)
```
bryce-mcmath: "

app/src/components/AddCredentialButton.tsx
wadeking98: it should support both since it's built in to react-native. The documentation does seem to be lacking, but I figured it's best to use something that is already in react native rather than importing a 3rd party library. Here's some documentation I found, the best documentation seems to be for a similar module `NativeEventEmitter`:
https://reactnative.dev/docs/native-modules-android#sending-events-to-javascript
https://scarcoco.github.io/react-native/docs/native-modules-android#sending-events-to-javascript

wadeking98:
Depends on bifold submodule update

app/src/components/EmptyList.tsx
```diff
@@ -1,42 +1,41 @@
...
+    const { ListItems, } = useTheme()
+    const [addCredentialPressed, setAddCredentialPressed] = useState<boolean>(false)
 
-  const addCredentialPress = useCallback(() => {
-    dispatch({
-      type: BCDispatchAction.ADD_CREDENTIAL_PRESSED,
-      payload: [true],
-    })
-  }, [])
+    DeviceEventEmitter.addListener(BCWalletEventTypes.ADD_CREDENTIAL_PRESSED,
```
jleach: I **think** will this lead to multiple subscribers? Wondering if we need to remove the listener if it exists before adding it unless we know it has been cleaned up. My though is something like this.

```TypeScript
useEffect(() -> {

DeviceEventEmitter.addListener(BCWalletEventTypes.ADD_CREDENTIAL_PRESSED,
        (value?: boolean) => {
            const newVal = value === undefined ? !addCredentialPressed : value
            setAddCredentialPressed(newVal)
        }
    )

    return () => {
	DeviceEventEmitter.removeListener(BCWalletEventTypes.ADD_CREDENTIAL_PRESSED)
     }
}, [])
```

EDIT: Made the example more realistic as noted in Bryce's comment below.

jleach: @bryce-mcmath You know react - thoughts?

jleach (changes requested):
Requesting changes, just meeds lets discuss...

app/src/components/EmptyList.tsx
bryce-mcmath: agree, should unsubscribe in a cleanup function (assuming the above is just pseudocode, I don't think it's actually valid. You would need to add an empty dep array and return an actual function from the useEffect callback)

wadeking98: @jleach I'm not quite sure what you mean, the add listener is in the root of the component, not inside a useEffect or anything. It will only get called once

wadeking98: nevermind, you guys are right. I'll update

bryce-mcmath:
Are you running the linter + prettier?

wadeking98:
> Are you running the linter + prettier?
my bad, I thought it ran automatically

wadeking98:
> Are you running the linter + prettier?

okay I've updated with linting completed

bryce-mcmath:
> > my bad, I thought it ran automatically

I think it should, you just might have to run `npx husky@7.0.0 install` in the root of `bc-wallet-mobile`

bryce-mcmath (approved):

jleach (approved):

jleach:
@wadeking98 Ready to merge.

PR #844, opened 2023-01-25
feat: credential branding update

amanji:
Update with new branding assets and bundle configurations

PR #919, opened 2023-02-15
chore: configure linting, hooks, jest, run lint

bryce-mcmath:
Git hooks are now functioning as they are in bifold, the linter did catch quite a few issues so I made those changes. Jest is setup but only one test is running as it has trouble running tests of components that import from Bifold, so I'm gonna leave that ticket for someone else to take a swing at. I had to add a rule to the BC Wallet eslint config that isn't present in the bifold config, the rule to make unused-vars an error and not just a warning. This matches the behavior in bifold (not sure why bifold didn't need to explicitly state the rule.) You may need to run `npx husky@7.0.0 install` in the root once to get hooks working for you. Alternatively you can use the new `setup` script in the `package.json`, it does the same thing.

Apologies for touching so many files.

Blocked By
- [x] #844


amanji (dismissed):

amanji (commented):
For this line in the package.json:

```
"install:bifold": "cd bifold/core && yarn install"
```

Can we update it to:

```
"install:bifold": "cd bifold/core && yarn install --frozen-lockfile",
```

amanji (approved):

PR #934, opened 2023-02-24
fix: stuck on person screen

jleach:
Fix an issue where the user can end up stuck on the Person Credential screen after the worklfow is completed. When this happens the user may not know what to do next.


bryce-mcmath:
You might have to install git hooks with `npx husky@7.0.0 install` in the root of the project. Looks like the linter and tsc step weren't run

jleach:
Ok, ready now @bryce-mcmath. I needed to make the change in two places.

bryce-mcmath (approved):
Approved but I still think the linter needs to run, looks like there's at least one unused import (`ReducerAction`). Otherwise, LGTM

PR #946, opened 2023-03-02
fix: prevent back nav to splash page, run lint

bryce-mcmath:
Resetting the nav stack whenever the splash page is navigated from, this PR in combination with https://github.com/hyperledger/aries-mobile-agent-react-native/pull/647 should fix the majority of the back navigating issues we have been having with the iOS swipe-back and the Android hardware back button.

I also ran the linter and made the changes it demanded. In most cases where there were unused variables/imports I simply commented them out since you may want them for reference.

### DIFF
--- a/app/src/screens/PersonCredential.tsx
+++ b/app/src/screens/PersonCredential.tsx
@@ -3,18 +3,7 @@ import { useNavigation } from '@react-navigation/core'
 import { Button, ButtonType, Screens, useStore, useTheme } from 'aries-bifold'
 import React, { useState, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
-import {
-  StyleSheet,
-  Text,
-  View,
-  // Image,
-  Dimensions,
-  // ImageBackground,
-  TouchableOpacity,
-  Linking,
-  ScrollView,
-  SafeAreaView,
-} from 'react-native'
+import { StyleSheet, Text, View, Dimensions, TouchableOpacity, Linking, ScrollView, SafeAreaView } from 'react-native'
 
 import LoadingIcon from '../components/LoadingIcon'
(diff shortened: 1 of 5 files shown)


app/src/screens/PersonCredential.tsx
```diff
@@ -1,15 +1,15 @@
...
 import { Button, ButtonType, Screens, useStore, useTheme } from 'aries-bifold'
-import React, { ReducerAction, useState, useCallback } from 'react'
+import React, { useState, useCallback } from 'react'
 import { useTranslation } from 'react-i18next'
 import {
   StyleSheet,
   Text,
   View,
-  Image,
+  // Image,
```
jleach: Can we remove dead code?

app/src/screens/Splash.tsx
```diff
@@ -204,22 +205,42 @@ const Splash: React.FC = () => {
           })
 
           if (onboardingComplete(dataAsJSON) && !attemptData?.lockoutDate) {
-            navigation.navigate(Screens.EnterPIN as never)
+            navigation.dispatch(
+              CommonActions.reset({
```
jleach, line 209: Cool. Didn't know we could do this.

jleach (changes requested):
I think a few comments to clean up but otherwise LGTM.

app/src/screens/Splash.tsx
bryce-mcmath: Yeah it's pretty handy! Took some researching to find how to do it

app/src/screens/PersonCredential.tsx
bryce-mcmath: Done

jleach (approved):

PR #960, opened 2023-03-09
fix: person credential workflow fix

amanji:
Wallet will now route to person credential offer when received by IDIM.


app/src/components/AddCredentialSlider.tsx
```diff
@@ -21,24 +21,18 @@ const AddCredentialSlider: React.FC = () => {
...
+
   const credentials = [
     ...useCredentialByState(CredentialState.CredentialReceived),
     ...useCredentialByState(CredentialState.Done),
   ]
+
   const navigation = useNavigation()
   const [canUseLSBCredential] = useState<boolean>(true)
 
-  useEffect(() => {
```
wadeking98, line 31: any reason why we're removing this here?

amanji: Just organizing them all in one place

wadeking98: ah okay, I see it now

wadeking98 (approved):
Couple code smells to address but otherwise looks good

PR #973, opened 2023-03-13
fix: person credential workflow pt. 2

amanji:
This adds a fix when the person credential workflow is activated from the Home Screen


bifold
cvarjao, line 1: Did you intend to update bifold submodule as requirement for this PR? or just a side effect?

app/src/components/AddCredentialSlider.tsx
```diff
@@ -79,39 +77,34 @@ const AddCredentialSlider: React.FC = () => {
...
+      setAddCredentialPressed(newVal)
+    })
+
+    return () => {
+      handle.remove()
+    }
+  }, [])
 
   useEffect(() => {
     const handle = DeviceEventEmitter.addListener(BCWalletEventTypes.ADD_CREDENTIAL_PRESSED, (value?: boolean) => {
```
bryce-mcmath: Maybe my eyes are bugging out but is this useEffect duplicated here? Also we could return just `handle.remove` as the cleanup function

app/src/components/CredentialOfferTrigger.tsx
```diff
@@ -0,0 +1,34 @@
...
+      if (credential.state == CredentialState.OfferReceived && credential.connectionId === workflowConnectionId) {
+        goToCredentialOffer(credential.id)
+      }
+    }
+  }, [offers, workflowConnectionId])
+
+  return null
+}
+
+export default CredentialOfferTrigger
```
bryce-mcmath: You probably have a good reason for this but why an empty component rather than a custom hook? Would `useNavigation` and `useCredentialByState` cause issues with that approach?

bryce-mcmath (changes requested):
Just a duplicated `useEffect`, other than that LGTM

app/src/components/AddCredentialSlider.tsx
amanji: Merge conflicts...

app/src/components/CredentialOfferTrigger.tsx
amanji: Nope, good suggestion!

bryce-mcmath (dismissed):
Slick! Nice work

bryce-mcmath (approved):

PR #991, opened 2023-03-21
chore: add storybook to bc-wallet-mobile

bryce-mcmath:
I found this branch super useful, opening this PR to suggest moving it into main


app/metro.config.js
```diff
@@ -21,8 +23,9 @@ for (const packageDir of packageDirs) {
     return acc
   }, extraNodeModules)
 }
-console.dir(extraExclusionlist)
-console.dir(extraNodeModules)
+// console.dir(extraExclusionlist)
+// console.dir(extraNodeModules)
+// console.log(`inlineRequires = ${process.env.LOAD_STORYBOOK === 'true'}`)
```
wadeking98: should we delete these console lines?

app/index.js
```diff
@@ -44,12 +43,29 @@ const navigationTheme = {
...
+      <NavigationContainer theme={navigationTheme}>
+        <App />
+      </NavigationContainer>
+    )
+  }
+  AppRegistry.registerComponent(appName, () => Base)
+} else {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const App = require('./App').default
+  const Base = () => {
```
wadeking98: sonar cloud is complaining about this duplicate `Base` function maybe we can factor it out to a higher scope so we don't have to duplicate it

wadeking98:
after the sonarcloud issues are addressed it looks good to merge

wadeking98 (approved):

PR #1006, opened 2023-03-28
fix: android and ios build and pipeline issues

bryce-mcmath:
Included `MaterialCommunityIcons.tff` in Android since it was already included in iOS. Updating the Gradle version fixed the Android build issue, I guess that was changed in the Storybook PR. Changed the `.ipa` name to match BCWallet rather than AriesBifold since the product name changed, hoping that fixes the iOS pipeline issue.


app/android/app/src/main/java/ca/bc/gov/BCWallet/MainApplication.java
```diff
@@ -4,6 +4,7 @@
 import android.content.Context;
 import com.facebook.react.PackageList;
 import com.facebook.react.ReactApplication;
+import com.oblador.vectoricons.VectorIconsPackage;
```
cvarjao, line 7: what is this extra import for? it doesn't seem to be used

app/android/app/src/main/assets/fonts/MaterialCommunityIcons.ttf
cvarjao, line 1: This should be taken care by react native auto-linking feature:
https://github.com/oblador/react-native-vector-icons#option-with-gradle-recommended

I do see the [apply](https://github.com/bcgov/bc-wallet-mobile/blob/main/app/android/app/build.gradle#L96) in the gradle file as per documentation.
```
apply from: "../../node_modules/react-native-vector-icons/fonts.gradle"
```
Did you try if after updating gradle it fixed itself without the extra copy?

app/android/app/src/main/java/ca/bc/gov/BCWallet/MainApplication.java
bryce-mcmath: That came from linking `react-native-vector-icons`. I can try removing it and see if it still works

app/android/app/src/main/assets/fonts/MaterialCommunityIcons.ttf
bryce-mcmath: I just added it to be safe since all the other (`BCSans-Regular.ttf`, `MaterialIcons.ttf`, etc.) fonts are in that directory. But I can remove it if you want, I think it should still work

cvarjao (approved):

PR #1041, opened 2023-04-19
fix: Use credential definition id as restriction for proof request templates

Artemkaaas:
Updated proof request templates to use credential definition id instead of the schema as a restriction.  
Depends on the changes in bifold: https://github.com/hyperledger/aries-mobile-agent-react-native/pull/719


wadeking98:
Hi Artem, I've already updated the requests in my PR https://github.com/bcgov/bc-wallet-mobile/pull/1040. Thanks for all your work, one thing I noticed is that the restriction still says `schema_id` but is using a credential definition id. In my PR I switched to using `issuer_did` and `schema_id` just so we don't have to updated the template when an issuer updates their credential definition.

PR #1048, opened 2023-04-24
chore: gate give feedback

jleach:
Only allow the give feedback option to be displayed if the app is running in developer mode.


app/src/components/HomeContentView.tsx
```diff
@@ -80,28 +82,34 @@ const HomeContentView: React.FC<HomeContentViewProps> = ({ children }) => {
 
   const toggleSurveyVisibility = () => setSurveyVisible(!surveyVisible)
 
+  console.log(store.preferences.developerModeEnabled)
```
bryce-mcmath: Leftover console log

bryce-mcmath (approved):

PR #1060, opened 2023-04-26
add Person DEV cred def, and map bundle

marcos-carretero:


cvarjao:
@marcos-carretero , just need missing DCO signoff:
use `git commit` with `-s` (lower case)  - [reference](https://probot.github.io/apps/dco/)

bryce-mcmath (approved):

PR #1064, opened 2023-04-26
feat: add demo req templates

wadeking98:
added proof request templates for IDIM, LSBC, and BC Wallet Showcase.
(Warning): This means that the Showcase credential will be accepted by the bc wallet verifier functionality. Users can make credentials that satisfy a given proof request template for the Person Credential and Member Card

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -16,4 +16,4 @@ sonar.exclusions=**/localization/**/index.ts
 #sonar.sourceEncoding=UTF-8
 
 # Exclusions for copy-paste detection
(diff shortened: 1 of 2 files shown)


.sonarcloud.properties
```diff
@@ -4,7 +4,7 @@
 
 # Path to sources
 sonar.sources=.
-sonar.exclusions=**/localization/**/index.ts
+sonar.exclusions=**/localization/**/index.ts,**/**/request-template.ts
```
bryce-mcmath: maybe add it to `sonar.cpd.exclusions=` on line 19? Not sure

bryce-mcmath (approved):

PR #1072, opened 2023-04-27
feat: enable retries for failures on init/splash 

bryce-mcmath:
retries


app/src/screens/Splash.tsx
```diff
@@ -74,6 +79,9 @@ const Splash: React.FC = () => {
   const { indyLedgers } = useConfiguration()
   const [stepText, setStepText] = useState<string>(t('Init.Starting'))
   const [progressPercent, setProgressPercent] = useState(0)
+  const [initOnboardingCount, setInitOnboardingCount] = useState(0)
```
jleach, line 82: Not in this PR but I'm wondering if we should get in the habit of using `useReducer` when we have quite a few state vars like `Splash.tsx` or just using one single state object.

jleach (approved):

PR #1093, opened 2023-05-09
feat: add photo attribute

wadeking98:
Added photo attribute display to the showcase credentials. This won't effect production person credentials
Screen Shot 2023-05-09 at 8 57 04 AM
Screen Shot 2023-05-09 at 9 32 23 AM

### DIFF
--- a/app/src/assets/branding/credential-branding.ts
+++ b/app/src/assets/branding/credential-branding.ts
@@ -119,7 +119,7 @@ const memberCardBundle = {
   ],
 }
 
-const createPersonCredentialBundle = (backgroundImageSource: string, verified = true) => {
+const createPersonCredentialBundle = (backgroundImageSource: string, verified = true, dev = false) => {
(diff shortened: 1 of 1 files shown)


jleach (approved):

app/src/assets/branding/credential-branding.ts
```diff
@@ -214,6 +224,10 @@ const createPersonCredentialBundle = (backgroundImageSource: string, verified =
       } as LabelOverlay,
     ],
   }
+  if (dev && overlay.captureBase.attributes){
```
cvarjao, line 227: why is this applied only for "dev"?

cvarjao: I think I got it. Is it so the "picture" attribute doesn't show up for prod credentials just yet?

wadeking98: dev just means that it's only applied to the showcase credentials, not the IDIM person credentials

Issue #1087, opened 2023-05-04
verifier template for custom age shows as green/passed when it is not


cvarjao:
How to produce:
- Create a proof Request with a proof of age that is larger than you current have (e.g.: `70`)
- The verifier will receive the response and erroneously show as passed (green)

![screenshot](https://user-images.githubusercontent.com/1348549/236327442-64806f55-db37-4c6d-9650-58cb6e032317.png)

PR #1105, opened 2023-05-12
fix: fixed date of birth check in proof template

### LINKED ISSUE #1084: verifier template for 19+ is actually asking for 18+
https://github.com/bcgov/bc-wallet-mobile/blob/e22d6a2abeace80de38afe46d4f83ee1f86a50cf/app/src/request-templates.ts#L46

wadeking98:
Previously the request template was checking that the date of birth in a credential was greater than or equal to 18. The date of birth in most credentials uses the `YYYYMMDD` format. This means that the birthdate would always pass the check since any date string will resolve to a number greater than 18.

I changed the request template to calculate the date 19 years before today and to use that in a proof request instead of the number 18

Resolves: #1084

### DIFF
--- a/app/src/request-templates.ts
+++ b/app/src/request-templates.ts
@@ -1,6 +1,12 @@
 import { PredicateType } from '@aries-framework/core'
 import { ProofRequestTemplate, ProofRequestType } from 'aries-bifold'
 
+const calculatePreviousYear = (yearOffset: number) => {
+  const pastDate = new Date()
+  pastDate.setFullYear(pastDate.getFullYear() - yearOffset)
+  return parseInt(pastDate.toISOString().split('T')[0].replace(/-/g, ''))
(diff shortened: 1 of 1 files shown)


cvarjao:
Does this also fix #1087? By including the date in that format it will make it less user friendly for parameterized inputs

swcurran:
To do a predicate, it must be in that format. But what about defining a way to do some processing on the presentation request between what the user writes, and what is sent in the protocol.  That was the thought I had with the “Presentation Request Template” idea.

wadeking98:
> Does this also fix #1087? By including the date in that format it will make it less user friendly for parameterized inputs

Yes this should also fix that issue

jleach (approved):

bryce-mcmath (approved):

PR #1144, opened 2023-06-02
chore: convert the project to use yarn

### LINKED ISSUE #281: Convert our bc wallet project to use yarn
We are going to convert the main Bifold repo to use yarn as per [this issue](https://github.com/hyperledger/aries-mobile-agent-react-native/issues/291) so we should follow suite for consistency.

Questions:
- [ ] Continue with git submodule or use git url dependency?
- [ ] Check Jest
- [ ] Use `yarn patch` rather than patch-package to match Bifold

If we use `yarn patch` I think we can just move the patches to `.yarn/patch` like in Bifold and it will just work. Then we can remove the `patch-package` package.

jleach:
Convert the project to use yarn, this includes:

- user yarn as the only package manager;
- resolve dependency issue caused by move between "@react-navigation/core" or "@react-navigation/native";
- move patching mechanics to use yarn;
- general related cleanup.

Fixes #281


app/package.json
```diff
@@ -1,44 +1,43 @@
...
-  "description": "",
+  "description": "BC Wallet to hold Verifiable Credentials",
   "main": "index.js",
   "private": true,
+  "license": "Apache-2.0",
   "engines": {
-    "node": ">= 16.10.0",
-    "npm": ">= 8.0.0"
+    "node": ">=16.15.0 <18.0.0",
+    "yarn": "^1.22.19"
```
bryce-mcmath: I noticed you're using 3.3.1 in other locations but here you have 1.22.19, is there a reason for that?

app/src/screens/Splash.tsx
```diff
@@ -10,8 +10,7 @@ import {
 import { useAgent } from '@aries-framework/react-hooks'
 import { agentDependencies } from '@aries-framework/react-native'
 import AsyncStorage from '@react-native-async-storage/async-storage'
-import { useNavigation } from '@react-navigation/core'
-import { CommonActions } from '@react-navigation/native'
+import { CommonActions, useNavigation } from '@react-navigation/native'
```
bryce-mcmath, line 13: Curious, why are we switching from core to native here?

bryce-mcmath (commented):
Just a couple questions

app/src/screens/Splash.tsx
bryce-mcmath: Just saw your other PR in Bifold, you can ignore the above comment. Does this PR depend on that PR at all then?

jleach: No dependency and that one will be more complicated. What happened was we were mixing and matching where `useNavigation` came from. Some times `core` and others `native`. But the hooks need (in BC Wallet's case) to be one or the other - not both. I settled on the one I though was best. This was causing a problem when switching to yarn, probably because of how they're resolved under the hood. Opinions welcome.

app/package.json
jleach: Fixed.

bryce-mcmath (approved):

PR #1189, opened 2023-06-15
feat: rocketchat notification

bryce-mcmath:
Making use of https://github.com/RocketChat/Rocket.Chat.GitHub.Action.Notification


.github/workflows/main.yaml
```diff
@@ -235,6 +235,19 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+      - name: Rocket.Chat iOS Notification
+        if: failure() && github.ref_name == 'main' && needs.check-ios-secrets.outputs.isReleaseBuild == 'true'
+        uses: RocketChat/Rocket.Chat.GitHub.Action.Notification@master
+        with:
+          type: ${{ job.status }}
+          job_name: 'BC Wallet iOS Build - Run number ${{ github.run_number }}'
+          mention: 'jason.leach @cvarjao @wade.king @bryce.mcmath'
```
jleach: Do I need an`@` in my name?

.github/workflows/main.yaml
```diff
@@ -344,6 +357,19 @@ jobs:
           if-no-files-found: error
           retention-days: 7
 
+      - name: Rocket.Chat Android Notification
+        if: failure() && github.ref_name == 'main' && needs.check-android-secrets.outputs.isReleaseBuild == 'true'
+        uses: RocketChat/Rocket.Chat.GitHub.Action.Notification@master
+        with:
+          type: ${{ job.status }}
+          job_name: 'BC Wallet Android Build - Run number ${{ github.run_number }}'
+          mention: 'jason.leach @cvarjao @wade.king @bryce.mcmath'
```
jleach: Same.

jleach (changes requested):
One change on my name. Since its the same YAML in 2x places consider making it a separate action like we do for installing node.

.github/workflows/main.yaml
bryce-mcmath: See here how mention is used: https://github.com/RocketChat/Rocket.Chat.GitHub.Action.Notification/blob/2966f5f2e0313a94f965a51c08700cfecf0b6572/src/rocketchat.ts#L120

It's just string interpolation, an '@' followed by whatever you put as `mention`. So the first name doesn't need an @. Not pretty but that's how they implemented it

.github/workflows/main.yaml
bryce-mcmath: See my other comment

jleach (approved):

PR #1222, opened 2023-06-22
chore: adapt to use AFJ 0.4.0

jleach:
Adapt the BC Wallet to work with AFJ 0.4.0 and shared components provided by the newly updated Bifold package. Depends on hyperledger/aries-mobile-agent-react-native#666


bryce-mcmath (approved):
Loving the deletions. Nice work

PR #1229, opened 2023-06-30
feat: implement wallet naming in bc wallet

bryce-mcmath:
Here's what it looks like: 
name_bc_wallet

Here's a contact with name showing up:
bryces_wallet

### DIFF
--- a/app/src/screens/Splash.tsx
+++ b/app/src/screens/Splash.tsx
@@ -47,13 +47,35 @@ enum InitErrorTypes {
   Agent,
 }
 
-const onboardingComplete = (state: OnboardingState): boolean => {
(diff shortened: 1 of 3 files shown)


jleach (approved):

jleach (dismissed):

app/src/screens/Splash.tsx
```diff
@@ -304,7 +321,7 @@ const Splash: React.FC = () => {
         setStep(5)
         const options = {
           config: {
-            label: 'BC Wallet',
```
jleach, line 307: Would it be possible for the user to enter a blank name and break the wallet install? Since our poking around the other day it needs to be file-system safe characters. Probably not names like `../../my-wallet`.

jleach (approved):

PR #1247, opened 2023-07-07
feat: dev only verifier templates

### LINKED ISSUE #1213: Review/Update proof request templates descriptions
Review and update the description for all proof request templates

https://github.com/bcgov/bc-wallet-mobile/blob/384077e2ce14cbb4dabed67783e55b88e65b020e/app/src/request-templates.ts#LL54C53-L54C53

wadeking98:
Sheldon asked for Unverified Person templates for ease of testing, Decided to make them development only templates so they don't end up in prod.
- Added extra toggle on developer menu to include development templates
- Bound inputs so that when the dev templates toggle is on, verifier feature must be enabled. Conversely when the verifier feature is turned off, dev template must also be turned off 

Resolves: #1213 

https://github.com/hyperledger/aries-mobile-agent-react-native/assets/36937407/d2910669-c842-4501-b2da-3d351cb80ed2

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -187,7 +187,9 @@ jobs:
           -archivePath AriesBifold.xcarchive \
           -sdk iphoneos \
           -verbose \
-          archive
+          archive \
+          CODE_SIGNING_ALLOWED=NO \
+          CODE_SIGNING_REQUIRED=NO
 
       - name: Debug build
         if: github.ref_name != 'main' || needs.check-ios-secrets.outputs.isReleaseBuild != 'true'
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -187,9 +187,7 @@ jobs:
           -archivePath AriesBifold.xcarchive \
           -sdk iphoneos \
           -verbose \
-          archive \
-          CODE_SIGNING_ALLOWED=NO \
-          CODE_SIGNING_REQUIRED=NO
+          archive
 
       - name: Debug build
         if: github.ref_name != 'main' || needs.check-ios-secrets.outputs.isReleaseBuild != 'true'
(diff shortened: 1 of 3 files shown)


app/src/request-templates.ts
```diff
@@ -50,41 +64,25 @@ export const proofRequestTemplates: Array<ProofRequestTemplate> = [
...
-                { schema_id: '2K2h7kf8VGTLtfoxJgWazf:2:Person:1.1', issuer_did: '2K2h7kf8VGTLtfoxJgWazf' }, // Dev & Test
-              ],
+              restrictions: personRestrictions,
             },
           ],
           requestedPredicates: [
             {
               name: 'birthdate_dateint',
-              predicateType: '>=',
+              predicateType: '<=',
```
bryce-mcmath, line 83: Are we sure this is a change we want? It seemed to be working before

.github/workflows/main.yaml
```diff
@@ -187,7 +187,9 @@ jobs:
           -archivePath AriesBifold.xcarchive \
           -sdk iphoneos \
           -verbose \
-          archive
+          archive \
+          CODE_SIGNING_ALLOWED=NO \
+          CODE_SIGNING_REQUIRED=NO
```
bryce-mcmath: Did you mean to include this? If so, should probably be a separate PR I think

bryce-mcmath (changes requested):

app/src/request-templates.ts
wadeking98: It only works on the same day the credential was issued. If the showcase person credential was issued on the previous day (July 9th) and today is July 10th then the request will fail because July 9th is less than July 10th. The reason it was set to ">=" in the first place was because when the templates were originally created, they used exact age instead of date of birth. ie: "date of birth >= 19" instead of "date of birth >= current date - 19 years" which was a bug because any date_int number would be greater that 19. When I fixed the bug I forgot to flip the direction of the ">=" sign.

.github/workflows/main.yaml
wadeking98: woops ya I don't know how that got there

bryce-mcmath (approved):

PR #1310, opened 2023-07-29
feat: updated traction branding

wadeking98:
Updated branding and request templates for the new traction agent


wadeking98:
The Code smell here is intentional, dev and test use the same traction instance right now

bryce-mcmath (approved):

PR #1356, opened 2023-08-16
fix: image size issue

wadeking98:
Patched the askar node wrapper to allow processing of large credential values such as images.


cvarjao (approved):
LGTM

swcurran:
What was the issue and what was the fix?  Nice work!

wadeking98:
> What was the issue and what was the fix? Nice work!

The issue was that text decoder was being called on too big of an input so it was failing. @cvarjao 's solution was to break it up into more manageable chunks that the text decoder could handle and then join all the decoded text at the end

PR #1389, opened 2023-08-17
feat: push notifications

jamshale:
Adds firebase push notification capability when using a mediator with the configured firebase plugin.

Things of note:

- Began adding this to bifold but was blocked by the fact that changes were needed to the android and ios build files, and that the required libraries wouldn't work without a valid firebase account and project. Could maybe figure out a way to add this as a bifold option and pull it out of the bc wallet repo at some point but I don't have a path forward with this right now.
- Because I didn't create the firebase apps in bifold or add the react native libraries I tried to have bifold contain no code about the push notification logic. As a result I didn't follow the same patterns in some places as the bifold supported functionality. For instance I created a modal in this repo instead of bifold and didn't follow the action pattern in the Developer view.
- Most of the development was done in android. IOS actually receiving push notifications is untested. This is because apple requires the app have APNS push notification capability to produce a firebase token, and that requires an apple developer account with signing privileges. I added the logic that I think is needed and added an attempt at a github action to load the required entitlements file. (the env variables are currently hardcoded as well).
- Because I've been unable to test on IOS the push icon is currently the default. I added a BC icon to android.
- The firebase account is my own personal account and they should be replaced by a bc gov managed one.

Here is some sequence diagrams of current features: https://hackmd.io/@NQjeSsXCR3Cmgs2Lr-M8ow/SJCS87BC2

Here is a demo video of current status:

Mediator doesn't support firebase:

[push-notifications-no-mediator.webm](https://github.com/bcgov/bc-wallet-mobile/assets/31809382/29d64449-639d-40d0-88f1-cfd67569a11c)

Mediator does support firebase:

[push-notifications-with-mediator-plugin.webm](https://github.com/bcgov/bc-wallet-mobile/assets/31809382/0f9195c6-09ca-4cad-8a42-45db100aa782)

There are 2 new env variables:

- MEDIATOR_USE_PUSH_NOTIFICATIONS --> if false or undefined the push notification logic will be completely ignored. Set to true if enabled
- MEDIATOR_LABEL (Required) --> This is what is uses to discover firebase protocol and send the token. Otherwise it would need to discover on every agent. If this is a problem then could work around it.

** Currently it will ask for push notifications permission but send a blank token. It will be defaulted to off and will need to be turned on via development screen before receiving push notifications. **

I would like to write some unit tests. I had some earlier but deleted them after the modal broke the required mocks.


jleach:
@jamshale Is this one ready for review? If not are you able to move it to "Draft", otherwise we can remove the "do not merge" label and review it.

jamshale:
> @jamshale Is this one ready for review? If not are you able to move it to "Draft", otherwise we can remove the "do not merge" label and review it.

Awesome! I did some work recently and want to go over it and update the description. I should have it ready for review today.
The 'do not merge' was more about the firebase config files being on my personal account. I have added some additional functionality recently after a discussion with Clecio but it should be ready soon.

cvarjao:
can you please make it "off" by default. We will start using internally, and once we have fully tested we can change to default "on"

jamshale:
> can you please make it "off" by default. We will start using internally, and once we have fully tested we can change to default "on"

Yes will do.

.github/workflows/main.yaml
```diff
@@ -139,6 +139,14 @@ jobs:
           echo "MEDIATOR_URL=${MEDIATOR_URL}" >.env
           echo "IAS_PORTAL_URL=${IAS_PORTAL_URL}" >>.env
           echo "IAS_AGENT_INVITE_URL=${IAS_AGENT_INVITE_URL}" >>.env
+
```
jamshale, line 144: This is my attempt at adding the push notification capability to IOS and adding the env variables. Obviously these shouldn't be hard coded.
I also think this will only work for development builds. See `DEVELOPMENT.AriesBifold.entitlements`

app/__mocks__/@aries-framework/react-hooks.ts
```diff
@@ -23,6 +23,11 @@ const useAgent = () => ({
   agent: {
```
jamshale, line 23: These mocks were added for the unit tests I needed to remove. Will probably come in handy though, so i left them.

jamshale:
@jleach Ok, I would like to open this for review now.

app/src/components/PushNotifications.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const { t } = useTranslation()
+
+  const setupPushNotifications = async () => {
+    setInfoModalVisible(false)
+    if (!agent || (await isUserDenied())) return
+    setup(agent, true)
+  }
+
+  const initializeCapabilityRequest = async () => {
+    if (!agent || !(await isMediatorCapable(agent)) || (await isRegistered())) return
```
bryce-mcmath: Can we make this a bracketed if statement for readability?

app/src/components/PushNotifications.tsx
```diff
@@ -0,0 +1,37 @@
...
+import PushNotificationsModal from '../modals/PushNotificationsModal'
+
+const PushNotifications = () => {
+  const { agent } = useAgent()
+  const [infoModalVisible, setInfoModalVisible] = useState(false)
+  const { t } = useTranslation()
+
+  const setupPushNotifications = async () => {
+    setInfoModalVisible(false)
+    if (!agent || (await isUserDenied())) return
```
bryce-mcmath: Can we make this a bracketed if statement for readability?

app/src/helpers/PushNotificationsHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Checks if the device token is already registered by checking the permission was granted and the storage key was used
+ * @param token - If defined will use this token instead of fetching with firebase
+ * @returns {Promise<boolean>}
+ */
+const isRegistered = async (): Promise<boolean> => {
+  const authorized = (await messaging().hasPermission()) === messaging.AuthorizationStatus.AUTHORIZED
+
+  // Need to register for push notification capability on iOS
+  if (Platform.OS === 'ios' && !messaging().isDeviceRegisteredForRemoteMessages)
+    await messaging().registerDeviceForRemoteMessages()
```
bryce-mcmath: Same as above, would be nice if this was bracketed for readability

app/src/helpers/PushNotificationsHelper.ts
```diff
@@ -0,0 +1,173 @@
...
+/**
+ * Attempts to send the device token to the mediator agent. If the token is blank this is equivalent to disabling
+ * @param agent - The active aries agent
+ * @param blankDeviceToken - If true, will send an empty string as the device token to the mediator
+ * @returns {Promise<void>}
+ */
+const setDeviceInfo = async (agent: Agent<any>, blankDeviceToken = false): Promise<void> => {
+  let token
+  if (blankDeviceToken) token = ''
+  else token = await messaging().getToken()
```
bryce-mcmath: Same as above

app/src/modals/PushNotificationsModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+          <Text style={[TextTheme.modalNormal]}>{'    \u2022 ' + t('PushNotifications.BulletOne')} </Text>
+          <Text style={[TextTheme.modalNormal]}>{'    \u2022 ' + t('PushNotifications.BulletTwo')} </Text>
+          <Text style={[TextTheme.modalNormal]}>{'    \u2022 ' + t('PushNotifications.BulletThree')} </Text>
+          <Text style={[TextTheme.modalNormal]}>{'    \u2022 ' + t('PushNotifications.BulletFour')} </Text>
+        </ScrollView>
+        <View style={[styles.controlsContainer]}>
+          <View style={styles.buttonContainer}>
+            <Button
+              title={t('CameraDisclosure.Allow')}
+              accessibilityLabel={t('CameraDisclosure.Allow')}
```
bryce-mcmath: We had an issue with Apple not liking us using 'Allow' for buttons related to permissions, we swapped it out with 'Continue' instead. We should probably do the same here.

app/src/screens/Developer.tsx
```diff
@@ -166,6 +171,30 @@ const Settings: React.FC = () => {
...
+  }
+
+  const initializePushNotificationsToggle = async () => {
+    setEnablePushNotifications(await PushNotificationHelper.isEnabled())
+  }
+
+  const toggleDevPushNotificationsSwitch = () => {
+    if (!pushNotificationCapable || !agent) return
+    if (enablePushNotifications) PushNotificationHelper.setDeviceInfo(agent, true)
+    else PushNotificationHelper.setup(agent)
```
bryce-mcmath: Gonna stop pointing these out because there's a lot of them but I think generally in Bifold and BC Wallet we tend to prefer bracketed `if/else` statements that span lines. Makes it easier to read if it's consistent.

bryce-mcmath (changes requested):
Some small things, overall seems fairly sturdy. I'll check out the branch and test locally later and see if I find any other issues.

app/src/screens/Developer.tsx
jamshale: Thanks for the review. No problem. Will update the syntax.

jamshale:
I can give access to my firebase account if anyone wants to test with the mediator plugin. Or just send the service account file. It's just on free tier, so it's not much of a concern giving it out.

bryce-mcmath (approved):
Looks good to me. @jleach @wadeking98 @cvarjao how do we feel about merging this?

PR #1427, opened 2023-09-05
feat: remote oca

wadeking98:
- Added support for remote oca
- Removed hardcoded oca bundles and assets. 
  

After merging this PR you will have to add `OCA_URL` environment variable to your .env file.
Right now you should use `OCA_URL=https://raw.githubusercontent.com/wadeking98/aries-oca-bundles/fix-script-trailing-comma/` because there's a bug in the main branch of the ares-oca-bundles repo that I'm fixing. Soon you'll be able to use the official repo which is: `https://raw.githubusercontent.com/bcgov/aries-oca-bundles/main/`


bryce-mcmath (approved):
Looks good. Not sure why those Bifold changes are being included in the diff when they are already in main in this repo

PR #1472, opened 2023-09-18
feat: revocation improvement

### LINKED ISSUE #1439: Proof request templates should not accepts revoked credentials
### Description of problem
Proof request templates in mobile verifier allows holders to share information from revoked credentials

### Expected behavior
For now, proof request templates should not accept revoked credentials. 

We want to go through all the proof templates that relate to LSBC, ACM or OK Corrections and make sure the templates ask for non-revoked credentials only.

### Steps to reproduce
- Have a revoked credential
- Receive a proof request
- Send information over

### Environment
- [X] Occurs on Android
- [X] Occurs on iOS

Build #: 1193

Android Device Model: 

iOS Device Model: 

### Severity
- [X] High
- [ ] Medium
- [ ] Low

wadeking98:
Added revocation intervals to proof templates to ensure that mobile verifiers don't accept revoked credentials.
Depends on: https://github.com/hyperledger/aries-mobile-agent-react-native/pull/972
Resolves: https://github.com/bcgov/bc-wallet-mobile/issues/1439

### DIFF
--- a/app/src/request-templates.ts
+++ b/app/src/request-templates.ts
@@ -66,7 +66,11 @@ export const useProofRequestTemplates = (useDevRestrictions: boolean) => {
           {
             schema: personSchema,
             requestedAttributes: [
-              { names: ['given_names', 'family_name', 'picture'], restrictions: verifiedPersonRestrictions },
+              {
+                names: ['given_names', 'family_name', 'picture'],
+                restrictions: verifiedPersonRestrictions,
+                nonRevoked: { to: 1694794371 },
+              },
             ],
           },
           {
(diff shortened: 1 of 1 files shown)


app/src/request-templates.ts
```diff
@@ -66,7 +68,11 @@ export const useProofRequestTemplates = (useDevRestrictions: boolean) => {
           {
             schema: personSchema,
             requestedAttributes: [
-              { names: ['given_names', 'family_name', 'picture'], restrictions: verifiedPersonRestrictions },
+              {
+                names: ['given_names', 'family_name', 'picture'],
+                restrictions: verifiedPersonRestrictions,
+                nonRevoked: { to: now() },
```
bryce-mcmath, line 74: If a user has first opened the wallet and then afterwards a credential is revoked, will it still be accepted because it's after the `now()`?

wadeking98: now() gets called whenever the template is fetched, so any time the QR code is refreshed the time will be updated. I've tested the changes and it seems to be working as expected, if a credential is revoked before receiving a proof then it will not be valid. I've added the functionality so support this in my bifold PR.

bryce-mcmath (approved):

PR #1507, opened 2023-09-28
feat: add initial support for app attestation

jleach:
The third and final component to the application attestation mechanics is the logic in the BC Wallet to bridge the gap between the RN module and the attestation controller (via traction).

This PR adds support to interface with the [RN module @hyperledger/aries-react-native-attestation](https://www.npmjs.com/package/@hyperledger/aries-react-native-attestation) in the Bifold repo and use it to communicate over DIDComm basic messages to a controller that supports attestation.


jleach:
Converted back to draft. Few conflicts to fix first.

app/ios/Podfile.lock
```diff
@@ -362,6 +362,8 @@ PODS:
   - React-jsinspector (0.66.5)
   - React-logger (0.66.5):
     - glog
+  - "react-native-attestation (1.0.0-alpha.99+1f2c467)":
```
bryce-mcmath, line 448: What is this `+hash` at the end of the version?

bryce-mcmath (dismissed):
Clean! LGTM

app/ios/Podfile.lock
jleach: Just something added by yarn so I'll leave it.

cvarjao: It is actually a lerna thing:
https://github.com/lerna/lerna/tree/main/libs/commands/publish#--canary

bryce-mcmath (approved):

PR #1511, opened 2023-09-30
fix: app crash on install/startup on iOS < 14

cvarjao:
Reference:
https://stackoverflow.com/a/69546401


nodlesh:
This has potentially solved an iOS 14 and below crash issue and, when running in the test pipeline, fixed most of the skipped test issues. For example, we usually have 3-7 tests get skipped for weird reasons that I and Sauce Labs have been trying to diagnose, but this new build has ran for 2 regression runs with only 1 skipped test. 
I say potentially here because we need more runs to establish the pattern.

jleach (approved):

PR #1517, opened 2023-10-04
chore: re-enable ledger update

### LINKED ISSUE #1521: Fix issue with outdated nodes in genesis blocks
Quite a few IPs are dead in the current genesis blocks. We need to update them to prevent unnecessary delays when using the BC Wallet.

- Fix issue with outdated nodes in genesis blocks.
- Enable genesis block build in CICD pipeline.

jleach:
With the switch to AFJ 4 we had to disable our automatic ldeger updates where we refresh the genesis block to reflect any recent node changes. This PR re-enables this feature.

Ref:

hyperledger/indy-did-networks#3
hyperledger/aries-mobile-agent-react-native#987

Fixed #1521

### DIFF
--- a/.github/workflows/actions/update-ledgers/action.yml
+++ b/.github/workflows/actions/update-ledgers/action.yml
@@ -4,6 +4,7 @@ runs:
   using: composite
   steps:
     - name: Update ledgers
+      shell: bash
       run: |
         node bifold/scripts/make-blocks.js
 
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -118,9 +118,8 @@ jobs:
           yarn install --immutable && \
           git status
 
-      # Ledgers are missing namespace, not compatible.
-      # - name: Update ledgers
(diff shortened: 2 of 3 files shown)


jleach:
This PR is in Draft until the Bifold PR noted in the description is merged.

bryce-mcmath (dismissed):

bryce-mcmath:
@cvarjao do you see any issues with the yarn lock file changes here?

jleach:
> @cvarjao do you see any issues with the yarn lock file changes here?

I was getting issues with from yarn install saying the lock file was out of sync so I had to update it:
` YN0028: │ The lockfile would have been modified by this install, which is explicitly forbidden.`

cvarjao (approved):
I don't see any issues. But we might need to do the same dependency alignment we did for bifold.  But I don't see any red flags. You updated the bifold submodule and it is expected it would change with the new changes to bifold (dependency alignment)

PR #1526, opened 2023-10-06
fix: lock ruby environment

jleach:
Lock the ruby environment similar to the changes made in hyperledger/aries-mobile-agent-react-native#991


.github/workflows/main.yaml
```diff
@@ -66,19 +68,10 @@ jobs:
       - name: Setup NodeJS
         uses: ./.github/workflows/actions/setup-node
 
-      - name: Setup Java
```
bryce-mcmath, line 69: Why are we getting rid of this step? I thought it was recently added by @cvarjao

jleach: ya, iOS does not use Java for anything so its unnecessary.

cvarjao: if we don't need it, I am fine with taking it out

bryce-mcmath (approved):

PR #1576, opened 2023-10-30
feat: refresh pool on connect

wadeking98:
Added a patch to refresh the pool when connecting, solves the slowness issue with lsbc credential
together with: https://github.com/hyperledger/aries-mobile-agent-react-native/pull/1015 this should speed up the wallet quite a bit


jleach:
@wadeking98 I will test this and see what it looks like at the firewall level where I can monitor network connections. You might be able to do this also on your linux laptop with wireshark - not sure. I think Wade B made the comment we probably only want to rebuild / reconcile the pool infrequently. He was thinking no more than every 24h. But, if this is not an overly slow process we can keep it as an interim step.

PR #1579, opened 2023-10-31
feat: improved performance on pool refresh

wadeking98:
Updated splash screen to refresh ledgers on startup.
Added patch to AFJ to allow manually refreshing each of the pools asynchronously.

### DIFF
--- a/.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch
+++ b/.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch
@@ -38,19 +38,19 @@ index 0ec57415429a0b97844b2d65a8ee83f68ad10466..56dc4c3702a0a540fda7cf1f97c2233f
          if (!this._pool)
              this.connect();
 diff --git a/build/pool/IndyVdrPoolService.d.ts b/build/pool/IndyVdrPoolService.d.ts
-index a21dca227076c7097bea4409233bf547eb229f11..fbe578777746097269aeaccb16c0cdf73c4f7b43 100644
+index a21dca227076c7097bea4409233bf547eb229f11..432a2dd842076776b90b0cce08e658054fd677fe 100644
 --- a/build/pool/IndyVdrPoolService.d.ts
 +++ b/build/pool/IndyVdrPoolService.d.ts
 @@ -31,6 +31,7 @@ export declare class IndyVdrPoolService {
      }>;
      private getPoolForLegacyDid;
      private getSettledDidResponsesFromPools;
-+    public refreshPoolConnections(): Promise<void>;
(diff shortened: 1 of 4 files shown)


.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch
```diff
@@ -0,0 +1,68 @@
...
++++ b/build/pool/IndyVdrPoolService.js
+@@ -113,6 +113,12 @@ let IndyVdrPoolService = class IndyVdrPoolService {
+             successful,
+         };
+     }
++    /**
++     * Refresh the pool connections asynchronously
++     */
++    async refreshPoolConnections() {
++        return await Promise.all(this.pools.map(async (pool) => await pool.refreshConnection()))
```
cvarjao: any idea how many items will be spawn? are gonna run into some sort of too many connections/handles open simultaneously?
also, can the `map()` just return the promises?
```
Promise.all(this.pools.map((pool) => pool.refreshConnection()))
```

jleach: Same comment I added to AFJ:

May not be a problem here because we don't care about results but in `Promise.all()` will throw an error and continue if any one of the promises fails. If any others fail you may get an unhandled rejection error. Consider wrapping this in a try/catch block with a debug log message an/or use `Promise.allSettled()`. If you don't care about using catching a potential error `allSettled()` may still be better behaviour.

jleach (changes requested):

.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch
wadeking98: It will spawn one process per ledger in ledgers.json, so for bc wallet that's 7

wadeking98: got it, switched to `allSettled()`

cvarjao: that doesn't seem too bad

package.json
```diff
@@ -34,7 +34,7 @@
     "test": "cd app && yarn test",
     "pre-commit-lint:app": "cd app && yarn pre-commit-lint",
     "pre-commit-lint": "yarn pre-commit-lint:app",
-    "typecheck": "cd app && yarn typecheck"
```
cvarjao: why removing typecheck? we can switch to yarn workspace syntax though (no need to `cd`)

wadeking98: my bad, on my main machine it fails locally so I had it temporarily disabled, must have accidentally committed the change

jleach (approved):

PR #1592, opened 2023-11-08
Updated DEVELOPER.md documentation

esune:
Updated documentation in `DEVELOPER.md`:
- Added required dependencies (Yarn, JDK, Gradle) and tips for installing `cocoapods`
- Revisited steps to add custom ledgers to the project

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -19,6 +19,7 @@ Before you can proceed with building and testing the BC Wallet app, you must ins
   > **Tip**: use [nvm](https://github.com/nvm-sh/nvm) to install node & npm. It helps to easily switch node & npm
   > version for each project.
 - [Yarn](https://yarnpkg.com/)
+- A [JDK](https://openjdk.org/) and [Gradle](https://gradle.org) - Make sure you install compatible versions, see [here](https://docs.gradle.org/current/userguide/compatibility.html) for more information.
 
 ## Prerequisite services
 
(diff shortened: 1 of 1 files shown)


DEVELOPER.md
```diff
@@ -49,7 +51,9 @@ git submodule update --init
 
 React Native environment setup instructions are documented [here](https://reactnative.dev/docs/environment-setup). Be sure to select the correct React Native version (currently 0.66.x) from the dropdown. This will guide you through setting up your development environment for your operating system and choice of iOS (only if you are using a Mac) or Android.
 
-Following along, you should end up using Android SDK Platform 30 with Android 11 (API Level 30) for Android development and/or iOS 11 for iOS development.
+Following along, you should end up using Android SDK Platform 33 with Android 13 (API Level 33) for Android development and/or iOS 11 for iOS development.
+
+**Notes:** you may need to install `cocoapods` using `brew` rather than Ruby's package manager on MacOS 14.x if you encounter installation errors.
```
cvarjao: I would recommend using `bundle install ...` as we are using in the [GHA workflow](https://github.com/bcgov/bc-wallet-mobile/blob/main/.github/workflows/main.yaml#L125C12-L125C12). Installing at system level might break existing dev environment

esune: Good tip, I wish I had known when I went through the setup process - I'll try it out and update accordingly if it works as expected.

esune: I refreshed the instructions: the installation of `cocoapods` now seems to succeed by following the instructions in the react native/Cocoapods getting started, however I had to add explicit version dependencies on Ruby and Bundler to resolve this. Additionally, it looks like the version of Ruby being used (2.7.8) is past its end-of-life: maybe something to track in an issue? Only installing pods as in the GHA does not work so the pre-requisites still apply.

bryce-mcmath (approved):

PR #1594, opened 2023-11-08
feat: enable and configure push notifications

bryce-mcmath:
Mostly config changes, plus some native code changes to remove app badge counter on load on both iOS and Android. Also an update to find the most recent mediator rather than the mediator that just matches the .env label (would result in errors if multiple mediator connections existed with the same label.)


jleach (approved):
LGTM. Is it ok to check `google-services.json` into SCM? I found [this](https://stackoverflow.com/questions/37358340/should-i-add-the-google-services-json-from-firebase-to-my-repository) which indicates it is ok. Thoughts?

bryce-mcmath:
> LGTM. Is it ok to check `google-services.json` into SCM? I found [this](https://stackoverflow.com/questions/37358340/should-i-add-the-google-services-json-from-firebase-to-my-repository) which indicates it is ok. Thoughts?

Yes, [here](https://groups.google.com/g/firebase-talk/c/bamCgTDajkw/m/uVEJXjtiBwAJ?pli=1) is a Firebase engineer saying those files are fine to check in. Same goes for the iOS equivalent. The info in these files only act as an identifier, the only sensitive info is needed in the mediator which shouldn't be (and isn't) checked in.

PR #1596, opened 2023-11-09
chore: upgrade node version used in action

bryce-mcmath:
Updating the Node version and GHA so that the builds don't fail


bryce-mcmath:
@jleach @wadeking98 Shel ran the smoke tests with this build, they all passed so I think we're good to merge

wadeking98 (dismissed):

jleach (approved):

PR #1613, opened 2023-11-20
feat: added ability to cache genesis trans for 24h

wadeking98:
implemented caching for ledger genesis transactions. 
The wallet will refresh the cache every 24 hours on load.
if the cache needs to be refreshes the wallet will refresh the pool and fetch the new genesis transactions. doing this also automatically updates the current agent transaction pool. The transactions are saved in local storage so we don't have to fetch them the next time we startup.

Note: this change also seems to improve performance with sorvin:staging network but more testing is needed

The following gifs were created with a local production build of the wallet:
## Accept LSBC credential:
lsbc-credential-accept

## a2a proof request:
a2a-dev-proof


package.json
```diff
@@ -59,6 +60,6 @@
     "@aries-framework/react-native@0.4.0": "patch:@aries-framework/react-native@npm:0.3.3#./.yarn/patches/@aries-framework-react-native-npm-0.3.3-bb75ece22d.patch",
     "@aries-framework/anoncreds@^0.4.0": "patch:@aries-framework/anoncreds@npm%3A0.4.0#./.yarn/patches/@aries-framework-anoncreds-npm-0.4.0-4d3b4e769d.patch",
     "@aries-framework/anoncreds@0.4.0": "patch:@aries-framework/anoncreds@npm%3A0.4.0#./.yarn/patches/@aries-framework-anoncreds-npm-0.4.0-4d3b4e769d.patch",
-    "@aries-framework/indy-vdr@^0.4.0": "patch:@aries-framework/indy-vdr@npm%3A0.4.0#./.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch"
+    "@aries-framework/indy-vdr@^0.4.0": "patch:@aries-framework/indy-vdr@patch%3A@aries-framework/indy-vdr@npm%253A0.4.0%23./.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch%3A%3Aversion=0.4.0&hash=0e7997&locator=bc-wallet-mobile%2540workspace%253A.#./.yarn/patches/@aries-framework-indy-vdr-patch-28d47565c5.patch"
```
bryce-mcmath: This patch looks different than all the rest, is there something special about it? Maybe we need to delete the existing patch and then add the new one back afterwards?

bryce-mcmath (commented):
Something seems odd about the one patch, everything else LGTM

package.json
wadeking98: resolved

cvarjao (approved):
LGTM

PR #1614, opened 2023-11-22
feat: implement new scans screen 

jleach:
This PR implements the new scan workflow and screens in the BC Wallet by updating the localization and constants.

### DIFF
--- a/app/src/constants.ts
+++ b/app/src/constants.ts
@@ -1,3 +1,4 @@
 export const surveyMonkeyUrl = 'https://www.surveymonkey.com/r/7BMHJL8'
 export const surveyMonkeyExitUrl = 'https://www.surveymonkey.com/survey-thanks'
+export const whereToUseWalletUrl = 'https://www2.gov.bc.ca/gov/content/governments/government-id/bc-wallet#where'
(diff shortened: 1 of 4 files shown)


app/src/constants.ts
```diff
@@ -1,3 +1,4 @@
 export const surveyMonkeyUrl = 'https://www.surveymonkey.com/r/7BMHJL8'
 export const surveyMonkeyExitUrl = 'https://www.surveymonkey.com/survey-thanks'
+export const whereToUseWalletUrl = 'https://www2.gov.bc.ca/gov/content/governments/government-id/bc-wallet#where'
```
bryce-mcmath, line 3: Does this really get picked up by the ScanHelp screen or are you adding this for use later?

bryce-mcmath (dismissed):
Just a comment, looks good

app/src/constants.ts
jleach: Tested. Nope, not exposed. I've created a ticket to fix the problem because I'm not sure how to go about fixing it and want to think through it a bit before jumping into a solution #1615.

bryce-mcmath (approved):

PR #1646, opened 2023-12-01
Added proper direct domains to sauce tunnel in build pipeline smoke test

nodlesh:
This PR updates the direct domains on the Sauce Tunnel. This needed to be changed a while ago and was in the test pipeline but the build pipeline was overlooked. This causes the iOS smoke test to fail on an iOS build.


bryce-mcmath:
@jleach any idea why the check-secrets and check-vars steps are failing on this change? I can't see how those steps would be affected by this one-liner

bryce-mcmath (approved):
Not sure why the secrets and vars steps are acting up, but LGTM

PR #1647, opened 2023-12-02
feat: use published bifold

jleach:
Adapt BC Wallet to use the `@hyperledger/aries-bifold-core` packages published to NPM.


package.json
```diff
@@ -59,7 +56,6 @@
     "@aries-framework/react-hooks@^0.4.2": "patch:@aries-framework/react-hooks@npm:0.4.2#./.yarn/patches/@aries-framework-react-hooks-npm-0.4.2-84b7eb8764.patch",
     "@aries-framework/react-native@0.4.0": "patch:@aries-framework/react-native@npm:0.3.3#./.yarn/patches/@aries-framework-react-native-npm-0.3.3-bb75ece22d.patch",
     "@aries-framework/anoncreds@^0.4.0": "patch:@aries-framework/anoncreds@npm%3A0.4.0#./.yarn/patches/@aries-framework-anoncreds-npm-0.4.0-4d3b4e769d.patch",
-    "@aries-framework/anoncreds@0.4.0": "patch:@aries-framework/anoncreds@npm%3A0.4.0#./.yarn/patches/@aries-framework-anoncreds-npm-0.4.0-4d3b4e769d.patch",
-    "@aries-framework/indy-vdr@^0.4.0": "patch:@aries-framework/indy-vdr@npm%3A0.4.0#./.yarn/patches/@aries-framework-indy-vdr-npm-0.4.0-b01067562c.patch"
```
cvarjao, line 63: is this being removed by accident?

jleach: I'm not sure what happened here. I still see `@aries-framework/anoncreds@0.4.0` and added VDR back.

package.json
```diff
@@ -55,6 +52,13 @@
...
+    "@aries-framework/anoncreds": "0.4.0",
+    "@aries-framework/anoncreds-rs": "0.4.0",
+    "@aries-framework/askar": "0.4.0",
+    "@aries-framework/core": "0.4.0",
+    "@aries-framework/indy-sdk-to-askar-migration": "0.4.0",
+    "@aries-framework/indy-vdr": "0.4.0",
+    "@aries-framework/react-native": "0.4.0",
     "@aries-framework/core@0.3.3": "patch:@aries-framework/core@npm:0.3.3#./.yarn/patches/@aries-framework-core-npm-0.3.3-dd6486de3d.patch",
     "@aries-framework/react-hooks@^0.4.2": "patch:@aries-framework/react-hooks@npm:0.4.2#./.yarn/patches/@aries-framework-react-hooks-npm-0.4.2-84b7eb8764.patch",
     "@aries-framework/react-native@0.4.0": "patch:@aries-framework/react-native@npm:0.3.3#./.yarn/patches/@aries-framework-react-native-npm-0.3.3-bb75ece22d.patch",
```
bryce-mcmath: Is there no issue with having two resolutions for different versions of the same dep? ie. core 0.3.3 vs core 0.4.0, react-native 0.3.3 vs 0.4.0

cvarjao: Agreed. line 58 (`"@aries-framework/core": "0.4.0"`) and line 62 (`"@aries-framework/core@0.3.3": "patch:@..."`) are conflicting with each other. That means the patch won't apply to core 0.4.0

jleach: Ok. I'll remove 0.3.3 since we don't use it. I though those lines meant that if you were installing v0.3.3 do one thing, if you were installing 0.4.0 do something different.

package.json
wadeking98: there were two anoncreds patches here, I believe one was the patch to the revocation interval that jason made, The other was a speed improvement from me when looking up the revocation registry for multiple credentials. Have the patches been combined or was one accidentally deleted?

wadeking98:
Other than my one comment LGTM

package.json
bryce-mcmath: they appear to have been duplicates, looking at the hash

wadeking98 (approved):

PR #1662, opened 2023-12-07
chore: make pod and package.json versions more deterministic

bryce-mcmath:
Clecio and I found that having ranges on npm package versions in `package.json` was causing the `pod install` to inconsistently generate pod versions in `Podfile.lock`. I've removed the carets from our npm versions and updated relevant versions so that no podfile versions changed


yarn.lock
```diff
@@ -3705,19 +4216,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@hyperledger/indy-vdr-react-native@patch:@hyperledger/indy-vdr-react-native@npm:0.1.0#./.yarn/patches/@hyperledger-indy-vdr-react-native-npm-0.1.0-10c696b320.patch::locator=bc-wallet-mobile%40workspace%3A.":
```
wadeking98: seems like the patch is deleted here

app/package.json
```diff
@@ -58,109 +58,109 @@
...
-    "@hyperledger/aries-oca": "1.0.0-alpha.150",
-    "@hyperledger/aries-react-native-attestation": "1.0.0-alpha.150",
-    "@hyperledger/indy-vdr-react-native": "^0.1.0",
+    "@hyperledger/anoncreds-react-native": "0.1.0",
+    "@hyperledger/aries-askar-react-native": "0.1.1",
+    "@hyperledger/aries-bifold-core": "1.0.0-alpha.155",
+    "@hyperledger/aries-bifold-verifier": "1.0.0-alpha.155",
+    "@hyperledger/aries-oca": "1.0.0-alpha.155",
+    "@hyperledger/aries-react-native-attestation": "1.0.0-alpha.155",
+    "@hyperledger/indy-vdr-react-native": "0.1.0",
```
wadeking98, line 67: maybe changing indy-vdr-react-native back to `^0.1.0` would fix the issue?

wadeking98:
@bryce-mcmath The error you're seeing during initialization is because the patch isn't being applied to `@hyperledger/indy-vdr-react-native`

app/package.json
bryce-mcmath: instead I removed the caret from the resolution block in the top level package.json and that worked

yarn.lock
bryce-mcmath: it's been added back now after I removed the caret from the resolution block

wadeking98 (approved):

PR #1684, opened 2023-12-20
feat: add new proof requests for serving it right etc.

bryce-mcmath:


proof-templates.json
```diff
@@ -690,5 +690,299 @@
...
+                        {
+                           "schema_id": "M6dhuFj5UwbhWkSLmvYSPc:2:Person:1.2",
+                           "issuer_did": "M6dhuFj5UwbhWkSLmvYSPc"
+                        },
+                        {
+                           "schema_id": "L6ASjmDDbDH7yPL1t2yFj9:2:Person:1.2",
+                           "issuer_did": "L6ASjmDDbDH7yPL1t2yFj9"
+                        },
+                        {
+                           "schema_id": "4eCXHS79ykiMv2PoBxPK23:2:unverified_person:0.1.0",
```
wadeking98: I don't think the unverified person credential will work here. It doesn't have the picture attribute and I think it has a different attribute name for first and last names

proof-templates.json
```diff
@@ -690,5 +690,299 @@
...
```
wadeking98: same here

wadeking98 (commented):
just remove the unverified person from your dev restrictions and then it looks good to merge

wadeking98 (approved):

PR #1704, opened 2024-01-12
feat: remote logging

### LINKED ISSUE #1654: Add troubleshooting logging
Existing logging libraries we could integrate with:
- [winston](https://github.com/winstonjs/winston)
- [pino](https://www.npmjs.com/package/pino)
- [bunyan](https://www.npmjs.com/package/bunyan)

Acceptance Criteria
- Enable/Disable via dev mode
- If logging enabled, set collection level to "debug" level
- keep logging file to a reasonable small size
- a way to extract and send to us
-

Reference:
- [Example of integrating with Pino](https://github.com/cvarjao/afj-perf-test/blob/main/src/index.ts)
- [React Native FS](https://www.npmjs.com/package/react-native-fs#API)

jleach:
Incorporated the new remote logging package for local testing and update Bifold NPM package.

Fixes #1654


app/src/screens/Developer.tsx
```diff
@@ -237,6 +261,16 @@ const Settings: React.FC = () => {
 
   return (
     <SafeAreaView edges={['bottom', 'left', 'right']}>
+      <Modal
```
bryce-mcmath, line 264: Just to confirm my understanding, this modal will only appear in developer settings when they tap the logging option, yeah?

bryce-mcmath (dismissed):
Just a question but otherwise looks good to me! Nice work on this one.

app/src/screens/Developer.tsx
jleach: Confirmed.

bryce-mcmath (dismissed):

bryce-mcmath (approved):

PR #1711, opened 2024-01-16
feat: trigger attestation via proof, fix developer setting ui bug

bryce-mcmath:
Notes for @jleach: This PR depends on the PR to the controller repo. To test once both are merged, do the usual webhook setup and run the controller, create a multi-use connection and put it in `attestation.tsx` as `attestationInviteUrl`. Then when you get a proof request that wants one of our attestation credentials, it will trigger the attestation flow. Sidenote, TS complained about sessionId being a number not a string so I changed the type and just called `toString` on it. We might want to do something more about that later.


app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
     }
   }
 
+  const handleProofs = async (proof: ProofExchangeRecord, agent: BifoldAgent): Promise<void> => {
+    try {
+      // 0. if the proof is just now being requested, handle. Otherwise do nothing
+      if (proof.state === ProofState.RequestReceived) {
```
jleach: How do you feel about using an "early return"? Rather than wrapping all the code in an "if" and indenting, just return if the condition is not met:

```javascript
if (proof.state != ProofState.RequestReceived) {
   return
}
```

I find they also help avoid Christmas trees of logic which can be error prone hand hard to follow:

```text
x
   x
      x
         x
      x
   x
x
```

app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
...
+        // 1. Is the proof requesting an attestation credential
+        const format = (await agent.proofs.getFormatData(proof.id)) as unknown as AttestationProofRequestFormat
+        const formatToUse = format.request?.anoncreds ? 'anoncreds' : 'indy'
+        const isRequestingAttestation = format.request?.[
+          formatToUse
+        ]?.requested_attributes?.attestationInfo?.restrictions?.some((rstr) =>
+          attestationCredDefIds.includes(rstr.cred_def_id)
+        )
+
+        if (isRequestingAttestation) {
```
jleach: Another early return opportunity?

app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
...
+          const credentials = await agent.credentials.getAll()
+          const availableAttestationCredentials = credentials.filter((record) => {
+            const credDefId = record.metadata.get(AnonCredsCredentialMetadataKey)?.credentialDefinitionId
+            return credDefId && attestationCredDefIds.includes(credDefId)
+          })
+
+          // 3. If no, start attestation flow by requesting a nonce
+          if (availableAttestationCredentials.length < 1) {
+            // change this URL to a multi use connection from your traction instance for testing
+            const attestationInviteUrl =
```
jleach: Can we move this to the top of the file for easier access?

app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
...
+              'https://traction-acapy-dev.apps.silver.devops.gov.bc.ca?c_i=eyJAdHlwZSI6ICJodHRwczovL2RpZGNvbW0ub3JnL2Nvbm5lY3Rpb25zLzEuMC9pbnZpdGF0aW9uIiwgIkBpZCI6ICI3YjNhMGE5Yi05YzBiLTRjYmUtODRlZC05Y2MwNmEyNmE0ZjYiLCAibGFiZWwiOiAiYnJ5Y2VtY21hdGgiLCAicmVjaXBpZW50S2V5cyI6IFsiMnlKMW9WMVlXcDJGTGIyVGR0ZmU2M1lKVTVEb0dHcHZuc3FkeXVTU3NUQnEiXSwgInNlcnZpY2VFbmRwb2ludCI6ICJodHRwczovL3RyYWN0aW9uLWFjYXB5LWRldi5hcHBzLnNpbHZlci5kZXZvcHMuZ292LmJjLmNhIn0='
+            const invite = await agent.oob.parseInvitation(attestationInviteUrl)
+
+            if (!invite) {
+              throw new Error('Bad invitation')
```
jleach: Do we need to throw a BifoldError with a human readable message and code?

app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
...
+            }
+
+            await removeExistingInvitationIfRequired(agent, invite.id)
+
+            const { connectionRecord } = await agent.oob.receiveInvitation(invite)
+
+            if (!connectionRecord) {
+              throw new Error('Cannot receive invitation')
```
jleach: BifoldError?

app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
...
+            }
+
+            const connectedRecord = await agent.connections.returnWhenIsConnected(connectionRecord.id)
+
+            if (!connectedRecord) {
+              throw new Error('Connection not active')
```
jleach: BifoldError?

app/src/services/attestation.tsx
```diff
@@ -158,6 +210,87 @@ export const AttestationProvider: React.FC<AttestationProviderParams> = ({ child
     }
   }
 
+  const handleProofs = async (proof: ProofExchangeRecord, agent: BifoldAgent): Promise<void> => {
```
jleach, line 253: Pretty long `fn` would be a good candidate to break into smaller more surgical `fn` and call them.

jleach (changes requested):

app/src/services/attestation.tsx
```diff
@@ -39,6 +64,70 @@ type AttestationProviderParams = {
 
 export type EventListenerFn = (event: BaseEvent) => void
 
+const attestationCredDefIds = [
+  'J6LCm5Edi9Mi3ASZCqNC1A:3:CL:109799:dev-attestation',
+  'NxWbeuw8Y2ZBiTrGpcK7Tn:3:CL:48312:default',
+]
+
+// change this URL to a multi use connection from your traction instance for testing
+const attestationInviteUrl =
```
jleach, line 73: Should this find its way into .env or contstants?

jleach (approved):

app/src/services/attestation.tsx
bryce-mcmath: Eventually but not yet, cause you change it fairly frequently for development so it's nice to not have to rebuild the entire project when you do

PR #1719, opened 2024-01-18
feat: add attestation url env variable, update github action

bryce-mcmath:
The `ATTESTATION_INVITE_URL` will have to be added to the repo as a var


jleach (approved):

jleach:
@bryce-mcmath Just waiting until I get the schema published then we can merge this one with the correct URL.

jleach (approved):

PR #1765, opened 2024-02-01
fix: add buffer peer dependency

jleach:
Add the buffer package which is required as a peer dependency and update the Bifold packages.


yarn.lock
```diff
@@ -4151,7 +4151,7 @@ __metadata:
     react-native-keychain: ^8.1.1
     react-native-localize: ^2.2.4
     react-native-orientation-locker: "*"
-    react-native-permissions: ^3.8.4
+    react-native-permissions: ^4.0.1
```
bryce-mcmath, line 4154: Are we sure we want to go up a major version here?

jleach: That's not my change. It comes from Bifold. But I can test it on a fresh install tomorrow and know better.

bryce-mcmath (approved):

PR #1833, opened 2024-02-29
fix: issue with message pickup

### LINKED ISSUE #1767: The attestation controller should correctly populate VC fields
The schema has been decided on for the attestation credential (#1714) and now the controller should correctly populate these values.

```json
{
  "attributes": [
    "issue_date_dateint",
    "validation_method",
    "app_version",
    "app_vendor",
    "app_id",
    "operating_system",
    "operating_system_version"
  ],
  "schema_name": "app_attestation",
  "schema_version": "1.0"
}
```

Values to Use

- "issue_date_dateint". Set to the time the credential was issued.
- "validation_method". Enum that denotes if App Store Receipts (< iOS 14), Apple Attestation (>= iOS 14), or Google Integrity API was used.
- "app_version". The mobile version number of the application ie. v1.0.15-1543
- "app_vendor". The mobile application vender ie. "Government of British Columbia". This will not change for us and can be imported from constatns.
- "app_id". Use the bundle ID. ie. "ca.bc.gov.BCWallet"
- "operating_system". Enum that is either "ios" or "android" all lower case. This can be sent by mobile application but I think it is just as accurate and more simple to determine this based on the attestation technique (google vs apple).
- "operating_system_version". The current version of the mobile operating system. ie. 14.0.3

jleach:
This trial is to fix an issue where when the Wallet is abruptly killed and messages are sent to it while off-line, the messages are queued on the mediator but they are not delivered at once. Rather, each message is delivered one at a time requiring restarts in between.

This is a trial test to update to ADF 0.4.2 and add some logic to improve mediator message pickup. 

Fixes #1767


bryce-mcmath:
Should probably be fine but might want to watch out for patches no longer being applied with new versions of AFJ deps.

bryce-mcmath (approved):
Just one comment about patches potentially no longer being applied

jleach:
I did a quick check on one when I did `yarn install` and it looked like one was being applied. Will check the others out also and monitor.

swcurran:
From what I can tell, you are saying the wallet is not picking up queued messages from the mediator. I’m not an expert on the protocol, but AFAIK, this should just work in Credo/Bifold — when you connect and there are queued messages, you should get all of them in one or more batches.  Did you do something specific in BC Wallet or are you following the protocol?  

Pickup Protocol: https://github.com/hyperledger/aries-rfcs/tree/main/features/0212-pickup

If this was not “just working” — you might want to check with Credo/Indicio folks, just in case.

PR #1843, opened 2024-03-06
fix: prevent duplicate connections from being scanned

bryce-mcmath:
This PR prevents the camera from freezing when scanning an invalid QR code and also prevents duplicate connections from being made when scanning a valid QR twice


app/src/index.ts
```diff
@@ -77,7 +77,7 @@ const configuration: ConfigurationContext = {
   disableOnboardingSkip: true,
   useCustomNotifications: useNotifications,
   useAttestation: useAttestation,
-  enableUseMultUseInvitation: true,
+  enableUseMultUseInvitation: false,
```
jleach, line 80: Did I just recently see a PR go through that changed this value?

jleach (approved):

app/src/index.ts
bryce-mcmath: yes, that change needs to be reverted as it's part of the cause of the duplicate connection issue

PR #1848, opened 2024-03-08
chore: smoke test in parallel

### LINKED ISSUE #1563: Refactor Test Pipeline for iOS and Android runs in Parallel
We will be getting an upgraded parallel Sauce Labs license. We would need to restructure the test pipeline to run iOS and Android tests in parallel instead of in series as it is now. 

It may also be wise to take a look at the build pipeline and see if we can't parallel iOS and Android smoke tests. However that depends on if the app builds are done in parallel as well.

jleach:
Our account should now support parallel builds in sauce labs. This PR makes the appropriate adjustment to run both iOS and Android smoke tests simultaneously. 

Fixes #1563

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -407,7 +407,7 @@ jobs:
     needs: [ship-to-saucelabs]
     runs-on: ubuntu-22.04
     strategy:
-      max-parallel: 1
+      max-parallel: 2
       fail-fast: false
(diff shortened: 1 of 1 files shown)


nodlesh (approved):
This doesn't completely fix #1563, That ticket also mentions paralleling the nightly regression tests as well, but I can cover that when I get a chance.

PR #1857, opened 2024-03-12
feat: implement new remote logging features and screen, add tests

bryce-mcmath:
This PR is to update the remote logging screen and add some new features like the banner on the home screen and the additional session ID area under the developer setting. Here's what it looks like:
remote_logging_updates

### DIFF
--- a/app/src/components/ErrorTextBox.tsx
+++ b/app/src/components/ErrorTextBox.tsx
@@ -1,0 +1,51 @@
+import { useTheme } from '@hyperledger/aries-bifold-core'
+import React from 'react'
+import { StyleSheet, Text, View } from 'react-native'
+import Icon from 'react-native-vector-icons/MaterialIcons'
+
+interface ErrorTextBoxProps {
+  children: React.ReactElement | string
+}
+
+const iconSize = 30
+const offset = 10
+
+const ErrorTextBox: React.FC<ErrorTextBoxProps> = ({ children }) => {
+  const { ColorPallet, TextTheme } = useTheme()
+  const styles = StyleSheet.create({
+    container: {
+      backgroundColor: ColorPallet.notification.error,
+      borderRadius: 5,
+      borderWidth: 1,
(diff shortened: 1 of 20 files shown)


app/src/components/HomeHeaderView.tsx
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  return logger?.remoteLoggingEnabled ? (
+    <Pressable onPress={onPressBanner} testID={testIdWithKey('SessionIdBanner')}>
+      <View style={styles.banner}>
+        <Text style={styles.bannerText}>{t('RemoteLogging.Banner', { sessionId: logger.sessionId.toString() })}</Text>
+      </View>
+    </Pressable>
+  ) : (
+    <View></View>
```
jleach: null?

app/src/components/ErrorTextBox.tsx
```diff
@@ -0,0 +1,51 @@
+import { useTheme } from '@hyperledger/aries-bifold-core'
```
jleach, line 1: I think we have a component called `InfoTextBox` that takes in an option to be Info, Warning, or Error. Can we-reuse that component rather than creating ErrorTextBox? I think it appears at the top of the Terms screen.

jleach (changes requested):
LMK if the InfoBox can be reused.

app/src/components/ErrorTextBox.tsx
bryce-mcmath: There is a component called InfoBox which is configurable but it is made for modal use and doesn't work for this use case. There is another component from Bifold called InfoTextBox which is a bit similar but it is not configurable and doesn't have quite the same positioning style as in the wireframes.

app/src/components/HomeHeaderView.tsx
bryce-mcmath: fixed

jleach (approved):

PR #1858, opened 2024-03-13
feat: added Push notificaiton screen

wadeking98:
Added push notification screen in onboarding and settings menu.
Depends on: https://github.com/openwallet-foundation/bifold-wallet/pull/1111


app/src/helpers/PushNotificationsHelper.ts
```diff
@@ -27,34 +33,51 @@ const _foregroundHandler = (): (() => void) => {
...
+    case RESULTS.DENIED:
+      return NotificationPermissionStatus.DENIED
+    case RESULTS.BLOCKED:
+      return NotificationPermissionStatus.DENIED
+    default:
+      return NotificationPermissionStatus.UNKNOWN
+  }
+}
+
+const _requestPermission = async (): Promise<NotificationPermissionStatus> => {
```
jleach: I think _ is meant to be "private" but we don't use the pattern elsewhere in the wallet. Are we able to assume because its not exported its meant to be internal only?

app/src/index.ts
```diff
@@ -78,6 +83,18 @@ const configuration: ConfigurationContext = {
...
+    status: status,
+    setup: setup,
+    toggle: async (state: boolean, agent: Agent<any>) => {
+      if (state) {
+        await activate(agent)
+      } else {
+        await deactivate(agent)
+      }
+    },
+    check: async () => 'granted',
```
jleach: Should this be `NotificationPermissionStatus.GRANTED` (or `.Granted`)?

jleach (changes requested):

jleach:
@wadeking98 If push notifications are going GA do we need to remove the on-off capability from Developer settings?

app/src/index.ts
wadeking98: actually that's an artifact from an old change, I'll remove it

app/src/helpers/PushNotificationsHelper.ts
wadeking98: ya I didn't add the `_` but I can remove them. I'm assuming that's why they're there

app/ios/Podfile.lock
```diff
@@ -954,4 +954,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 557ce83fddb4b3974ada332a9ba0251f0ffcc639
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.14.3
```
bryce-mcmath: I think Clecio wants us to stick with 1.13.0 cause that's what our Gemfile config and CICD uses

bryce-mcmath (changes requested):
Just cocoapods version change, other than that looks good

bryce-mcmath (approved):

app/ios/Podfile.lock
wadeking98: got it, I changed it back to 1.13

jleach (approved):

PR #1865, opened 2024-03-19
feat: fixed terms & cond buttons

wadeking98:
Quick fix to make terms and condition buttons fixed.
Removed checkbox from terms screen during setup

## Onboarding terms screen:
https://github.com/bcgov/bc-wallet-mobile/assets/36937407/d6a19d3e-7df2-4d38-ba7e-b8c06fc2d062

## App startup terms screen:
https://github.com/bcgov/bc-wallet-mobile/assets/36937407/53dfc69c-4066-421c-ad3d-992c43d14735


bryce-mcmath (changes requested):
Should `onSubmitPressed` always navigate to the CreatePIN screen? Even if they've already completed onboarding in the past?

wadeking98:
> Should `onSubmitPressed` always navigate to the CreatePIN screen? Even if they've already completed onboarding in the past?

I agree it's not the most readable, and I'm not 100% sure why this happens, but it seems like since onboarding is completed it will actually navigate to the splash screen instead of the next onboarding screen, and then the splash screen will take the user to the enter pin screen.

I could add a bit of logic to navigate to the enter pin screen. I think it would make it more readable, but it would essentially do the same thing

jleach:
Fixed in #1869.

PR #1868, opened 2024-03-19
feat: automatic attestation for person credential

jleach:
When retrieving a person credentnial app attestation is required. This PR adds mechanics to automatically fetch the attestation credential if it is not present and accept the proof request.

### DIFF
--- a/app/src/helpers/Attestation.ts
+++ b/app/src/helpers/Attestation.ts
@@ -127,6 +127,36 @@ export const getAvailableAttestationCredentials = async (
 }
 
 /**
+ * Check if existing credentials satisfy the proof request
+ *
+ * Detailed check if we have the necessary credentials to fulfill the
+ * proof request in the required format.
+ *
+ * @param agent The AFJ agent
+ * @param proof The proof request
+ * @param filterByNonRevocationRequirements Whether to filter by non-revocation requirements
+ * @returns Credentials that match the given proof request
+ * @throws {Error} Will throw an error if a problem looking up data occurs
+ */
+export const credentialsMatchForProof = async (
+  agent: BifoldAgent,
+  proof: ProofExchangeRecord,
+  filterByNonRevocationRequirements = true
+): Promise<GetCredentialsForProofRequestReturn> => {
+  const proofFormats = await formatForProofWithId(agent, proof.id, filterByNonRevocationRequirements)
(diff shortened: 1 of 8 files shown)


app/src/helpers/Attestation.ts
```diff
@@ -36,25 +41,157 @@ export interface AttestationProofRequestFormat {
...
+  // TODO:(jl) Should we be checking the length of the attributes matches some
+  // expected length in the proof request?
+
+  if (credentials.proofFormats.indy) {
+    return credentials.proofFormats.indy.attributes.attestationInfo.length == 0
+  }
+
+  if (credentials.proofFormats.anoncreds) {
+    return credentials.proofFormats.anoncreds.attributes.attestationInfo.length == 0
+  }
```
bryce-mcmath: should this be `!== <the number of attributes on our attestation credentials>`? Should be strict too I think (`!==` not `!=`)

bryce-mcmath (changes requested):
One initial change requested that we talked about yesterday. Other than that, does this logic handle both situations with the BCSC app? ie. does the person flow work both with and without attestation when attestation support is enabled in the wallet?

jleach:
> One initial change requested that we talked about yesterday. Other than that, does this logic handle both situations with the BCSC app? ie. does the person flow work both with and without attestation when attestation support is enabled in the wallet?

The reason for the additional logic is that Credo does not select the newest non-revoked credential which we would like in all situations. I created [this](https://github.com/orgs/bcgov/projects/108?pane=issue&itemId=57285368) to track revisiting the issue to make sure it still persists in Credo 0.4.x and 0.5.x.

The Person workflow does work in both cases although once IDIM has deployed the change they will not issue a Person credential without a correct response to the proof.

app/src/helpers/Attestation.ts
jleach: Will fix.

jleach: @bryce-mcmath I fixed the strict equality check. I think it should still be `===`. That credential will have no attributes if its missing. As part of a more advanced "walk" feature we should experiment to see if the proof is asking for 5 attributes then our fetched credential should have at least 5 attributes. I'm not sure if it works this way which is why I didn't add it now (probably needs a spike.

bryce-mcmath (approved):

PR #1869, opened 2024-03-20
fix: eula onboarding

wadeking98:
added fix for onboarding where there was a new eula screen and new push notification screen. Previously the Use Push notification screen would take the user to the create pin screen even if they had completed their onboarding. now the Use Push Notification screen does not try to navigate users to the create pin screen 

depends on: https://github.com/openwallet-foundation/bifold-wallet/pull/1117


app/src/screens/Terms.tsx
```diff
@@ -100,19 +105,15 @@ const Terms = () => {
...
-            agreedToPreviousTerms
-              ? 'Terms of use update'
-              : 'Please agree to the terms of use below before using this application.'
-          }
-        ></InfoBox>
+        {agreedToPreviousTerms && (
+          <InfoBox
+            notificationType={InfoBoxType.Info}
+            description={
+              'BC Wallet has updated the end user licencse agreement. To continue using BC Wallet, please reveiw and accept the updated terms.'
```
bryce-mcmath: typos: licenscse, reveiw

bryce-mcmath (changes requested):

bryce-mcmath (approved):

PR #1876, opened 2024-03-21
chore: disable analytics, add privacy declaration, bump version

bryce-mcmath:
Found this StackOverflow post regarding the recent notices from the App Store: https://stackoverflow.com/a/78168190

This led me to making a Privacy manifest following the recommendations here: https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api

You can see which reasons were chosen for each notice we received in the `PrivacyInfo.xcprivacy` file and compare with the other reason options at the above link.

Also in this PR is some iOS and Android specific settings to disable Firebase Analytics

### DIFF
--- a/app/ios/PrivacyInfo.xcprivacy
+++ b/app/ios/PrivacyInfo.xcprivacy
@@ -1,0 +1,41 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>0A2A.1</string>
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryDiskSpace</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>E174.1</string>
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>35F9.1</string>
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>CA92.1</string>
+			</array>
+		</dict>
+	</array>
+</dict>
+</plist>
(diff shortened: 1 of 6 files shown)


app/ios/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,41 @@
...
+			<array>
+				<string>0A2A.1</string>
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryDiskSpace</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>E174.1</string>
```
jleach, line 20: Resonable.

app/ios/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,41 @@
...
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategorySystemBootTime</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>35F9.1</string>
```
jleach, line 28: Resonable

app/ios/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,41 @@
...
+			</array>
+		</dict>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryUserDefaults</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>CA92.1</string>
```
jleach, line 36: Good

app/ios/PrivacyInfo.xcprivacy
```diff
@@ -0,0 +1,41 @@
...
+<plist version="1.0">
+<dict>
+	<key>NSPrivacyAccessedAPITypes</key>
+	<array>
+		<dict>
+			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>0A2A.1</string>
```
jleach: I think "C617.1" might be better. I read "0A2A.1" to mean that BC Wallet is itself a 3rd-party SDK which we're not. This is probably used by AFJ to store the tails file and other caching. It notes "This reason may only be declared by third-party SDKs." Thoughts?

[Ref Doc](https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api)

jleach (changes requested):

app/ios/PrivacyInfo.xcprivacy
cvarjao: I agree with using `C617.1`

app/ios/PrivacyInfo.xcprivacy
cvarjao: LGTM

app/ios/PrivacyInfo.xcprivacy
cvarjao: I am okay with this reason. I don't think it is something we use directly though. Seems like a side effect of using react native, indy-vdr, askar

app/ios/PrivacyInfo.xcprivacy
cvarjao: LGTM

app/ios/PrivacyInfo.xcprivacy
bryce-mcmath: Agree and fixed. I misunderstood this one

jleach (approved):

PR #1884, opened 2024-03-26
feat: fix accept proof

jleach:
Allow AFJ to automatically accept proof requests.

### DIFF
--- a/app/src/screens/PersonCredential.tsx
+++ b/app/src/screens/PersonCredential.tsx
@@ -100,12 +100,18 @@ export default function PersonCredential() {
 
     const acceptAttestationProofRequest = async () => {
       if (!attestationLoading && didStartAttestationWorkflow && remoteAgentConnectionId) {
+        // TODO:(jl) These proofs are hidden. If we find any stale ones we should remove
+        // them by declining them or deleting them.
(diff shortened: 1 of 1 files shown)


app/src/screens/PersonCredential.tsx
```diff
@@ -100,23 +100,29 @@ export default function PersonCredential() {
 
     const acceptAttestationProofRequest = async () => {
       if (!attestationLoading && didStartAttestationWorkflow && remoteAgentConnectionId) {
+        // TODO:(jl) These proofs are hidden. If we find any stale ones we should remove
+        // them by declining them or deleting them.
         const proofRequest = receivedProofRequests.find((proof) => proof.connectionId === remoteAgentConnectionId)
         if (proofRequest) {
-          const credentials = await credentialsMatchForProof(agent, proofRequest)
+          // This will throw if we don't have the necessary credentials
+          const credentials = await agent.proofs.selectCredentialsForRequest({
```
cvarjao, line 108: Is the fetching of new app attestation if one doesn't exist/match handled somewhere else?

jleach: Yes, that code is re-used from Bryce's work on fetching it for the proof request screen if it's ever asked for. We should not get to this part of the code unless we have an attestation credential.

cvarjao (approved):

PR #1890, opened 2024-03-28
fix: attestation workflow 2

jleach:
Update the person credential to use an alternate workflow where it waits for all attestation steps to complete before proceeding with BSCS authentication and person credential acceptance.


bryce-mcmath (approved):
Looks good. And this works when attestation is disabled from BCSC side?

PR #1891, opened 2024-03-28
feat: add timer and text

jleach:
Add a timeout waiting on the attestation proof request and some text to explain to the user what is going on.

### DIFF
--- a/app/src/screens/PersonCredential.tsx
+++ b/app/src/screens/PersonCredential.tsx
@@ -10,9 +10,10 @@ import {
   BifoldAgent,
   Screens,
   Stacks,
+  InfoTextBox,
 } from '@hyperledger/aries-bifold-core'
 import { useNavigation } from '@react-navigation/native'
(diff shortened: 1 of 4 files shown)


app/src/screens/PersonCredential.tsx
```diff
@@ -139,6 +143,14 @@ export default function PersonCredential() {
     // and the user needs to wait.
     setWorkflowInProgress(true)
 
+    timer.current = setTimeout(() => {
+      if (!didCompleteAttestationProofRequest) {
```
bryce-mcmath: Should this be checking if the proof request was completed or if it was recieved?

bryce-mcmath (commented):
question

app/src/screens/PersonCredential.tsx
jleach: Fixed.

bryce-mcmath (approved):

PR #1914, opened 2024-04-12
fix: bug switching to new verison of bc wallet

wadeking98:
Previously the call in container-imp was using a shallow copy with the initial state, this meant that the new onboarding state wouldn't be copied and would result in postAuthScreens and didCompleteOnboarding to be undefined. I changed the functionality to have a deeper copy and that fixed the issue

### DIFF
--- a/app/container-imp.ts
+++ b/app/container-imp.ts
@@ -67,12 +67,12 @@ export class AppContainer implements Container {
       ])
       const state: BCState = {
         ...initialState,
-        loginAttempt,
-        preferences,
-        migration,
-        tours,
-        onboarding,
-        dismissPersonCredentialOffer: personCredOfferDissmissed,
+        loginAttempt: {...initialState.loginAttempt, ...loginAttempt},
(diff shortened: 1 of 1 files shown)


jleach (approved):

app/container-imp.ts
```diff
@@ -67,12 +67,12 @@ export class AppContainer implements Container {
       ])
       const state: BCState = {
         ...initialState,
-        loginAttempt,
-        preferences,
-        migration,
-        tours,
-        onboarding,
-        dismissPersonCredentialOffer: personCredOfferDissmissed,
+        loginAttempt: {...initialState.loginAttempt, ...loginAttempt},
```
jleach, line 70: Using `{...}` to copy an object in JavaScript is good for a basic copy and works in simple cases. But, it only copies the first layer deeply. If you have objects or arrays inside, they're still linked to the original, which may not be what you want for a complete deep copy. If you have nested objects or arrays you may want `_.deepCopy()` or something similar.

PR #1945, opened 2024-04-22
feat: attestation logging

bryce-mcmath:
Additional logging to check timing of attestation as well as help troubleshoot if any issues arise


app/src/screens/PersonCredential.tsx
```diff
@@ -100,17 +99,13 @@ export default function PersonCredential() {
 
   // Use this function to accept the attestation proof request.
   const acceptAttestationProofRequest = async (agent: BifoldAgent, proofRequest: ProofExchangeRecord) => {
-    // Sanity check to make sure we have the necessary credentials
-    const credential = await getAvailableAttestationCredentials(agent)
```
wadeking98, line 104: should this be removed from attestation helpers? idk if it's being used anywhere else

wadeking98 (approved):

app/src/screens/PersonCredential.tsx
bryce-mcmath: Yeah there are a couple of changes we should make but I wanted to keep this PR to be only logging changes

PR #1958, opened 2024-04-25
chore: remove postinstall script and update docs

bryce-mcmath:
Fix to get bundle install working properly to avoid using local cocoapods version

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -78,8 +78,6 @@ jobs:
   build-ios:
     needs: [check-secrets, check-vars]
     runs-on: macos-13
-    env:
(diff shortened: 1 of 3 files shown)


jleach (dismissed):

cvarjao (changes requested):
Can we also align with GHA?
https://github.com/bcgov/bc-wallet-mobile/blob/main/.github/workflows/main.yaml#L138

I think we can use a script in the package.json so that we can use consistently, locally and GHA. Something like:
```
"ios:setup": "bundle install --froze && cd ios && bundle exec pod install"
```
And we can update the docs and GHA to use the same way, so we don't get out of sync

cvarjao (approved):
LGTM

PR #1979, opened 2024-05-09
chore: update bifold to 230

jleach:
Update the Bifold package to Alpha 230 to incorporate:

- Logger from dependency injection
- Remote OCA bundle resolution caching.


bryce-mcmath (approved):
That OCA change is clean, I like it

PR #1986, opened 2024-05-16
refactor: attestation support

jleach:
This refactor the attestation DRCP to consistently use request/response mechanics for communication, eliminating the need for out-of-band request.

Key changes:
1. **Nonce Handling**: The DRPC call now gets the nonce value from the response callback.
2. **Attestation Request**: The DRPC call now gets the attestation status and, if needed, an error message from the callback response.

Related to bcgov/mobile-attestation-vc-controller#49


app/src/hooks/useAttestation.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+export const AttestationContext = createContext<AttestationX>(null as unknown as AttestationX)
+
+type AttestationProviderParams = {
+  children: React.ReactNode
+}
+
+export const AttestationProvider: React.FC<AttestationProviderParams> = ({ children }) => {
+  const { agent } = useAgent()
+  // const { t } = useTranslation()
```
bryce-mcmath: nitpicky, but looks like we can safely remove

app/src/screens/Splash.tsx
```diff
@@ -406,6 +406,9 @@ const Splash = () => {
         setStep(6)
         await createLinkSecretIfRequired(newAgent)
 
+        // const am = new AttestationMonitor(agent!, logger, { attestationInviteUrl, attestationCredDefIds })
+        // container.registerInstance(TOKENS.UTIL_AGENT, { useValue: newAgent })
```
bryce-mcmath: can be safely removed?

app/src/services/attestation.ts
```diff
@@ -0,0 +1,430 @@
...
+      }
+    }
+  }
+}
+
+interface AttestationProofRequestFormat {
+  request: IndyRequest & AnonCredsRequest
+}
+
+// type t = (key: string | Array<string>, options?: object) => string
```
bryce-mcmath: safely removable?

app/src/services/attestation.ts
```diff
@@ -0,0 +1,430 @@
...
+
+    if (proof.state !== ProofState.RequestReceived) {
+      return
+    }
+
+    this.log?.info('Checking if proof is requesting attestation')
+
+    try {
+      // 1. Is the proof requesting an attestation credential
+      if (!(await isProofRequestingAttestation(proof, this.agent, this.options.attestationCredDefIds))) {
```
bryce-mcmath, line 243: Is this check necessary if it is repeated within `this.attestationCredentialRequired` on line 408?

bryce-mcmath (changes requested):
Couple small considerations, but really like the new pattern. Great PR

app/src/services/attestation.ts
jleach: They're a little different. One is checking if a recently arrived proof request is requesting attestation. If not it can safely be ignored. The second (408) is checking to see if we already have an attestation credential or not. If we have one we don't need to start the process of getting one - we can just use it. I think they're both needed.

bryce-mcmath: I think I worded that poorly. Within the `attestationCredentialRequired` method, on line 432, there's a second call of `isProofRequestingAttestation` that seems to also check if the same proof is requesting attestation, and also does an early return if it's not. I could be confused though. Not a big deal if it gets called an extra time anyway, it takes less than 50ms.

bryce-mcmath (dismissed):
Just one clarifying comment, but approved anyway

app/src/services/attestation.ts
jleach: I removed the second check.

bryce-mcmath (approved):

PR #1992, opened 2024-05-23
fix: add fully qualified did

jleach:
Testing if fully qualified DIDs will work in our proof request templates for the mobile verifier.


jleach:
This was for testing. Probably not needed.

PR #2019, opened 2024-06-18
feat: added indy-vdr caching to bc wallet

wadeking98:
added indy-vdr cache functionality to BC Wallet


app/package.json
```diff
@@ -61,12 +61,12 @@
...
-    "@hyperledger/aries-bifold-verifier": "1.0.0-alpha.248",
-    "@hyperledger/aries-oca": "1.0.0-alpha.248",
-    "@hyperledger/aries-react-native-attestation": "1.0.0-alpha.248",
-    "@hyperledger/indy-vdr-react-native": "0.2.2",
+    "@hyperledger/aries-bifold-core": "1.0.0-alpha.249",
+    "@hyperledger/aries-bifold-remote-logs": "1.0.0-alpha.249",
+    "@hyperledger/aries-bifold-verifier": "1.0.0-alpha.249",
+    "@hyperledger/aries-oca": "1.0.0-alpha.249",
+    "@hyperledger/aries-react-native-attestation": "1.0.0-alpha.249",
+    "@hyperledger/indy-vdr-react-native": "patch:@hyperledger/indy-vdr-react-native@npm%3A0.2.2#~/.yarn/patches/@hyperledger-indy-vdr-react-native-npm-0.2.2-627d424b96.patch",
```
bryce-mcmath: Should this be kept as it's usual version here and only changed to the patch in the top-level resolutions block? (Like was done for @credo-ts/core patch and others)

bryce-mcmath (changes requested):

app/package.json
wadeking98: Ya I could do that, this is just the way yarn did it but I'm fine doing it that way too

bryce-mcmath (approved):

PR #2030, opened 2024-06-25
fix: updated dev proof template restrictions to new showcase creds

wadeking98:


proof-templates.json
```diff
@@ -967,6 +967,32 @@
                            "issuer_did": "RGjWbW1eycP7FrMf4QJvX8"
                         }
                      ],
+                     "devRestrictions": [
```
bryce-mcmath: Does Kyle and co. want dev restrictions for this proof? When I first made it he told me not to add any

wadeking98: okay I'll remove it then

bryce-mcmath (approved):

PR #2031, opened 2024-06-25
feat: added support for oca branding on missing creds from bifold

wadeking98:
supporting PR for adding OCA branding to missing credentials in a proof request


app/container-imp.ts
```diff
@@ -41,6 +44,27 @@ export class AppContainer implements Container {
     // Here you can register any component to override components in core package
     // Example: Replacing button in core with custom button
     this.container.registerInstance(TOKENS.SCREEN_PREFACE, Preface)
+    this.container.registerInstance(TOKENS.CRED_HELP_ACTION_OVERRIDES, [
+      {
+        credDefIds: [
+          'RGjWbW1eycP7FrMf4QJvX8:3:CL:13:Person',
```
jleach, line 50: Thoughts on putting these into "constants"? We have a number of special DIDs / Schema ID / Cred Defn Id in the app now and maybe it's easier to put them all in once place for updates?

bryce-mcmath (changes requested):
Will also have to update the podfile, but looks good otherwise

app/container-imp.ts
wadeking98: IMHO I think dependency injection is a cleaner approach than having it in a constants file. Ideally I think we want to be migrating away from the useConfiguration hook or from the constants.ts file

bryce-mcmath (approved):

PR #2039, opened 2024-06-27
refactor: proof templates to DI

### LINKED ISSUE #2024: Move a couple of configurations to DI
Bryce will bake brownies for whoever takes this ticket if they also implement array / object destructuring for the container, so instead of:
```
const logger = container.resolve(TOKENS.UTIL_LOGGER)
const indyLedgers = container.resolve(TOKENS.UTIL_LEDGERS)
const ocaBundleResolver = container.resolve(TOKENS.UTIL_OCA_RESOLVER) as RemoteOCABundleResolver
```

we would use:
```
const { logger, indyLedgers, ocaBundleResolver } = container.resolve([TOKENS.UTIL_LOGGER, TOKENS.UTIL_LEDGERS, TOKENS.UTIL_OCA_RESOLVER])
```

jleach:
A little cleanup and refactoring of the dependency injection and moved the proof templates into dependency injection (DI).

Depends on openwallet-foundation/bifold-wallet#1181

Fixes #2024


app/src/screens/Splash.tsx
```diff
@@ -134,6 +134,11 @@ const Splash = () => {
   const logger = container.resolve(TOKENS.UTIL_LOGGER)
   const indyLedgers = container.resolve(TOKENS.UTIL_LEDGERS)
   const ocaBundleResolver = container.resolve(TOKENS.UTIL_OCA_RESOLVER) as RemoteOCABundleResolver
+  // const [logger, indyLedgers, ocaBundleResolver] = container.resolveSome([
```
bryce-mcmath: Can safely remove?

bryce-mcmath (changes requested):
Podfile.lock should be updated, and one comment to remove

app/src/screens/Splash.tsx
jleach: Done.

app/ios/Podfile.lock
```diff
@@ -980,4 +980,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: f043d17a36d946983917d5651311899450c41e12
 
-COCOAPODS: 1.13.0
+COCOAPODS: 1.15.2
```
bryce-mcmath: Should update with `yarn ios:setup` from `app/` so it uses the Gemfile and locked versions. Cocoapods version should be the same

bryce-mcmath: Rather than just `pod install` I mean

bryce-mcmath (changes requested):

bryce-mcmath (approved):

PR #2054, opened 2024-07-08
feat: persistent remote debugging

jleach:
Once enabled, allow remote debugging to persist through application restarts as long as the time limit has not expired.


app/container-imp.ts
bryce-mcmath, line 1: Is the initialization if it picks up the old session id, not going to restart the time until it's auto turned off? So if someone keeps logging into and out of the wallet it will continue remote logging well over the 3 hours or whatever the limit is? I might be missing something.

jleach: What how about I add this to the logger as a solution:

```TypeScript
 public overrideCurrentAutoDisableExpiration(expirationInMinutes: number) {
   if (expirationInMinutes <= 0) {
      return
   }

    if (this.remoteLoggingAutoDisableTimer) {
      clearTimeout(this.remoteLoggingAutoDisableTimer)
    }

    this.remoteLoggingAutoDisableTimer = setTimeout(() => {
      this.remoteLoggingEnabled = false
    }, expirationInMinutes * 60000)
  }
  ```

bryce-mcmath: I think that should work

bryce-mcmath:
Looks good but maybe we should hold until we get a release out? What do you think?

jleach:
> Looks good but maybe we should hold until we get a release out? What do you think?

There is no rush for this change. I'm good with holding off.

bryce-mcmath:
The four new issues seem like they would be easy to fix (and probably should have gotten picked up by the linter?)

bryce-mcmath (approved):

PR #2058, opened 2024-07-09
chore: add missing permission description for unused perm

bryce-mcmath:
This PR adds a description for an unused permission entry that triggered a warning from the App Store.

Reference: https://github.com/zoontek/react-native-permissions/issues/266#issuecomment-422443290

### DIFF
--- a/app/ios/AriesBifold/Info.plist
+++ b/app/ios/AriesBifold/Info.plist
@@ -63,7 +63,7 @@
 	<key>NSFaceIDUsageDescription</key>
 	<string>$(PRODUCT_NAME) wants to use your FaceID/TouchID to authenticate your identity</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string></string>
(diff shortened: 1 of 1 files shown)


app/ios/AriesBifold/Info.plist
```diff
@@ -63,7 +63,7 @@
 	<key>NSFaceIDUsageDescription</key>
 	<string>$(PRODUCT_NAME) wants to use your FaceID/TouchID to authenticate your identity</string>
 	<key>NSLocationWhenInUseUsageDescription</key>
-	<string></string>
+	<string>Our app does not request this permission or utilize this functionality but it is included in our Info.plist since our app utilizes the react-native-permissions library, which references this permission in its code.</string>
```
jleach: Kind of technical for a user facing explanation. How about:

```
We include this permission because a library our app uses references it, even though we don't use it directly.
```

jleach (changes requested):

app/ios/AriesBifold/Info.plist
bryce-mcmath: done

jleach (approved):

PR #2077, opened 2024-07-16
chore: update bifold packages

jleach:
Update Bifold package to 1.0.0-alpha.271.


cvarjao (dismissed):

bryce-mcmath (changes requested):
Should also update the `Podfile.lock` with `yarn ios:setup` from `/app`

bryce-mcmath (approved):

PR #2089, opened 2024-07-25
feat: proxy setup

bryce-mcmath:
Will need to add an env variable `INDY_VDR_PROXY_URL` and set it to the test URL. But I'll let this get reviewed first


jleach (dismissed):
LGTM. Do we need to add anything to the [docs](https://github.com/bcgov/bc-wallet-mobile/blob/main/DEVELOPER.md)?

app/src/screens/Developer.tsx
```diff
@@ -251,6 +252,14 @@ const Settings: React.FC = () => {
     setEnableShareableLink((previousState) => !previousState)
   }
 
+  const toggleEnableProxySwitch = () => {
+    dispatch({
+      type: BCDispatchAction.TOGGLE_PROXY,
+      payload: [!enableProxy],
+    })
+    setEnableProxy((previousState) => !previousState)
```
wadeking98, line 260: do we need to re-initialize the agent here or something so that the proxy gets activated/deactivated when toggled? or are we okay with restarting the app?

wadeking98 (dismissed):

app/src/screens/Developer.tsx
bryce-mcmath: I think for now we're okay with restarting the app

bryce-mcmath:
> LGTM. Do we need to add anything to the [docs](https://github.com/bcgov/bc-wallet-mobile/blob/main/DEVELOPER.md)?

Done, added to docs and sample env

jleach (approved):

PR #2094, opened 2024-07-26
feat: added person credential loading screen

wadeking98:
Added loading screen to person credential flow instead of button activity indicator

https://github.com/user-attachments/assets/bd4aaa00-cb15-4781-add0-510972102be0


app/src/components/PersonCredentialSpinner.tsx
```diff
@@ -0,0 +1,53 @@
+import { useTheme } from '@hyperledger/aries-bifold-core'
```
jleach, line 1: Curious if there was a component in Bifold that could just be modified to take in a different graphic?

app/src/screens/PersonCredentialLoading.tsx
```diff
@@ -0,0 +1,258 @@
...
+      try {
+        const remoteAgentDetails = await connectToIASAgent(agent, store, t)
+        setRemoteAgentDetails(remoteAgentDetails)
+        logger.error(`Connected to IAS agent, connectionId: ${remoteAgentDetails.connectionId}`)
+      } catch (err) {
+        logger.error(`Failed to connect to IAS agent, error: ${(err as BifoldError).message}`)
+      }
+    }
+    connect()
+  }, [])
```
jleach, line 95: I think the rule of thumb, any variables used inside the `useEffect` hook should be included in the dependencies array ([`agent`, `store` and `t`]), even if you don't care if they change. This ensures that the effect is aware of all the variables it relies on and can react appropriately when they change.

From the docs: "One important principle in React is local reasoning: the ability to understand what a component or hook does by looking at its code in isolation. Hooks should be treated like “black boxes” when they are called. For example, a custom hook might have used its arguments as dependencies to memoize values inside it:"

https://react.dev/reference/rules/components-and-hooks-must-be-pure

Maybe this would work around the issue?

```
  const connect = async () => {
    try {
      const remoteAgentDetails = await connectToIASAgent(agent, store, t);
      setRemoteAgentDetails(remoteAgentDetails);
      logger.error(`Connected to IAS agent, connectionId: ${remoteAgentDetails.connectionId}`);
    } catch (err) {
      logger.error(`Failed to connect to IAS agent, error: ${(err as BifoldError).message}`);
    }
  };

  useEffect(() => {
    connect();
  }, []); // Empty dependencies array since connect function is independent of props or state
```

app/src/screens/PersonCredentialLoading.tsx
```diff
@@ -0,0 +1,258 @@
...
+      const proofRequest = receivedProofRequests.find((proof) => proof.connectionId === remoteAgentDetails.connectionId)
+
+      if (!proofRequest) {
+        // No proof from our IAS Agent to respond to, do nothing.
+        logger.info(`Waited ${attestationProofRequestWaitTimeout / 1000}sec on attestation proof request, continuing`)
+
+        setDidCompleteAttestationProofRequest(true)
+      }
+    }, attestationProofRequestWaitTimeout)
+  }, [remoteAgentDetails])
```
jleach:
```suggestion
  }, [remoteAgentDetails, receivedProofRequests, attestationProofRequestWaitTimeout, logger]);
```

app/src/screens/PersonCredentialLoading.tsx
```diff
@@ -0,0 +1,258 @@
...
+    subscriptions.push(DeviceEventEmitter.addListener(AttestationEventTypes.FailedHandleProof, handleFailedAttestation))
+    subscriptions.push(DeviceEventEmitter.addListener(AttestationEventTypes.FailedHandleOffer, handleFailedAttestation))
+    subscriptions.push(
+      DeviceEventEmitter.addListener(AttestationEventTypes.FailedRequestCredential, handleFailedAttestation)
+    )
+
+    return () => {
+      subscriptions.forEach((subscription) => subscription.remove())
+    }
+  }, [])
```
jleach, line 156: I think in this one you may need:

- The **navigation** object is included in the dependencies array to ensure that the effect re-runs if navigation changes.
- **DeviceEventEmitter, BifoldEventTypes, or AttestationEventTypes** are constants and do not change, they don't need to be in the dependencies array. If they can change, they should be added to the array as well.

app/src/screens/PersonCredentialLoading.tsx
```diff
@@ -0,0 +1,258 @@
...
+  useEffect(() => {
+    for (const credential of receivedCredentialOffers) {
+      if (
+        credential.state == CredentialState.OfferReceived &&
+        credential.connectionId === remoteAgentDetails?.connectionId
+      ) {
+        goToCredentialOffer(credential.id)
+      }
+    }
+  }, [receivedCredentialOffers, remoteAgentDetails])
```
jleach, line 226: `store`?

jleach (commented):
Some thoughts on dependencies in the `useEffect`. LMK if I'm misunderstanding the docs I ref in the first comment.

app/src/screens/PersonCredentialLoading.tsx
wadeking98: attestationProofRequestWaitTimeout and logger should be constant for the duration of the screen no? I don't think its possible for either of them to change without reloading the app

app/src/screens/PersonCredentialLoading.tsx
wadeking98: Not quite sure what you mean with adding the navigation object to the dependancy array. The navigation object should only change when we trigger a navigation event like navigating to a new screen, in which case we wouldn't want to re-run the useEffect because we'd be leaving this screen right?

app/src/screens/PersonCredentialLoading.tsx
wadeking98: maybe store.developer.environment.iasPortalUrl so we're not reacting to every change that happens on the store object? IMHO iasPortalUrl shouldn't be changing while we're on this screen so should we even just treat it as a constant?

jleach (approved):

PR #2098, opened 2024-07-31
feat: attestation everywhere

jleach:
Bring in the Bifold changes that allow attestation to work form anywhere in the application.


bryce-mcmath:
@jleach some sonarcloud issues

app/ios/Podfile.lock
bryce-mcmath, line 1: Does this work on iOS? It looks like you've removed the attestation pod and haven't replaced it with anything?

bryce-mcmath (changes requested):
Question about Podfile

app/ios/Podfile.lock
jleach: I must be missing something in the attestation controller that is causing it to not be picked up. Looking.

jleach: Found it. Missing `react-native-attestation.podspec`.

bryce-mcmath (approved):

PR #2100, opened 2024-07-31
fix: negated restriction check

jleach:
Fix issue with restriction check on proof request.


jleach:
Pushed to the wrong org. Fixing.

bryce-mcmath (approved):

PR #2110, opened 2024-08-05
feat: refactored notifications

wadeking98:
Refactored the way we handle notifications in bc wallet, moved to tsyringe.
This change also fixes a couple bugs on the person credential screen, mainly the navigation failing to go back if the user closes the in app browser


bryce-mcmath (changes requested):
Gotta update the podfile.lock with `yarn ios:setup` from `/app` otherwise looks good to go

bryce-mcmath (approved):

PR #2130, opened 2024-08-16
feat!: added greater support for dependancy injection through tsyringe

wadeking98:
replaced our old useConfiguration hook for the updated tsyringe container


app/container-imp.ts
```diff
@@ -107,6 +151,92 @@ export class AppContainer implements Container {
...
+              accessibilityLabel: this.t('Settings.PlayWithBCWallet'),
+              testID: testIdWithKey('PlayWithBCWallet'),
+              onPress: () => Linking.openURL('https://digital.gov.bc.ca/digital-trust/showcase/'),
+            },
+          ],
+        },
+      ],
+      enableTours: false,
+      supportedLanguages: ['en'],
+      showPreface: false,
```
bryce-mcmath: Shouldn't `showPreface` be true?

app/container-imp.ts
```diff
@@ -107,6 +151,92 @@ export class AppContainer implements Container {
...
+            {
+              title: this.t('Settings.PlayWithBCWallet'),
+              accessibilityLabel: this.t('Settings.PlayWithBCWallet'),
+              testID: testIdWithKey('PlayWithBCWallet'),
+              onPress: () => Linking.openURL('https://digital.gov.bc.ca/digital-trust/showcase/'),
+            },
+          ],
+        },
+      ],
+      enableTours: false,
```
bryce-mcmath: Shouldn't `enableTours` be true?

app/container-imp.ts
```diff
@@ -107,6 +151,92 @@ export class AppContainer implements Container {
...
+      supportedLanguages: ['en'],
+      showPreface: false,
+      disableOnboardingSkip: false,
```
bryce-mcmath: Shouldn't `disableOnboardingSkip` be true?

app/App.tsx
```diff
@@ -1,9 +1,5 @@
 import {
-  Stacks,
-  Screens,
   AgentProvider,
-  AnimatedComponentsProvider,
-  animatedComponents,
```
bryce-mcmath, line 6: Why were these two removed? I still see `useAnimatedComponents` throughout Bifold

bryce-mcmath (changes requested):
Some small things and a question

app/App.tsx
wadeking98: not sure, must have gotten deleted when I was moving the settings to container-imp

bryce-mcmath (approved):

PR #2145, opened 2024-08-27
fix: fixed notifications on home screen when returning from mob verifier

wadeking98:
Previously when exiting the mobile verifier page and returning to the home screen the notification state wouldn't update, This would mean that there were two notifications on the home page, one from the person credential, and one from the proof request that had already been viewed. Once you navigate away and then back to the home screen the erroneous notification would be removed, but the notification bubble on the tabstack would still say there's 2 notifications.

This change fixes the useNotifications hook so that the notifications hook is updated whenever the proof request is marked as seen. This ensures we don't get an erroneous proof notification on the home screen

### DIFF
--- a/app/src/hooks/notifications.ts
+++ b/app/src/hooks/notifications.ts
@@ -67,12 +67,12 @@ export const useNotifications = (): Array<BasicMessageRecord | CredentialRecord 
     const invitationDate = new Date()
     const custom =
       showPersonCredentialSelector(credentialDefinitionIDs) &&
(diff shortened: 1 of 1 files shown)


bryce-mcmath (dismissed):

bryce-mcmath:
I'm assuming this won't hide proof requests that have been seen but not responded to. But I'm not sure how...

wadeking98:
@bryce-mcmath made a couple fixes and I'll need your review again

bryce-mcmath (approved):

PR #2167, opened 2024-09-16
feat: added ability to disable data retention

wadeking98:
added option to configuration to disable the proof verifier data retention. This prevents the verifier from saving information about people who have presented proofs to them


bryce-mcmath (changes requested):
Podfile should probably be updated as well

bryce-mcmath (approved):

PR #2175, opened 2024-09-18
chore: bump version number to 1.0.20

cvarjao:


jleach (dismissed):

jleach:
@cvarjao Looks like we need to bump our min target version. Able to update to 12 as part of this PR? Although I'm not sure what changed for this to happen. I don't think we updated any Pods recently. Or is it the pod that needs to be updated to support 12?

```
/Users/runner/work/bc-wallet-mobile/bc-wallet-mobile/app/ios/Pods/Pods.xcodeproj: warning: The iOS deployment target 'IPHONEOS_DEPLOYMENT_TARGET' is set to 11.0, but the range of supported deployment target versions is 12.0 to 17.2.99. (in target 'react-native-orientation-locker' from project 'Pods')
```

jleach (approved):

PR #2189, opened 2024-09-27
fix: lint issues

bryce-mcmath:
This PR applies React linting rules and fixes the existing lint errors in place. There is some room for refactoring, particularly in the Splash screen (as it was in bifold) where there are too many dependencies per useEffect, but that can be in a future PR. I didn't want to change any functionality at this point.


jleach (approved):
LGTM. Checked out and ran the branch with a fresh install, got and used student credential. All ok.

PR #2192, opened 2024-10-01
fix: update bifold

jleach:
Update Bifold packages to account for latest updates.


app/ios/Podfile.lock
```diff
@@ -445,7 +445,7 @@ PODS:
   - React-jsinspector (0.72.5)
   - React-logger (0.72.5):
     - glog
-  - "react-native-attestation (1.0.0-alpha.330+a657da40)":
+  - react-native-attestation (1.0.0):
```
bryce-mcmath: I think this podfile needs to be updated again with the actual package rather than the linked version

bryce-mcmath (changes requested):

app/ios/Podfile.lock
jleach: Fixed.

bryce-mcmath (approved):

PR #2199, opened 2024-10-07
feat: no backup afj folder

jleach:
Signal to iOS that it should not backup the `.afj` folder in the applications Documents so that it cannot be restored to alternative devices which can cause errors. Also, this is not meant to be backed up.

### DIFF
--- a/app/ios/AriesBifold/AppDelegate.m
+++ b/app/ios/AriesBifold/AppDelegate.m
@@ -48,19 +48,28 @@
 
 // The .afj folder from Credo cannot be restored.
 - (void)excludeDotAFJFolderFromBackup {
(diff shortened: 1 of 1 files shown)


bryce-mcmath (commented):
Will this throw an error if they attempt to do a device transfer before they have created the .afj folder? ie they only got to onboarding and didn't finish?

jleach:
> Will this throw an error if they attempt to do a device transfer before they have created the .afj folder? ie they only got to onboarding and didn't finish?

🤔  Maybe. I'll add a check to see if the folder exists, if not do nothing.

bryce-mcmath (approved):

PR #2228, opened 2024-10-21
feat: alt person flow framing skeleton

bryce-mcmath:
This PR adds in the first steps towards building out the verified in person workflow as a developer option. Some of the content is just placeholders and should be replaced with dynamic content when we get to that stage for each step. But the placeholder's for the step screen should make it pretty easy to update as most of the styling for that screen is already done. I did have to use a colour hex that isn't in our theme to match the wireframes.

Here's what it looks like:
alt_person_flow


app/__tests__/screens/__snapshots__/Developer.test.tsx.snap
bryce-mcmath, line 1: Snapshot just updated to include new option

app/src/navigators/VerifiedPersonStack.tsx
```diff
@@ -0,0 +1,59 @@
...
+        component={VerificationSteps}
+        options={{
+          title: t('Screens.VerificationSteps'),
+          headerRight: () => (
+            <IconButton
+              buttonLocation={ButtonLocation.Right}
+              icon={'help-circle'}
+              accessibilityLabel={t('PersonCredential.HelpLink')}
+              testID={testIdWithKey('Help')}
+              onPress={() => null}
```
bryce-mcmath, line 34: Once we have an actual external URL to navigate to for help content, we can add that here. For now the button does nothing.

app/src/navigators/VerifiedPersonStack.tsx
```diff
@@ -0,0 +1,59 @@
...
+        component={ChooseID}
+        options={{
+          title: t('Screens.ChooseID'),
+          headerRight: () => (
+            <IconButton
+              buttonLocation={ButtonLocation.Right}
+              icon={'help-circle'}
+              accessibilityLabel={t('PersonCredential.HelpLink')}
+              testID={testIdWithKey('Help')}
+              onPress={() => null}
```
bryce-mcmath, line 50: Same as above.

app/src/screens/ChooseID.tsx
bryce-mcmath, line 1: Just added one placeholder screen to show the process for adding new screens and steps.

app/src/screens/PersonCredential.tsx
bryce-mcmath, line 1: If the alt person flow is enabled, the start process button navigates to the new stack. Otherwise, everything is exactly the same.

app/src/screens/VerificationSteps.tsx
```diff
@@ -0,0 +1,134 @@
...
+      flexWrap: 'wrap',
+      fontSize: 16,
+    },
+    contentEmailButton: {
+      alignSelf: 'flex-end',
+    },
+  })
+
+  const steps: SetupStep[] = useMemo(() => {
+    return [
```
jleach, line 66: Just noting these might / will probably come from the server.

jleach (approved):

app/src/screens/VerificationSteps.tsx
bryce-mcmath: True

PR #2229, opened 2024-10-22
feat: added inactivity wrapper

al-rosenthal:
## What's Changed
- added `InactivityWrapper` to `App.tsx`
- fixed typos in localization
- updated hyperledger packages to get latest changes


bryce-mcmath (changes requested):
Need to run yarn ios:setup from the app folder to update the podfile.lock, otherwise looks good

app/ios/AriesBifold.xcodeproj/project.pbxproj
bryce-mcmath, line 1: This file shouldn't be committed, it has a lot of local config that can change fairly often

bryce-mcmath (approved):

PR #2231, opened 2024-10-23
docs: developer doc update

al-rosenthal:
## Whats changed:
- updated `example.env`
- updated developer docs to better reflect environment variables
- added new section for adding QR codes to android simulator


app/.env.sample
```diff
@@ -1,6 +1,7 @@
 OCA_URL=https://raw.githubusercontent.com/bcgov/aries-oca-bundles/main/
 MEDIATOR_URL=https://f326-207-194-65-204.ngrok.io?c_i=eyJAdHlwZSI6ICJodHRwczovL2RpZGNvbW0ub3JnL2Nvbm5lY3Rpb25zLzEuMC9pbnZpdGF0aW9uIiwgIkBpZCI6ICI2MjQ0ZThiNS0wNWYzLTRhYWItYjM1Yy1lYWVlMWNmZTAyM2MiLCAicmVjaXBpZW50S2V5cyI6IFsiQ3lqM1BHRUJzQ3RyUGFtTTQyRngza3BlYmR2QWdNd1lGejlFS3RmNnlUN3giXSwgImxhYmVsIjogIk1lZGlhdG9yIiwgInNlcnZpY2VFbmRwb2ludCI6ICJodHRwczovL2YzMjYtMjA3LTE5NC02NS0yMDQubmdyb2suaW8ifQ==
```
al-rosenthal, line 2: I know we spoke about this in a meeting, but is there a docs page we can store working copies of the `.env`?

If there is I can add the link as part of the PR

jleach: I think we agreed this will just be something you ask for in RC and one of us will send it - no?

jleach (approved):

PR #2236, opened 2024-10-24
fix: fix issue when action run on main

jleach:
When run on main the diff can not be agains the same ref.

### DIFF
--- a/.github/workflows/actions/early-exit-check/action.yml
+++ b/.github/workflows/actions/early-exit-check/action.yml
@@ -9,23 +9,16 @@ runs:
     - name: Check location of changed files
       shell: bash
       run: |
-        git branch
+        # On main branch, compare with the previous
+        # commit otherwise (for PRs) compare with the
+        # current commit.
+        if [ "${GITHUB_REF_NAME}" = "main" ]; then
+          end_ref="HEAD^"
+        else
+          end_ref="HEAD"
+        fi
 
-        echo "a"
-
-        pwd
-        echo "b"
-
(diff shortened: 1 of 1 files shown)


.github/workflows/actions/early-exit-check/action.yml
```diff
@@ -9,7 +9,16 @@ runs:
...
+        # On main branch, compare with the previous
+        # commit otherwise (for PRs) compare with the
+        # current commit.
+        if [ "${GITHUB_REF_NAME}" = "main" ]; then
+          end_ref="HEAD^"
+        else
+          end_ref="HEAD"
+        fi
+
+        change_count=$(git diff --name-only origin/main..${end_ref} | grep -E '^(app/.*)|(.yarn/.*)|(.github/workflows/.*)' | wc -l)
```
cvarjao: can you make `origin/main` dynamic? We can use `${GITHUB_BASE_REF:-HEAD}` since `GITHUB_BASE_REF` is only provided for PRs, the `:-HEAD` will head the default to `HEAD` when not from a PR.

jleach: Fixed, I had to include `origin` otherwise git complains if it being ambitious:

```console
Run # On main branch, compare with the previous
Comparing main with HEAD
fatal: ambiguous argument 'main..HEAD': unknown revision or path not in the working tree.
```

I also tested `origin/HEAD..HEAD^` works.

cvarjao (approved):

PR #2259, opened 2024-11-07
feat: updated bifold packages and config settings

al-rosenthal:
### What's Changed:
- Bifold packages updated to get latest changes
- Warning of "non-serializable" navigation fixed when navigating to credential details
- Config registry updated with new settings
- `OpenIDCredentialRecordProvider` added to `App.tsx`


app/container-imp.ts
```diff
@@ -221,6 +221,11 @@ export class AppContainer implements Container {
       showScanHelp: true,
       showScanButton: true,
       showDetailsInfo: true,
+      contactDetailsOptions: {
+        showConnectedTime: false,
+        enableEditContactName: false,
```
bryce-mcmath: I think these two (`showConnectedTime` and `enableEditContactName`) should be set to `true` actually. They weren't configurable before but we did have them. My bad for not mentioning these ones

bryce-mcmath (changes requested):
Small change, and also missing `Podfile.lock` update. Need to run `yarn ios:setup` from the `/app` dir to update that file. The Bifold packages include an native module package (`@hyperledger/aries-react-native-attestation`) that includes a pod, so when you update the package version, the `Podfile.lock` has to be updated too.

app/container-imp.ts
al-rosenthal: removed that object and spread the default config

app/ios/Podfile.lock
```diff
@@ -445,7 +445,7 @@ PODS:
   - React-jsinspector (0.72.5)
   - React-logger (0.72.5):
     - glog
-  - "react-native-attestation (1.0.0-alpha.353+7257e199)":
+  - react-native-attestation (1.0.0):
```
bryce-mcmath: You might be linked still. It should look like:
```suggestion
  - "react-native-attestation (1.0.0-alpha.359+<some hash>)":
```

bryce-mcmath (changes requested):
I think you may have been linked when you ran the `yarn ios:setup` command

bryce-mcmath (approved):

PR #2260, opened 2024-11-07
refactor: add initialize agent hook

bryce-mcmath:


app/src/hooks/initialize-agent.ts
```diff
@@ -0,0 +1,262 @@
...
+          autoUpdateStorageOnStartup: true,
+          autoAcceptConnections: true,
+        },
+        dependencies: agentDependencies,
+        modules: getBCAgentModules({
+          indyNetworks: ledgers,
+          mediatorInvitationUrl: Config.MEDIATOR_URL,
+          txnCache: {
+            capacity: 1000,
+            expiryOffsetMs: 1000 * 60 * 60 * 24 * 7,
```
al-rosenthal, line 106: would it make sense to keep this math in a const file?
nit: might also be nice to have a comment like: // 7 days in milliseconds

jleach (changes requested):
Testing.

al-rosenthal (commented):
small nit pick but otherwise looks and runs well.

app/src/hooks/initialize-agent.ts
```diff
@@ -0,0 +1,262 @@
...
+import { activate } from '../helpers/PushNotificationsHelper'
+import { getBCAgentModules } from '../helpers/bc-agent-modules'
+import { BCState, BCLocalStorageKeys } from '../store'
+
+const loadCachedLedgers = async (): Promise<IndyVdrPoolConfig[] | undefined> => {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const cachedTransactions = await PersistentStorage.fetchValueForKey<any>(BCLocalStorageKeys.GenesisTransactions)
+  if (cachedTransactions) {
+    const { timestamp, transactions } = cachedTransactions
+    return moment().diff(moment(timestamp), 'days') >= 1 ? undefined : transactions
```
jleach, line 30: 1 might be a magic number here. maxAgeInDays or cacheAgeInDays?

app/src/hooks/initialize-agent.ts
```diff
@@ -0,0 +1,262 @@
...
+        if (transactions) {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          await PersistentStorage.storeValueForKey<any>(BCLocalStorageKeys.GenesisTransactions, {
+            timestamp: moment().toISOString(),
+            transactions,
+          })
+        }
+      }
+
+      credDefs.forEach(async ({ did, id }) => {
```
jleach, line 186: Is forEach await safe? I though that using await inside a forEach() loop doesn't work as expected because forEach doesn't handle asynchronous functions properly—it doesn't wait for promises to resolve. Instead, you should use a `for...of` loop or `Promise.all()` with `.map()`.

app/src/hooks/initialize-agent.ts
```diff
@@ -0,0 +1,262 @@
...
+        const pool = await poolService.getPoolForDid(newAgent.context, did)
+        const credDefRequest = new GetCredentialDefinitionRequest({ credentialDefinitionId: id })
+        await pool.pool.submitRequest(credDefRequest)
+      })
+
+      schemas.forEach(async ({ did, id }) => {
```
jleach, line 192: Same.

jleach (changes requested):

app/src/hooks/initialize-agent.ts
jleach: I'd probably put it in a const file if it were used in more than one file. If it's just local to this maybe define it at the top because it's used in two different places in the same file.

bryce-mcmath: This was copied exactly as is from the splash screen in the spirit of keeping all the logic the same but I can make this change anyway

app/src/hooks/initialize-agent.ts
bryce-mcmath: This was copied exactly as is from the splash screen in the spirit of keeping all the logic the same but I can make this change anyway

app/src/hooks/initialize-agent.ts
bryce-mcmath: This was copied exactly as is from the splash screen in the spirit of keeping all the logic the same. You are right about `await` in `forEach` not behaving the way the original author of this code may have intended but I think if we're going to change it we should do so as part of a separate PR and keep the logic the same for now since it seems to be working ok.

app/src/hooks/initialize-agent.ts
bryce-mcmath: Same response as above

jleach (approved):

PR #2287, opened 2024-11-22
fix: deep links

bryce-mcmath:
This is the BC Wallet work necessary to fix deeplink issues related to this PR in Bifold: https://github.com/openwallet-foundation/bifold-wallet/pull/1327

Will need to have packages updated


al-rosenthal (approved):
good work! 🍎

PR #2320, opened 2024-12-12
fix: loki dev charts

jleach:
Update the deployment for Loki so that it deploys correctly to our dev environment.


bryce-mcmath (approved):
Nice! Is the two pod fix included in this?

PR #2322, opened 2024-12-12
feat: implement report problem feature, revert pn change

bryce-mcmath:
Leaving in draft until Bifold package deps are ready and updated but it's here if you want to see how the Bifold changes are used. Depends on:
- https://github.com/openwallet-foundation/bifold-wallet/pull/1359


app/container-imp.ts
```diff
@@ -113,7 +114,6 @@ export class AppContainer implements Container {
       lokiUrl: Config.REMOTE_LOGGING_URL,
       lokiLabels: {
         application: getApplicationName().toLowerCase(),
-        job: 'react-native-logs',
```
bryce-mcmath, line 116: `job` is now specified by the remote logger, if we also specify it here, every job will be the same whether they are incident reports or remote logging sessions. Having separate jobs should allow us to have different lifespans for each job type

app/src/hooks/initialize-agent.ts
```diff
@@ -190,7 +190,7 @@ const useInitializeBCAgent = () => {
     
     if (store.preferences.usePushNotifications) {
       logger.info('Activating push notifications...')
-      await activate(newAgent)
+      activate(newAgent)
```
bryce-mcmath, line 193: I'm hoping this reverted line will fix the issue Shel mentioned that happens during SauceLabs app initialization

bryce-mcmath (commented):
Pre-review to clarify a couple things

al-rosenthal (approved):
🌮

jleach (approved):

PR #2355, opened 2025-01-23
fix: credo ws patch and agent restart

bryce-mcmath:
This PR adds the same patch that will be added to Bifold for the `@credo-ts/core` websocket issue, which was addressed by the following PR in Credo:
- https://github.com/openwallet-foundation/credo-ts/pull/2152

Additionally, it adds a patch to `@credo-ts/indy-vdr` to allow agent reinitialization as per this PR:
- https://github.com/openwallet-foundation/credo-ts/pull/2137

It also adds back agent restart if shutdown was successful to cutdown on start up time after being locked out etc.

Depends on:
- https://github.com/openwallet-foundation/bifold-wallet/pull/1413


app/src/hooks/initialize-agent.ts
```diff
@@ -42,6 +42,35 @@ const useInitializeBCAgent = () => {
...
+    try {
+      await agent.wallet.open({
+        id: walletSecret.id,
+        key: walletSecret.key,
+      })
+      await agent.initialize()
+    } catch {
+      // if the existing agents wallet cannot be opened or initialize() fails it was
+      // again not a clean shutdown and the agent should be replaced, not restarted
+      logger.warn('Failed to restart existing agent, skipping agent restart')
```
jleach: Warn or Error?

app/src/hooks/initialize-agent.ts
```diff
@@ -168,11 +197,19 @@ const useInitializeBCAgent = () => {
...
+      setAgent(existingAgent)
+      return existingAgent
+    }
+
     const cachedLedgers = await loadCachedLedgers()
     const ledgers = cachedLedgers ?? indyLedgers
 
     const newAgent = await createNewAgent(ledgers)
     if (!newAgent) {
+      logger.error('Failed to create a new agent')
```
jleach, line 212: Should this throw. Wondering if the app can even continue if we get here?

app/src/hooks/initialize-agent.ts
bryce-mcmath: Error :)

app/src/hooks/initialize-agent.ts
bryce-mcmath: Yes, good point. It's kind of impossible to get a falsey value for `newAgent` without createNewAgent throwing itself - the same early return check of `walletSecret` is in both functions - but just incase, I added a throw.

I added a note to refactor that early return check in our tech debt discussion, but I'll keep this PR as small as possible and make that change in a later one.

jleach (approved):
Tested locally. LGTM.

Issue #2243, opened 2024-10-29
Fix Credo agent shutdown / restart on mobile


cvarjao:
Currently in Credo there is a bug that occurs only in mobile where `agent.shutdown()` fails before `isInitialized` is set to false, which prevents us from directly using this Credo API for clean agent shutdown / restart in Bifold. Fixing this in Credo (and making a patch for Bifold / BC Wallet) may improve our websocket disconnect / reconnect process and will simplify the current shutdown / restart code in Bifold / BC Wallet some too.

Relevant issue: https://github.com/openwallet-foundation/credo-ts/issues/2103

AC: nothing got worse and maybe websocket handling got better

bryce-mcmath:
https://github.com/openwallet-foundation/credo-ts/issues/2103

jeznorth:
@bryce-mcmath please updated with AC as discussed.

Issue #2226, opened 2024-10-18
Unable to get person credential


jleach:
User reported unable to get person credential. Issue seems to be consistent for him but intermittent for others. On backgrounding / foregrounding the app message trickle in one at a time.

There may be two different issues:

1. Some times the wallet hangs and no messages (basic or offer) are received. Restarting the wallet fixes this and both messages arrive.
2. Some times the offer is received but the wallet does not automatically accept the offer. 

### 1. No Messages

- Was able to reproduce on the Indecio public mediator and BC mediator.
- Both messages (basic message and offer) were in notifications when I closed the "please wait..." screen.
- Connection issues to IAS. The wallet hangs at "Connected to IAS agent, connectionId: ....". Could be a mediator issue as I've never seen this when testing with ACA-py before.
- Tried `mediator.dev.animo.id` and that worked no better.
- The BC mediator seems the most stable and works the most reliably out of the other two tested.

### 2. Hang

1. This issue can be reproduced by restarting the app and get a person credential.
2. Then I delete it, and repeat the process. This time I pause on a BCSC screen until you see the following three lines appear in the mediator logs (1.0.20 may only print one line so repetition may be irrelevant):

```console
2024-10-18 21:10:18,830 aries_cloudagent.transport.inbound.ws ERROR Unexpected Websocket message type received: WSMsgType.CLOSED: None, None
2024-10-18 21:10:18,831 aries_cloudagent.transport.inbound.ws ERROR Unexpected Websocket message type received: WSMsgType.CLOSED: None, None
2024-10-18 21:10:18,831 aries_cloudagent.transport.inbound.ws ERROR Unexpected Websocket message type received: WSMsgType.CLOSED: None, None
```

3. Complete the process.

You will see this in the mediator logs, at which point the BC Wallet app will be stalled at "Just a moment....". For both 1.0.20 and 1.0.21 it repeats 3x.

```console
2024-10-18 21:10:25,945 aries_cloudagent.core.conductor WARNING Cannot queue message for delivery, no supported transport
2024-10-18 21:10:26,166 aries_cloudagent.core.conductor WARNING Cannot queue message for delivery, no supported transport
2024-10-18 21:10:26,529 aries_cloudagent.core.conductor WARNING Cannot queue message for delivery, no supported transport
```

* This happens with or without pausing mediator message pickup;
* Android 14
* 1.0.21 latest
[...]

PR #2370, opened 2025-02-05
fix(deps): patch for credo core and credo hooks

bryce-mcmath:
These patches are required for proper behaviour after `agent.shutdown()` and consecutive `agent.initialize()`. Without these patch updates it is currently broken in this situation (after lockout for example)

Note: I removed the `.js.map` parts of the existing patches, as they already weren't working and are only used for source maps and make the patches harder to read - all source maps are used for is making code easier to read after it's been minified, which is more relevant for web development rather than mobile.  

Relevant issues: #2226 and #2243


al-rosenthal (approved):
ran this locally and was able to get my person credential after a lockout, well done!

PR #2402, opened 2025-03-12
chore(deps): upgrade credo, bifold pkgs

bryce-mcmath:
This follows Mostafa's PR in Bifold and will need to be merged before BC + Bifold can work properly together again


al-rosenthal:
I got this running locally on my android device and cannot get my person credential, I get this error:

Screenshot 2025-03-12 at 12 51 08 PM

bryce-mcmath:
@al-rosenthal so running locally, prod Android attestation won't work because it's not signed by the play store, which is expected. We can hop in the bridge and I'll show you how to test Android attestation locally

jleach (approved):

jleach:
> @al-rosenthal so running locally, prod Android attestation won't work because it's not signed by the play store, which is expected. We can hop in the bridge and I'll show you how to test Android attestation locally

What Bryce said + switch to Test env in Developer settings.

PR #2406, opened 2025-03-13
refactor: bring in new onboarding mechanics

jleach:
This PR brings in the changes from https://github.com/openwallet-foundation/bifold-wallet/pull/1464 as well as:

- Renames "AppState" in App.tsx to be more descriptive to what it actually does;
- Integrates the new on boarding mechanics from [PR 1464](https://github.com/openwallet-foundation/bifold-wallet/pull/1464)
- Integrates the new update check mechanics from [PR 1466](https://github.com/openwallet-foundation/bifold-wallet/pull/1466)
- Excludes a working directory from git.
- Moves the app help URL to constants with some other URLs.


app/src/onboarding.ts
```diff
@@ -0,0 +1,84 @@
...
+    didConsiderPushNotifications,
+    didNameWallet,
+  } = state.onboarding
+  const { didAuthenticate } = state.authentication
+  const { servedPenalty } = state.loginAttempt
+  const { enableWalletNaming } = state.preferences
+  const { showPreface, enablePushNotifications } = config
+  const { needsUpdate, dismissed = false } = state.versionInfo
+
+  console.log('***** state.versionInfo', needsUpdate)
```
bryce-mcmath: safely remove?

app/src/screens/Splash.tsx
```diff
@@ -185,179 +109,43 @@ const Splash = () => {
...
-  useEffect(() => {
+    setStep(1)
+    setStep(2)
+
     const initAgentAsyncEffect = async (): Promise<void> => {
       try {
-        if (
-          !mounted ||
-          initializing.current ||
-          !store.authentication.didAuthenticate ||
```
bryce-mcmath, line 315: I think we still need just the `!store.authentication.didAuthenticate` check here as per this PR:
https://github.com/openwallet-foundation/bifold-wallet/pull/1477

app/src/screens/Splash.tsx
```diff
@@ -185,179 +109,43 @@ const Splash = () => {
...
-
         setStep(3)
         await (ocaBundleResolver as RemoteOCABundleResolver).checkForUpdates?.()
 
         setStep(4)
         const agent = await initializeAgent()
+        initializing.current = true
+
         if (!agent) {
-          initializing.current = false
```
bryce-mcmath, line 330: I think we still need these two, no? Or else if it fails initialization the first, the retry button won't work

app/src/screens/Splash.tsx
```diff
@@ -185,179 +109,43 @@ const Splash = () => {
...
         setStep(5)
+
         navigation.dispatch(
           CommonActions.reset({
             index: 0,
             routes: [{ name: Stacks.TabStack }],
           })
         )
       } catch (e: unknown) {
-        initializing.current = false
```
bryce-mcmath, line 342: "

bryce-mcmath (changes requested):

app/src/onboarding.ts
jleach: Ya, weird our robots didn't catch this.

jleach: Fixed.

app/src/screens/Splash.tsx
jleach: Fixed

app/src/screens/Splash.tsx
jleach: Fixed

app/src/screens/Splash.tsx
jleach: Fixed

app/src/screens/Splash.tsx
```diff
@@ -24,77 +22,11 @@ import TipCarousel from '../components/TipCarousel'
 import useInitializeBCAgent from '../hooks/initialize-agent'
 import { BCState } from '../store'
 
-import { TermsVersion } from './Terms'
-
-const OnboardingVersion = 2
```
bryce-mcmath, line 29: So `OnboardingVersion` just isn't needed anymore? That's awesome

app/src/screens/Splash.tsx
bryce-mcmath: Just curious, is there a reason you moved the early return after the `checkForUpdates()` call? Might cause it to be called many times

bryce-mcmath (changes requested):
One small nit about placement of `checkForUpdates()` call

app/src/screens/Splash.tsx
jleach: No, I can't think of any reason to keep it. We do use TermsVersion.

app/src/screens/Splash.tsx
jleach: I'm on the fence with this one. You can't get to the Splash screen unless you complete either the PIN Create or PIN Enter which both complete authentication. So if you're seeing Splash you must have completed authentication. So, on 2nd though - this change may not have been needed. 

EDIT: It was needed in the previous incarnation because Splash would have routed to the PIN mechanics. This is no longer the case.

Thoughts?

I moved it after because it should be cached at this point and `didAuthenticate` is only required for the agent. I think it would be better to move the initialized check up if we don't want to repeat this fn call. Thoughts?

release-fb.xcconfig
bryce-mcmath, line 1: accidentally committed?

bryce-mcmath (approved):

PR #2413, opened 2025-03-18
feat(deps): upgrade to react native 0.73.6

bryce-mcmath:
This PR closely follows the RN PR recently merged into Bifold. Here are some notes for devs on things I ran into in order to get things up and running smoothly:

- Upgrade Android Studio to Meerkat so we're all using the same version
- Delete all `node_modules` and `.yarn/cache`
- Get Java Zulu 17 as per the React Native 73 docs, and set that in your shell profile as so:
```
export JAVA_HOME=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home
export STUDIO_JDK=/Library/Java/JavaVirtualMachines/zulu-17.jdk/Contents/Home
```
- Restart your shell
- `yarn install` from `bc-wallet-mobile/`
- In Android Studio, click `Sync project with gradle files` icon in the upper right
- Change your Android Studio Java and Gradle SDK to your Zulu 17 path (should pop up as an option if you've set it to your JAVA_HOME and STUDIO_JDK)
- Create a file in `app/ios/` called `.xcode.env.local` that just contains this one line:
`export NODE_BINARY=$(which node)`
- From `/app` run `sudo bundle install --no-deployment`
- Run `yarn ios:setup`
- Build once from both Android Studio and Xcode (should work without any issue)
- Back to developing as usual


jleach (dismissed):

al-rosenthal:
only thing I had to do differently was run `yarn install` from `bc-wallet-mobile/app/` as well. The root install didn't seem to work when I was getting things setup

app/ios/AriesBifold/AppDelegate.mm
al-rosenthal, line 1: did you try this without changing it to `.mm` @bryce-mcmath ?

al-rosenthal (dismissed):

app/ios/AriesBifold/AppDelegate.mm
bryce-mcmath: I didn't actually, just because apparently `.mm` is the recommended way: https://react-native-community.github.io/upgrade-helper/?from=0.72.17&to=0.73.6

Good to stick with Bifold anyway

bryce-mcmath:
@al-rosenthal 

> only thing I had to do differently was run `yarn install` from `bc-wallet-mobile/app/` as well. The root install didn't seem to work when I was getting things setup

That's... interesting. I haven't had to do that. Did you remove both node_modules folders (one in root and one in /app) and the .yarn/cache folder before installing?

al-rosenthal:
> @al-rosenthal
> 
> > only thing I had to do differently was run `yarn install` from `bc-wallet-mobile/app/` as well. The root install didn't seem to work when I was getting things setup
> 
> That's... interesting. I haven't had to do that. Did you remove both node_modules folders (one in root and one in /app) and the .yarn/cache folder before installing?

I did, from `/app` then `/bc-wallet-mobile`. I was a little confused at first but the install might have just skipped or something? when I checked`/app/node_modules` it was empty and running install in `/app` did the trick

jleach (approved):

PR #2420, opened 2025-03-27
fix(deps): namewallet bug, patch credo deps, update bifold pkgs

bryce-mcmath:
This it to include these changes here: 
- https://github.com/openwallet-foundation/credo-ts/pull/2237
- https://github.com/openwallet-foundation/credo-ts/pull/2240

Also fixes a bug related to onboarding where the `NameWallet` screen was showing when it shouldn't.

It also updates Bifold packages, which includes some changes like:
- reasonable limits on text scaling
- theme updates
- lockout and update screen improves, especially at larger zoom / font sizes
- a bug fix where the inactivity based lockout didn't work 

Depends on: 
- openwallet-foundation/bifold-wallet#1484


.yarn/patches/@credo-ts-anoncreds-npm-0.5.13-446ac3168e.patch
```diff
@@ -0,0 +1,49 @@
...
+index c1eeeca1c9c254f47880909391b22f66598f42fd..779b88d59841070a5e298e0544908a7521bc1cf2 100644
+--- a/build/anoncreds-rs/AnonCredsRsHolderService.js
++++ b/build/anoncreds-rs/AnonCredsRsHolderService.js
+@@ -82,7 +82,8 @@ let AnonCredsRsHolderService = class AnonCredsRsHolderService {
+                         ].join('\n'));
+                     }
+                 }
+-                const { linkSecretId, revocationRegistryId, credentialRevocationId } = (0, w3cAnonCredsUtils_1.getAnoncredsCredentialInfoFromRecord)(credentialRecord, (0, proofRequest_1.proofRequestUsesUnqualifiedIdentifiers)(proofRequest));
++                const proofUsesUnqualifiedIdentifiers = (0, proofRequest_1.proofRequestUsesUnqualifiedIdentifiers)(proofRequest);
++                const { linkSecretId, schemaId, credentialDefinitionId, revocationRegistryId, credentialRevocationId } = (0, w3cAnonCredsUtils_1.getAnoncredsCredentialInfoFromRecord)(credentialRecord, (0, proofRequest_1.proofRequestUsesUnqualifiedIdentifiers)(proofRequest));
```
bryce-mcmath: @jamshale could we not just pass in the var above like this?
```suggestion
+                const { linkSecretId, schemaId, credentialDefinitionId, revocationRegistryId, credentialRevocationId } = (0, w3cAnonCredsUtils_1.getAnoncredsCredentialInfoFromRecord)(credentialRecord, proofUsesUnqualifiedIdentifiers);
```

jamshale: I think this must be a js compiler thing? That's how it is with the ts code. I think you could change it if you wanted.

image

.yarn/patches/@credo-ts-anoncreds-npm-0.5.13-446ac3168e.patch
```diff
@@ -0,0 +1,82 @@
...
+                 const timestamp = attribute.timestamp;
+@@ -112,6 +113,13 @@ let AnonCredsRsHolderService = class AnonCredsRsHolderService {
+                             credential: credentialRecord.credential,
+                         })
+                         : credentialRecord.credential;
++
++                    if (proofUsesUnqualifiedIdentifiers) {
++                        credential.schema_id = schemaId;
++                        credential.cred_def_id = credentialDefinitionId;
++                        credential.rev_reg_id = credentialRevocationId != null ? credentialRevocationId : undefined;
```
bryce-mcmath:
```suggestion
+                        credential.rev_reg_id = revocationRegistryId != null ? revocationRegistryId : undefined;
```

loneil: Error I see in Traction is
```
marshmallow.exceptions.ValidationError: {'identifiers': {0: {'rev_reg_id': ['Value 1565 is not an indy revocation registry identifier']}}}
```
So "1565" probably looks like a credential revocation ID instead of the rev reg ID

jamshale: That was my bad. I think I just saw the RevocationId part and overlooked it was assigning the wrong thing here. I should have tested better this scenario with revocation as well.

jleach (approved):

PR #2427, opened 2025-04-03
feat(deps): prevent screenshots, update bifold pkgs

bryce-mcmath:
BC Wallet-specific part of the work from these Bifold PRs:
- openwallet-foundation/bifold-wallet#1493
- openwallet-foundation/bifold-wallet#1494
- openwallet-foundation/bifold-wallet#1497
- openwallet-foundation/bifold-wallet#1499

This includes the ledger update which gets the time to initialize back down to a reasonable amount and speeds up every interaction that involves ledger lookup. May also speed things up further for users getting and using the new lawyer credential as they are now included in our caching


al-rosenthal (commented):
I tried this out on an iOS device locally and was able to snap screenshots without issue. I was expected credential details to be blanked out from a screen shot

bryce-mcmath:
> I tried this out on an iOS device locally and was able to snap screenshots without issue. I was expected credential details to be blanked out from a screen shot

@al-rosenthal It's only applied to the `ProofDetails` screen, which is seen when you are using the mobile verifier from the verifier perspective and someone sends you a proof. But for an easier way to see it, just add `usePreventScreenCapture()` to `bc-wallet-mobile/app/src/screens/Developer.tsx` locally and then try and screenshot from the developer screen

bryce-mcmath:
That screen was the only place they wanted them to be blocked (so far)

fc-santos:
Hey guys, we had a problem after adding the prevent screenshots lib when deploying to the Google Play Store. We probably need to add a permission in the android manifest. Here is the error we got in the pipeline: https://github.com/MCN-ING/Portefeuille-mobile-qc/actions/runs/14267963097/job/39994268320. Here are the permissions the lib itself requires in the android manifest : https://github.com/wn-na/react-native-capture-protection/blob/main/android/src/main/AndroidManifest.xml

bryce-mcmath:
@fc-santos fixed with https://github.com/openwallet-foundation/bifold-wallet/pull/1499

app/container-imp.ts
```diff
@@ -267,11 +268,10 @@ export class AppContainer implements Container {
       { did: 'TeT8SJGHruVL9up3Erp4o', id: 'TeT8SJGHruVL9up3Erp4o:3:CL:400095:SpecialEventServer' },
       { did: 'Ttmj1pEotg8FbKZZD81S7i', id: 'Ttmj1pEotg8FbKZZD81S7i:3:CL:184:SpecialEventServer' },
 
-      { did: '4xE68b6S5VRFrKMMG1U95M', id: '4xE68b6S5VRFrKMMG1U95M:3:CL:59232:default' },
+      { did: 'QzLYGuAebsy3MXQ6b1sFiT', id: 'QzLYGuAebsy3MXQ6b1sFiT:3:CL:2351:lawyer' }, // prod lawyer credential
```
jleach, line 271: Why did these need to be updated? Was it to match the new credential issued a few weeks ago?

al-rosenthal (approved):

app/container-imp.ts
bryce-mcmath: Yes

PR #2443, opened 2025-04-24
feat: bcsc core package

jleach:
Introduce the BCSC core package, including essential configurations, native module implementation, and initial test setup. This package provides functionality and data access for the legacy BCSC application.


bryce-mcmath:
if it's not too difficult should we maybe change the name to something like BCSCData or BCSCDataTransfer or BCSCDataMigrate? Not the end of the world if it stays BCSCCore it just sounds like it would be the core functionality of the BCSC app

jleach:
How about `react-native-bcsc-security` or `react-native-bcsc-auth` since this primarily deals with the security components of BCSC like RSA, Tokens, JWT and JWK. @bryce-mcmath

bryce-mcmath:
> How about `react-native-bcsc-security` or `react-native-bcsc-auth` since this primarily deals with the security components of BCSC like RSA, Tokens, JWT and JWK

love it, either works for me

packages/bcsc-core/.nvmrc
```diff
@@ -0,0 +1 @@
+v20.19.0
```
bryce-mcmath: Should we match BC Wallet + Bifold node versions for now?

packages/bcsc-core/ios/String.swift
```diff
@@ -0,0 +1,217 @@
...
+    var nonEmptyUnwrapped: String? {
+        if let unwrappedSelf = self, unwrappedSelf.trim() != "" {
+            return unwrappedSelf
+        }
+        
+        return nil
+    }
+}
+
+extension String {
```
al-rosenthal, line 189: nit: String gets extended 3 times in this file, should we merge them?

packages/bcsc-core/ios/StorageService.swift
```diff
@@ -0,0 +1,154 @@
...
+            }
+        } catch {
+            print("StorageService: Error - Could not access or read account_list: \(error).")
+            return nil
+        }
+    }
+    var basePath: String {
+        return "\(currentBundleID)/data/accounts_dir/\(currentEnvName)"
+    }
+    var provider = "https://idsit.gov.bc.ca/device/"
```
al-rosenthal, line 82: this is the dev environment right? should this change for production?

packages/bcsc-core/ios/String.swift
jleach: That file is taken from BCSC. I'd prefer to not edit legacy code if possible.

packages/bcsc-core/ios/StorageService.swift
jleach: When we're more advanced we'll need to pass in the provider (env) when the environment changes in developer settings. That's another feature tracked by [2485](https://github.com/bcgov/bc-wallet-mobile/issues/2485).

packages/bcsc-core/.nvmrc
jleach: Fixed.

al-rosenthal (approved):

PR #2481, opened 2025-06-02
feat:ias core api

al-rosenthal:
## What's changed:
- added api client for ias server
- added refresh token work for ias requests
- added pairing code api
- added dataloader component for api calls
- added hooks for api usage
- new status bar component
- pairing code screen wired up to api

Here's an example of the api client, api hooks, data loader, ui all hooked up:
data_load


.github/workflows/actions/setup-node/action.yml
```diff
@@ -7,7 +7,7 @@ inputs:
   node-version:
     description: Select the nodejs version to use
     required: false
-    default: "18.20.6"
```
al-rosenthal, line 10: shouldn't this be 18.20.6?

app/package.json
```diff
@@ -79,10 +79,14 @@
...
     "@react-navigation/stack": "^6.3.21",
     "axios": "1.4.0",
+    "base-64": "^1.0.0",
     "buffer": "6.0.3",
     "credo-ts-indy-vdr-proxy-client": "0.2.3",
     "fast-text-encoding": "^1.0.6",
     "i18next": "21.10.0",
+    "jsrsasign": "^11.1.0",
+    "jsrsasign-util": "^1.0.5",
+    "jwt-decode": "^4.0.0",
```
bryce-mcmath: @al-rosenthal are these still needed?

bryce-mcmath (commented):
Just one comment. I need to undo the AriesBifold.xcodeproj changes but then it should be good to go

app/package.json
al-rosenthal: nope, the bcsc package is handling all the signing now

al-rosenthal: I think the we need the jwt-decode package but the other jsrassign packages can be removed

bryce-mcmath (dismissed):

bryce-mcmath:
Just realized we should add an `IAS_URL` to our env before merging

bryce-mcmath (approved):

PR #2499, opened 2025-06-13
feat: bcsc core account

jleach:
This PR implements the foundational BCSC (BC Services Card) core data access functionality to create and write account data.

### DIFF
--- a/app/ios/Podfile.lock
+++ b/app/ios/Podfile.lock
@@ -1031,7 +1031,7 @@ PODS:
   - React-Mapbuffer (0.73.6):
     - glog
     - React-debug
-  - react-native-attestation (2.1.7):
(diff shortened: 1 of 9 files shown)


app/ios/Podfile.lock
bryce-mcmath, line 1: idk if we need to commit this file change

bryce-mcmath (changes requested):
Just the one unneeded diff. Are we going to need to setAccount every time the security method changes? ie. biometrics -> pin only or vice versa?

app/ios/Podfile.lock
jleach: I removed it. For me it complains. I matched it to `main`.

bryce-mcmath (approved):

PR #2501, opened 2025-06-13
fix: bcsc core android mock

jleach:
Fix android mocks so the current version works as expected.


travkhl (approved):
Fixes the build issue!

PR #2520, opened 2025-06-25
feat: add did:webvh resolver

brianorwhatever:
This PR adds the did:webvh resolver from [this](https://github.com/openwallet-foundation/credo-ts/pull/2238) open PR in credo.

Anything that is not related to the resolver is to appease a very finicky build / test / commit automation process. I'm happy to remove any of that stuff if there are better workarounds or I'm missing things

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -56,6 +56,7 @@
     "@credo-ts/question-answer": "0.5.13",
     "@credo-ts/react-hooks": "0.6.0",
     "@credo-ts/react-native": "0.5.13",
+    "@credo-ts/webvh": "0.5.17-alpha-20250818210456",
     "@formatjs/intl-datetimeformat": "~4.2.6",
     "@formatjs/intl-displaynames": "~5.2.6",
     "@formatjs/intl-getcanonicallocales": "~1.7.3",
--- a/app/src/utils/bc-agent-modules.ts
+++ b/app/src/utils/bc-agent-modules.ts
@@ -8,6 +8,7 @@ import {
   V1CredentialProtocol,
   V1ProofProtocol,
 } from '@credo-ts/anoncreds'
+import type { AnonCredsRegistry } from '@credo-ts/anoncreds'
 import { AskarModule } from '@credo-ts/askar'
 import {
   Agent,
@@ -15,6 +16,7 @@ import {
   AutoAcceptProof,
   ConnectionsModule,
   CredentialsModule,
+  DidResolver,
   DidsModule,
   DifPresentationExchangeProofFormatService,
   MediationRecipientModule,
@@ -23,6 +25,7 @@ import {
   V2CredentialProtocol,
   V2ProofProtocol,
 } from '@credo-ts/core'
+import { WebvhDidResolver, WebVhAnonCredsRegistry } from '@credo-ts/webvh'
 import { DrpcModule } from '@credo-ts/drpc'
 import { IndyVdrAnonCredsRegistry, IndyVdrModule, IndyVdrPoolConfig } from '@credo-ts/indy-vdr'
 import { PushNotificationsApnsModule, PushNotificationsFcmModule } from '@credo-ts/push-notifications'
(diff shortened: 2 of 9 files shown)


download-react-native-binaries.sh
cvarjao, line 1: Doesn't this happens automatically with `yarn install`?

packages/bcsc-core/package.json
```diff
@@ -80,7 +80,7 @@
     "prettier": "^3.0.3",
     "react": "19.0.0",
     "react-native": "0.78.2",
-    "react-native-builder-bob": "^0.40.6",
```
cvarjao: is there a reason to downgrade react-native-builder-bob from 0.40.6 to ^0.39.1?

app/package.json
```diff
@@ -57,6 +57,7 @@
     "@credo-ts/question-answer": "0.5.13",
     "@credo-ts/react-hooks": "0.6.0",
     "@credo-ts/react-native": "0.5.13",
+    "@credo-ts/webvh": "file:../credo-ts/packages/webvh",
```
cvarjao: a file dependency wont work. Is that because we are waiting for a new published package from from credo?

cvarjao (commented):
This PR seems to have a lot of dependency changes that I was not expecting it.

download-react-native-binaries.sh
brianorwhatever: I thought it would too but it didn't seem to be doing it so I had to write this script. Not sure why I wasn't getting the binaries automatically but am happy to remove this script

app/package.json
brianorwhatever: yes waiting for new credo

packages/bcsc-core/package.json
brianorwhatever: I was having issues building this package with the 0.40 version.

- react-native-builder-bob version 0.40.x introduced a dependency on arktype ^2.1.15
- arktype version 2.x is an ES Module ("type": "module" in package.json)
- react-native-builder-bob was trying to import arktype using CommonJS require() statements
- This created an incompatibility between ESM and CommonJS module systems

Happy to use another solution if there is one

brianorwhatever:
> This PR seems to have a lot of dependency changes that I was not expecting it.

I was also not expecting it.. Some of these things might not be requirements anymore as I started this work before quite a few changes that have happened since then..

Once the credo PR is merged I can investigate which of these changes are actually required. Definitely possible some of them were workarounds to things I was doing wrong..

app/src/utils/bc-agent-modules.ts
```diff
@@ -26,6 +26,7 @@ import {
 import { DrpcModule } from '@credo-ts/drpc'
 import { IndyVdrAnonCredsRegistry, IndyVdrModule, IndyVdrPoolConfig } from '@credo-ts/indy-vdr'
 import { PushNotificationsApnsModule, PushNotificationsFcmModule } from '@credo-ts/push-notifications'
+import { WebvhDidResolver } from '@credo-ts/webvh'
```
jamshale: I thought we'd want to load the WebvhAnoncredsRegistry as well?

app/ios/AriesBifold.xcodeproj/project.pbxproj
```diff
@@ -739,7 +703,7 @@
 				OTHER_LDFLAGS = "$(inherited)";
 				REACT_NATIVE_PATH = "${PODS_ROOT}/../../node_modules/react-native";
 				SDKROOT = iphoneos;
-				USE_HERMES = true;
```
cvarjao: hermes is the optimized and lightweight version of node that are used for mobile, any reason to turn that off?

bryce-mcmath:
@brianorwhatever safe to close this PR?

edit: can still use the closed PR as a reference when we come back to this in the future

brianorwhatever:
will create a new PR with the new code

PR #2526, opened 2025-06-27
feat(pickup): Manually batch pickup queued messages

jamshale:
I know the real goal is to get everything migrated to the credo mediator with pickupv2 support but this is going to take some time. To do it without the user noticing you need to update all the serviceEndpoints for the didDocs for all connections through the mediator.

This just fixes the single message pickup problem by repeatedly doing the pickupv1 ten times in the background on startup or restarts.

The only difference to the user is they will load with a small delay (.05secs) between the UI elements displaying.

Tested with the dev acapy mediator.


bryce-mcmath (approved):
Thanks for the PR 🙌

Issue #2251, opened 2024-10-29
Add logging to indicate when google play integrity is not available


jleach:
If Google Play integrity is not available on device log a message indicating this fact.

```javascript
    const available = await isPlayIntegrityAvailable()
    if (!available) {
      return null
    }
```

For a stretch goal, all logging should make it easier to search through by including a key word like:

```javascript
log.error(`${componentName}: Google Play Integrity is unavailable`)
```

Prints
```console
AttestationService: Google Play integrity is unavailable
```

cvarjao:
This seems to be important enough that should surface some sort of error message as opposed to silently fails, no?

PR #2540, opened 2025-07-09
feat: Add more logging and loud failures for Google play integrity issues 

travkhl:
Added error logging and error modals to Google Play integrity errors.

showndetails

Error modal shown when isPlayIntegrityAvailable() returns false

other
error shown when googleAttestation() fails

Related to https://github.com/bcgov/bc-wallet-mobile/issues/2251

### DIFF
--- a/app/src/services/attestation.ts
+++ b/app/src/services/attestation.ts
@@ -627,7 +627,7 @@ export class AttestationMonitor implements AttestationMonitorI {
         'Google Attestation Error',
         'There was a problem with the Google Integrity API.',
         (error as Error)?.message || 'No details provided.',
-        AttestationErrorCodes.FailedToGenerateAttestation,
+        AttestationErrorCodes.IntegrityUnavailable,
       )
       throw bifoldError
     }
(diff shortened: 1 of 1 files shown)


app/src/services/attestation.ts
```diff
@@ -601,18 +602,40 @@ export class AttestationMonitor implements AttestationMonitorI {
...
-      return null
+      this.log?.error('Google Play Integrity is unavailable')
+
+      const error = new BifoldError(
+        'Google Play Integrity Unavailable',
+        'Google Play Integrity is required for device attestation but is not available on this device.',
+        'The device attestation process cannot be completed without Google Play Integrity services.',
+        AttestationErrorCodes.IntegrityUnavailable,
+      )
+      DeviceEventEmitter.emit(AttestationEventTypes.FailedHandleOffer, error)
```
bryce-mcmath: Why is this error emitted but not the one below?

travkhl: I don't think this one needs to be emitted either.

bryce-mcmath (approved):
🔔

Issue #2533, opened 2025-07-03
Create UI: Barcode Scanning


travkhl:
Implementation of the Barcode Scanning interface.

## Acceptance Criteria
- Implement interface(s) and workflow as defined in design specifications listed below.
- For Combined bcsc dl flow:
  - add screen to scan serial number from 1d code
  - integrate with existing flow

## Design Specifications
- https://github.com/bcgov/bc-wallet-mobile/issues/2519

PR #2543, opened 2025-07-14
feat: code scanning screen and components for alt person flow

travkhl:
This pr adds the screen for scanning a barcode from a combined bcsc/dl as per https://github.com/bcgov/bc-wallet-mobile/issues/2533. It also adds a CodeScanningCamera component that can be reused when we need to scan other types of codes. More work needs to be done to sanitize inputs from scanned codes. Work for this needs to be defined in https://github.com/bcgov/bc-wallet-mobile/issues/2544.

untitled480

### DIFF
--- a/app/src/bcsc-theme/features/verify/ScanSerialScreen.tsx
+++ b/app/src/bcsc-theme/features/verify/ScanSerialScreen.tsx
@@ -1,0 +1,102 @@
+import { Button, ButtonType, testIdWithKey, ThemedText, useStore, useTheme } from '@bifold/core'
+import React, { useCallback, useState } from 'react'
+import { useTranslation } from 'react-i18next'
+import { StyleSheet, View } from 'react-native'
+
+import { BCDispatchAction, BCState } from '@/store'
+import { StackNavigationProp } from '@react-navigation/stack'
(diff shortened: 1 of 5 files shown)


app/src/bcsc-theme/features/verify/ScanSerialScreen.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+    buttonContainer: {
+      width: '100%',
+    },
+  })
+
+  const validateSerial = useCallback(() => {
+    // TODO: update this validation logic
+    // once we know the serial number format
+    if (serial.length < 1) {
```
al-rosenthal, line 51: do we just need a regex for this? or do we need something more comprehensive for the different types or cards?

al-rosenthal (approved):
tested and worked on both iOS and android devices, good job!

app/src/bcsc-theme/features/verify/ScanSerialScreen.tsx
travkhl: regex for checking for alphanumeric characters of a certain length should suffice but we will need to find the relevant spec

travkhl: related to https://github.com/bcgov/bc-wallet-mobile/issues/2544

PR #2572, opened 2025-07-24
feat: android feature parity

jleach:
Implement dynamic retrieval of the app version and build number in the BcscCoreModule, replacing hardcoded values with methods that fetch the current version and build number from package info. This enhances the accuracy of the build information.

### DIFF
--- a/app/src/bcsc-theme/features/verify/send-video/PhotoReviewScreen.tsx
+++ b/app/src/bcsc-theme/features/verify/send-video/PhotoReviewScreen.tsx
@@ -82,6 +82,7 @@ const PhotoReviewScreen = ({ navigation, route }: PhotoReviewScreenProps) => {
       }
 
       dispatch({ type: BCDispatchAction.SAVE_PHOTO, payload: [{ photoPath, photoMetadata }] })
+
       navigation.dispatch(
         CommonActions.reset({
           index: 2,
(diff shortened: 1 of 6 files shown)


bryce-mcmath (dismissed):

app/src/bcsc-theme/features/verify/send-video/PhotoReviewScreen.tsx
```diff
@@ -107,7 +108,7 @@ const PhotoReviewScreen = ({ navigation, route }: PhotoReviewScreenProps) => {
   return (
     <SafeAreaView style={styles.pageContainer}>
       <View style={styles.contentContainer}>
-        <Image source={{ uri: photoPath }} style={{ height: '100%', width: 'auto', resizeMode: 'cover' }} />
+        <Image source={{ uri: `file://${photoPath}` }} style={{ height: '100%', width: 'auto', resizeMode: 'cover' }} />
```
bryce-mcmath, line 111: Adding this prefix doesn't break it on iOS?

jleach: Correct. I've tested both. With it both work. Without it only iOS works. It seems to be more forgiving.

jleach:
@bryce-mcmath I fixed that missing path check we discussed on Teams.

bryce-mcmath (approved):

Issue #2550, opened 2025-07-16
BCSC Service Notices


cvarjao:
Show notifications from BC Service in the application, leveraging the new notification component built as part of #2385.

## Acceptance Criteria
- Service notices should appear in the notifications component
- Investigate / review if notices have associated 'severity' to ensure we are using an appropriate visual cue (e.g. Info, Alert, Danger variations).

## Developer Notes
- Consult @hannahgov @jeznorth if variations occur outside of A/C

PR #2583, opened 2025-07-28
feat: BCSC service notices banners (Issue #2550)

MacDeluca:
This pull request introduces enhancements to the server status handling, error logging, and user interface for the BCSC theme. The changes improve server status checks, add a notification banner for server issues, and refine error handling in the API client and Home component.

This pull request introduces the notification banners for the BCSC app / theme. It includes a new component `NotificationBannerContainer` which renders the banners while also fetching the IAS server status (to display outages as banners).

It also includes some additional documentation and a potential slight improvement for endpoints which require no bearer auth (open to feedback)


bryce-mcmath (approved):
Approved but noting that in future we should probably memoize all these API hooks the same way we did for `useUserApi.tsx` so we don't have to break linting rules to prevent re-renders. We should probably also make use of this new skipBearerAuth as needed instead of just having one long if statement we have to keep adding to. @al-rosenthal thoughts?

al-rosenthal:
> Approved but noting that in future we should probably memoize all these API hooks the same way we did for `useUserApi.tsx` so we don't have to break linting rules to prevent re-renders. We should probably also make use of this new skipBearerAuth as needed instead of just having one long if statement we have to keep adding to. @al-rosenthal thoughts?

I think the skipAuth flag is a great addition it is a much cleaner implementation

PR #2586, opened 2025-07-30
feat: refactor existing BCWallet theme to use ThemeBuilder

MacDeluca:
### Related Issues
This PR will need to go in once Bifold is released with this PR.
[Bifold ThemeBuilder fix](https://github.com/openwallet-foundation/bifold-wallet/pull/1602)
Bump Bifold to latest to resolve tests + linter.

### Description

Cross referenced the exsting theme in Bifold against the theme in BCWallet. Dropped properties that were exactly the same, and overrided the theme properties that were different.

This new ThemeBuilder will use all the properties from the base theme ie: `bifoldTheme` and override any differences we want.

### Diffs
I diff checked the output against main and both BCSC and BC-Wallet look the same
bc-wallet-theme-diff

Note: I did drop one property that was not being used anywhere.
bc-wallet-theme-dropped-property

### Testing Notes
The BCWallet project should look exactly the same as before. So if you run this PR and nothing looks different SUCCESS.
Will need to actually run this PR to verify that it looks the same.


bryce-mcmath:
Was going to attempt finishing this PR but even after importing the new Bifold packages the BC Wallet theme is broken for some reason. BCSC theme looks good but BC Wallet theme looks like this:
Screenshot 2025-08-05 at 4 11 40 PM

Gonna leave this PR as is for now

MacDeluca:
@bryce-mcmath
Running this with the latest Bifold looks as expected on my end. The test suites are also passing, the snapshots were previously failing when the theme was misconfigured.

image

bryce-mcmath (approved):
Re-tested, working well now. I must have used the wrong package version or something the first time I tried. Nice work 🎨

PR #2593, opened 2025-08-01
chore: cicd to support single app build

jleach:
Refactor the main workflow to include a dynamic application version based on the build number and add support for a new application ID for the BCSC development environment. Adjustments ensure proper handling of secrets and build configurations for both single app and BC Wallet builds.


patch/0001-chore-application-id-to-bcsc-dev.patch
```diff
@@ -0,0 +1,40 @@
...
+diff --git a/app/android/app/build.gradle b/app/android/app/build.gradle
+index 17d1efc6..a2fe13fc 100644
+--- a/app/android/app/build.gradle
++++ b/app/android/app/build.gradle
+@@ -95,7 +95,7 @@ android {
+     }
+ 
+     defaultConfig {
+-        applicationId "ca.bc.gov.BCWallet"
++        applicationId "ca.bc.gov.id.servicescard.dev"
```
cvarjao: can this be populated from build-time environment variable? e.g.
```
applicationId System.getenv("APP_ID") ?: "ca.bc.gov.BCWallet"
```
using [Groovy Elvis operator](https://groovy-lang.org/operators.html#_elvis_operator)

patch/0001-chore-application-id-to-bcsc-dev.patch
```diff
@@ -0,0 +1,40 @@
...
+         minSdkVersion rootProject.ext.minSdkVersion
+         targetSdkVersion rootProject.ext.targetSdkVersion
+         versionCode System.getenv("VERSION_CODE") ? System.getenv("VERSION_CODE").toInteger() : 1
+diff --git a/app/android/app/google-services.json b/app/android/app/google-services.json
+index a3a58790..13071b8f 100644
+--- a/app/android/app/google-services.json
++++ b/app/android/app/google-services.json
```
cvarjao: I am wondering if we need to make the content of this file configurable. We might need to add BCSC bundle ids to our existing firebase account. The docs recommends to download the file from the console:
https://firebase.google.com/docs/android/setup#add-config-file

jleach: @cvarjao The default one on the file system is the BC Wallet Firebase config and the patch just converts it to the one from BCSC to use that firebase when doing the BCSC build. I'm not manually editing them. I'm overwriting the BCW file with the BSCS file (both from the Google Console) and the patch just records the change.

patch/0001-chore-application-id-to-bcsc-dev.patch
jleach: My preference is to use the patch method. The key benefit is keeping our process consistent—this way, the steps to switch between a BCW and a BCSC build are exactly the same for both iOS and Android. Also it's I think this patch will only be needed for a few more months then we'll be running the Single App config exclusively. Thoughts? I'm happy to go either way.

cvarjao (dismissed):
This is good for now.  we don't have a plan to retire BC Wallet App yet, and there are still the many variations of BCSC which we will probably converge to 1. The only concern is with the hardcoded values, but we already have a ticket to address that

cvarjao (approved):

Issue #2579, opened 2025-07-25
Endpoints should only come from well known config endpoint


jleach:
### Acceptance Criteria

1. Not all things in that data are endpoints, the var name should be clarified i.e. `pairDeviceWithQRCodeSupported: true`
2. If any required URLs are not fetched from the `.well-known` then create tickets in Jira for the IAS back-end team to add them.
3. If we want to ship the app with some defaults then we should use the `baseUrl` to build them dynamically. I don't know if this is worth it because if we don't have internet access to fetch then defaults won't be useful.
4. **Before the next release, fetch of well known endpoint should only happen in BCSC Mode**

PR #2603, opened 2025-08-07
fix(deps): bcsc client fixes and package update

bryce-mcmath:
# Summary of Changes

- package updates include:
  - logging improvements
  - theme improvements
  - fix for overwritten preferences persisted state
 
- bcsc improvements include:
  - full logging outputs for errors so we can actually see why stuff is failing much more easily
  - "endpoints" separated into actual endpoints and config
  - dynamic default endpoints using baseURL
  - big effect in BCSC RootStack.tsx moved into a hook and well-known endpoint fetch moved into said hook from root app startup
  - memoize api hooks so they don't cause unnecessary re-renders

Tested fresh install verification flow on iOS, works just the same. I also used our updated label in our stale action

N/A

# Related Issues

#2579 

# Pull Request Checklist

### DIFF
--- a/app/src/bcsc-theme/hooks/useInitializeBcscApi.tsx
+++ b/app/src/bcsc-theme/hooks/useInitializeBcscApi.tsx
@@ -1,0 +1,85 @@
+import { DispatchAction, TOKENS, useServices, useStore } from '@bifold/core'
+import { useEffect, useState } from 'react'
+import { getToken, TokenType } from 'react-native-bcsc-core'
+
+import { BCDispatchAction, BCState } from '@/store'
+import client from '../api/client'
+import useApi from '../api/hooks/useApi'
+
+const useInitializeBcscApi = () => {
+  const [store, dispatch] = useStore<BCState>()
+  const { registration } = useApi()
+  const [logger] = useServices([TOKENS.UTIL_LOGGER])
+  const [loading, setLoading] = useState(true)
+
+  useEffect(() => {
+    setLoading(true)
+
+    if (!store.stateLoaded) {
+      return
+    }
+
+    const asyncEffect = async () => {
+      try {
+        await client.fetchEndpointsAndConfig(client.baseURL)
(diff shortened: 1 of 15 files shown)


app/src/bcsc-theme/hooks/useInitializeBcscApi.tsx
```diff
@@ -0,0 +1,85 @@
...
+        }
+      } catch (error) {
+        logger.error(`Error setting API client tokens: ${error}`)
+      } finally {
+        setLoading(false)
+      }
+    }
+
+    asyncEffect()
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
bryce-mcmath, line 79: in a later PR we should address this - if store.bcsc.refreshToken is included in the dep array, the "You're all set" screen will get skipped because this effect will rerun and navigate the user to the MainStack from the VerifyIdentityStack

app/src/bcsc-theme/hooks/useInitializeBcscApi.tsx
```diff
@@ -0,0 +1,85 @@
...
+          const tokenData = await client.getTokensForRefreshToken(token)
+          if (tokenData.bcsc_devices_count !== undefined) {
+            dispatch({
+              type: BCDispatchAction.UPDATE_DEVICE_COUNT,
+              payload: [tokenData.bcsc_devices_count],
+            })
+          }
+
+          dispatch({ type: BCDispatchAction.UPDATE_VERIFIED, payload: [true] })
+        }
```
travkhl, line 70: Should there be an else statement here to handle getToken failing and returning null?

bryce-mcmath: Yes. Right now we just handle happy path mostly and I didn't want to change any functionality in this PR, just do the api client cleanup so I just copied it over for the most part

travkhl (approved):

PR #2618, opened 2025-08-19
feat: add webvh support

brianorwhatever:
# Summary of Changes

- Added support for @credo-ts/webvh version 0.5.17-alpha-20250818210456 in package.json and yarn.lock.
- Introduced a temporary shim for compatibility between @credo-ts/core and @credo-ts/webvh.
- Updated metro.config.js to ensure proper resolution of @credo-ts/webvh dependencies.
- Modified bc-agent-modules to include WebVhAnonCredsRegistry and WebvhDidResolver for enhanced functionality.

There is currently a patch being applied that downgrades the webvh version of `@credo-ts/core` so that it is compatible with the version already installed. I also had to shim some of the newly exported functionality. These can go away once v0.5.17 is released and bc-wallet uses this version across the board

### DIFF
--- a/.yarn/patches/@credo-ts-webvh-npm-0.5.17-alpha-20250818210456-132526c9e9.patch
+++ b/.yarn/patches/@credo-ts-webvh-npm-0.5.17-alpha-20250818210456-132526c9e9.patch
@@ -1,0 +1,15 @@
+diff --git a/package.json b/package.json
+index 2bac336949dfc9bf0c4bc67b7800f4abf38a9d08..8abab78122903110b2dc44da72916f20322c1566 100644
+--- a/package.json
++++ b/package.json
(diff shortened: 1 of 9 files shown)


bryce-mcmath:
@brianorwhatever would it help if our team upgraded Bifold and BC Wallet and their patches to use 0.5.17 for all Credo deps first? Perhaps that way you wouldn't need some of these workarounds? Just a thought

brianorwhatever:
yeah, it would definitely be a good for all libraries to be on 0.5.17 as soon as possible and then I can remove bullet points 2 and 3

bryce-mcmath:
Closing in favour of https://github.com/bcgov/bc-wallet-mobile/pull/2675

Issue #2534, opened 2025-07-03
Setup Steps: Use Non-BCSC ID to Verify Identity


jeznorth:
As a user, I want to be able to use my Non-BCSC ID during the setup process to verify my identity if I don't have a BC Services Card available to use.

## Acceptance Criteria
- When in 'Step 2: Scan or Take Photos of your ID' of the app setup steps, users should be provided the option 'Setup up with other ID'.
- See existing wireframes for workflow or reference functionality in BCSC app.

## Design Specifications
**NOTE:** UX Team might want to do a pass on this flow; I currently can only find a representation of the form which is what I assume to be dependant on the type of ID the user choses to provide.

- https://www.figma.com/design/A664vlh8PpAyQ2i2YIq3Wh/Single-app-Phase-1?node-id=625-3991&m=dev&t=1lHQ8D4GEsaXsGH2-1

## Developer Notes
- Is there an API that has which fields/labels/decription to collect from each identity document type?
- Are all (3) verification methods available for non-BCSC registration?
- Is the logic to short list what documents goes with what? is that on the client side or server side?

bryce-mcmath:
works as expected

PR #2643, opened 2025-08-29
feat: create jwt for native bcsc core (#2534)

MacDeluca:
# Summary of Changes

Extends the BCSC Core API to include `createDeviceSignedJWT`, allowing the application to generate any shaped JWT token. This new function will be used to generate the id_token_hint JWT used in the device authorization flow.

N/A

# Related Issues
https://github.com/orgs/bcgov/projects/108/views/8?pane=issue&itemId=118415502&issue=bcgov%7Cbc-wallet-mobile%7C2534

# Pull Request Checklist


app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
```diff
@@ -26,21 +63,112 @@ const useAuthorizationApi = () => {
...
-      apiClient.logger.info(`Registration body: ${JSON.stringify(body, null, 2)}`)
+
+      apiClient.logger.info('useAuthorizationApi.authorizeDevice.body', body)
+
       const { data } = await apiClient.post<VerifyInPersonResponseData>(apiClient.endpoints.deviceAuthorization, body, {
         headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
         skipBearerAuth: true,
       })
+
+      // TODO (MD): Should we also check here if the request fails due to being previously registered?
```
al-rosenthal, line 74: does the API suggest it will throw if someone is already registered?

MacDeluca: From the docs: 
"invalid_registration_request: 
R3.8.0
 the device client registration is in an invalid state (i.e. already approved)"
 
 https://citz-cdt.atlassian.net/wiki/spaces/BMS/pages/301574688/5.1+System+Interfaces#Device-Authorization-Request
 
 The helper function I made checks for this type of error. Not certain what other state will trigger an "invalid_registration_request" on the IAS side, so I made it a little more specific to check the error description as well. But.... I think there is a spelling mistake in the current error description ie: "statue" vs "state" which this handles both.

app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
```diff
@@ -26,21 +63,112 @@ const useAuthorizationApi = () => {
...
+            given_name: config.firstName,
+            birthdate: config.birthdate,
+            address: {
+              street_address: config.address.streetAddress,
+              postal_code: config.address.postalCode,
+              locality: config.address.city,
+              region: config.address.province,
+              country: 'CA',
+            },
+            gender: config.gender ?? 'unknown',
```
jleach, line 112: Is `unknown` what the API expects as a string when it is unknown? Wonder if we should document that somewhere - even with just a comment?

jleach (approved):
LGTM. Most of the changes apper to be automated. Will we run into format issues if one developer uses xcode or Android Studio and another uses VSCode which may decide to format differently?

MacDeluca:
> LGTM. Most of the changes apper to be automated. Will we run into format issues if one developer uses xcode or Android Studio and another uses VSCode which may decide to format differently?
@jleach 

Yea I'm using a different editor than most. I'm assuming swift / kotlin both have formatting either built into the LSP or as external tools similar to prettier in Node. 

Probably best to do something with a github action that detects if the Kotlin/Swift code is formatted correctly, and include the scripts in the package.json to auto fix the issues for both android and ios directories.

That way the formatting is editor agnostic, and the Github action can just notify when it's formatted incorrectly.

app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
MacDeluca: Correct, unknown is what it expects when it is undefined, it's documented in their API design. I can add a comment that explains whats happening there.

Issue #2609, opened 2025-08-11
Long white screen before load on certain Android devices


bryce-mcmath:
### Description of problem
Discovered by @loneil, on some Android devices (confirmed on Pixel 8 with Android 15) there is a pause on a white screen before the app loads. The length of time is inconsistent and ranges from less than a couple seconds to twenty(!) seconds. There doesn't seem to be any specific configuration of the app that causes it to be a shorter or longer wait

### Expected behavior
Very brief, if any, time before app load

### Steps to reproduce
Load the app on a Pixel 8

### Screenshots and/or log output
https://github.com/user-attachments/assets/a1b8f7b7-10ea-43c3-a9be-2827746f1187

### Developer notes
One likely cause is improper setup for react-native-splash-screen on Android.
- [Docs](https://github.com/crazycodeboy/react-native-splash-screen?tab=readme-ov-file#manual-installation)
- [Noteworthy](https://github.com/crazycodeboy/react-native-splash-screen/issues/338#issuecomment-447251703)

### Environment
- [x] Occurs on Android
- [ ] Occurs on iOS

Build #:
1.0.25 - 2533

Android Device Model: 
Pixel 8 Android 15

iOS Device Model: 
N/A

### Workaround
N/A

### Severity
- [ ] High
- [x] Medium
- [ ] Low

al-rosenthal:
The splash screen on android isn't configured properly, which is why we are seeing a plain white screen. Still investigating the lengthy load times

jleach:
- Pixel 7a Android 16 works fine. No Delay. Tested main (54bc1ee)
- Pixel 8 Android 16 works fine. No Delay. Tested main (54bc1ee)

PR #2644, opened 2025-09-02
fix: android start up time

al-rosenthal:
# Summary of Changes
On some android devices, the bundled js was large enough to cause start up problems, steps taken to reduce bundled js size

- updated `app/bundle.gradle` to include proguard, which helps minify android code during builds
- updated `metro.config` to build with  `metro-minify-terser` to minify the bundled js
- added clean up to console.logs in production for additional package size reduction
- added default values for keystore for local builds

android-start-up-fix

this gif was captured with the an apk generated locally with the changes and installed directly onto the phone. 

# Related Issues

[BC-Wallet: 2609](https://github.com/bcgov/bc-wallet-mobile/issues/2609)

# Pull Request Checklist


app/android/app/build.gradle
```diff
@@ -75,7 +75,7 @@ afterEvaluate {
 /**
  * Run Proguard to shrink the Java bytecode in release builds.
  */
-def enableProguardInReleaseBuilds = false
+def enableProguardInReleaseBuilds = true
```
bryce-mcmath: Able to confirm Person credential flow still works with this turned on?

jleach: @bryce-mcmath Proguard does optimization and shrinking for a bundle in Android so I don't think it will impact app functionality. It should be on where possible IMHO.

EDIT: I doubt it will shrinking or optimize our JS code. Just Android native. We have it turned on in BCSC.

bryce-mcmath: Asked this cause a lot of the Android React Native dependency docs required adding certain proguard exclusions to make things work if you had proguard turned on, and our attestation package uses some Android dependencies. Should be fine, just might be worth a test.

al-rosenthal: @jleach  After testing the person credential flow with pro guard enabled it fails at a strange step. The app still initializes and the camera/ QR code scanning seems to work but that person workflow stops at the `Connecting to agent...` which is the very first step. Disabling progaurd fixes that issue and does not affect start up time.

bryce-mcmath (approved):

Issue #2645, opened 2025-09-03
Android Single App Image Preview


al-rosenthal:
### Description of problem
In the single app on Android, when taking selfies for verification, the previews are rotated 90 degrees

### Expected behavior
The selfie photo should look like it does on the previous preview screen (properly oriented)

### Steps to reproduce
- Install the app
- Go into single app mode
- Go through verification steps
- Verify by video
- Take selfie

### Environment
- [x] Occurs on Android
- [ ] Occurs on iOS

Build #:

Android Device Model: 

iOS Device Model: 

### Severity
- [ ] High
- [x] Medium
- [ ] Low

bryce-mcmath:
still rotated in latest main, Samsung Galaxy A53

bryce-mcmath:
None of the images appear rotated now on either platform

Issue #2577, opened 2025-07-25
Date picker does not register date correctly


jleach:
### Description of problem

The date picker on iOS does not attempt to register the date that was selected as user input. In the attached video Jan 3, 1990 is selected in the date picker but Jan 3, 2015 is what is displayed on the following screen, and presumably, what was sent to the API for verification.

### Expected behavior

The application should use the date selected by the user.

### Steps to reproduce

1. Fresh install
2. Register with combo card
3. Enter serial number
4. Enter date
5. Continue

### Screenshots and/or log output

https://github.com/user-attachments/assets/6ba0377b-a0c2-4bfa-a464-d325dfbf87ed

### Environment
- [x] Occurs on Android
- [x] Occurs on iOS

Build #:

Main on Jul 25.

Android Device Model: 

iOS Device Model: 

### Severity
- [ ] High
- [ ] Medium
- [ ] Low

al-rosenthal:
An alternate date picker library: https://www.npmjs.com/package/@react-native-community/datetimepicker
It is more popular than the one we currently use, it is also updated regularly

al-rosenthal:
I tried reproducing this today on my iPhone and couldn't, @jleach when you get the time can you try this out again

jeznorth:
Moving this back to the Sprint Backlog until it can be reproduced consistently.

MacDeluca:
Confirmed working.

Issue #2485, opened 2025-06-05
BcscCore should allow for the environment / provider to be changed.


jleach:
### Description of feature / user story

When the environment is changed from BC Wallet -> Developer -> IAS Environment the "env" should be passed to BcscCore so that it pulls in the correct information.

### Acceptance Criteria

BcscCore will read files for the correct environment when it changes in developer settings.

### Value / Outcome
This enable the QA team and testers to test the app against multiple environments

bryce-mcmath:
To test: change IAS Environment and check logs - you should see a new log for the `.well-known` endpoint fetch

PR #2665, opened 2025-09-15
feat: reactive api client

bryce-mcmath:
# Summary of Changes

Makes the API client reactive to whichever IAS Environment is set in developer settings

Notes:
- IAS_URL is no longer needed
- prod -> idqa, test -> idsit, dev -> iddev. prod is default, so you will need to switch to have the most success while developing
- renamed the client and some hooks to be more accurate
- fix for date picker
- fix for rotated android images
- also ran the gradle:write-locks script since some new deps were added / removed recently, picked up some changes

N/A

# Related Issues

#2485
#2645
#2577

# Pull Request Checklist


app/src/bcsc-theme/hooks/useInitializeBCSC.tsx
```diff
@@ -0,0 +1,71 @@
...
+        let token
+        if (!store.bcsc.refreshToken) {
+          const tokenInfo = await getToken(TokenType.Refresh)
+          token = tokenInfo?.token
+          dispatch({ type: BCDispatchAction.UPDATE_REFRESH_TOKEN, payload: [token] })
+        } else {
+          token = store.bcsc.refreshToken
+        }
+
+        if (token) {
```
MacDeluca, line 48: question: what happens if token is undefined?

bryce-mcmath: `verified` does not become true -> `loading` becomes false -> the user sees the `<VerifyIdentityStack />` rather than the `<BCSCMainStack />` (logged-in view)

app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
```diff
@@ -46,6 +46,8 @@ export interface AuthorizeDeviceUnknownBCSCConfig {
 }
 
 const useAuthorizationApi = () => {
```
MacDeluca: thought: injecting the client into these hooks might make testing easier

The `useApi` hook could initialize at the top and inject into all the children hooks, keeping the existing destructuring pattern the same for the components.

```ts
const { metadata } = useApi()
```

app/src/bcsc-theme/hooks/useInitializeBCSC.tsx
bryce-mcmath: essentially it just means they have not been verified

MacDeluca: I think a comment explaining that would be helpful

app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
MacDeluca: I really don't lean one way or the other, mostly just thinking out loud. 

If testing these isn't that much different I think this is great.

app/src/bcsc-theme/api/hooks/usePairingApi.tsx
```diff
@@ -11,20 +11,25 @@ export interface PairingCodeLoginResponseData {
...
-        `${apiClient.baseURL}/cardtap/v3/mobile/assertion`,
-        { assertion: signedCode },
-        { headers: { 'Content-Type': 'application/x-www-form-urlencoded' } }
-      )
-      return { success: true }
-    })
-  }, [])
+  const apiClient = useBCSCApiClient()
+
+  const loginByPairingCode = useCallback(
```
MacDeluca, line 14: nit: jsdoc / tsdoc comment

app/src/bcsc-theme/hooks/useBCSCApiClient.tsx
```diff
@@ -0,0 +1,33 @@
+import { useContext } from 'react'
+import BCSCApiClient from '../api/client'
+import { BCSCApiClientContext } from '../contexts/BCSCApiClientContext'
+
+// Hook to be used for API calls, will throw if not configured properly
+export const useBCSCApiClient = (): BCSCApiClient => {
```
MacDeluca, line 6: awesome ❤️

app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
bryce-mcmath: Do you mean something like this:

useApi.tsx
```
const useApi = () => {
  const client = useBCSCApiClient()
  const config = useConfigApi(client)
  ...
}
```

useConfigApi.tsx
```
const useConfigApi = (client: BCSCApiClient) => {
  const getServerStatus = useCallback(async () => {
    const { data } = await client.get(...)
    ...
  }, [client])
  ...
}

**Review by al-rosenthal — COMMENTED** [review]
I'm seeing a few issues with the date picker on android. It still sometimes selects the wrong date (off by one) but also now doesn't 'settle' like it should. I'll stop the picker on a date and it will jump to different day once my finger is off the device. I'll try and capture a video of that

File: app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx [hunk 5]
```diff
@@ -39,10 +39,19 @@ export interface RegistrationResponseData {
...
 }
 
+/**
+ * This hook uses useBCSCApiClientState (safe) instead of useBCSCApiClient (throwing)
+ * to avoid "client not ready" errors during app initialization
+ */
 const useRegistrationApi = () => {
-  const [state, dispatch] = useStore<BCState>()
+  const [store, dispatch] = useStore<BCState>()
+  const { client: apiClient, isReady } = useBCSCApiClientState()
```
jleach: `isClientRead` might be more descriptive over the long term. Why no - characters are free to use.

app/src/bcsc-theme/api/hooks/useTokens.tsx
```diff
@@ -19,20 +20,24 @@ export interface BcscJwtPayload {
...
+          headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+          skipBearerAuth: true,
+        })
+        apiClient.tokens = data
 
-      const bcsc_devices_count = await getDeviceCountFromIdToken(data.id_token, apiClient.logger)
-      return { ...data, bcsc_devices_count }
-    })
-  }, [])
+        const bcsc_devices_count = await getDeviceCountFromIdToken(data.id_token, apiClient.logger)
```
jleach, line 33: Camel case? I don't think we use snake case in the app.

app/src/bcsc-theme/api/hooks/usePairingApi.tsx
```diff
@@ -11,20 +11,25 @@ export interface PairingCodeLoginResponseData {
...
+    async (code: string) => {
+      return withAccount<PairingCodeLoginResponseData>(async (account) => {
+        const { issuer, clientID } = account
+        const { fcmDeviceToken, apnsToken } = await getNotificationTokens()
+        const signedCode = await signPairingCode(code, issuer, clientID, fcmDeviceToken, apnsToken)
+        await apiClient.post<PairingCodeLoginResponseData>(
+          // this endpoint is not available through the .well-known/openid-configuration so it needs to be hardcoded
+          `${apiClient.baseURL}/cardtap/v3/mobile/assertion`,
```
jleach, line 22: Do we know why it's not available? I'll mention to them we should add it unless there is a good reason not to.

app/src/bcsc-theme/contexts/BCSCApiClientContext.tsx
```diff
@@ -0,0 +1,75 @@
...
+      setError(null)
+
+      try {
+        const newClient = new BCSCApiClient(store.developer.environment.iasApiBaseUrl, logger as RemoteLogger)
+        await newClient.fetchEndpointsAndConfig(store.developer.environment.iasApiBaseUrl)
+
+        setClient(newClient)
+        setIsReady(true)
+      } catch (err) {
+        const errorMessage = `Failed to configure BCSC client for ${store.developer.environment.name}: ${err}`
```
jleach: Will this print `Failed to configure BCSC client for staging: [object Error]`? You might want `err.message`.

app/src/bcsc-theme/hooks/useBCSCApiClient.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const context = useContext(BCSCApiClientContext)
+
+  if (!context) {
+    throw new Error('useBCSCApiClient must be used within a BCSCClientProvider')
+  }
+
+  if (context.error) {
+    throw new Error(`BCSC client error: ${context.error}`)
```
jleach, line 14: Is error a string or an object? You might get `[Object object]`.

app/src/bcsc-theme/hooks/useInitializeBCSC.tsx
```diff
@@ -0,0 +1,71 @@
...
+const useInitializeBCSC = () => {
+  const [store, dispatch] = useStore<BCState>()
+  const { client, isReady } = useBCSCApiClientState()
+  const { register } = useRegistrationApi()
+  const [logger] = useServices([TOKENS.UTIL_LOGGER])
+  const [loading, setLoading] = useState(true)
+  const initializedRef = useRef(false)
+
+  useEffect(() => {
+    if (!store.stateLoaded || !isReady || !client || initializedRef.current) {
```
jleach: I wonder if we could make this var more clear so devs don't need to go figure out what it means? `initializedRef.current`? Maybe `initalizedClientRef`?

app/container-imp.ts
```diff
@@ -366,8 +367,10 @@ export class AppContainer implements Container {
       ])
 
       // Convert date string to Date object (async-storage converts Dates to strings)
+      // Use timezone-safe parsing to prevent off-by-one date errors (consistent with date picker)
       if (typeof bcsc.birthdate === 'string') {
-        bcsc.birthdate = new Date(Date.parse(bcsc.birthdate))
+        const dateOnly = moment(bcsc.birthdate).format('YYYY-MM-DD')
```
jleach: I feel like we need to move that date format `YYYY-MM-DD` into const so that we can change it in one place. Not now, just at some point in the future.

app/src/bcsc-theme/api/hooks/usePairingApi.tsx
bryce-mcmath: I don't know why exactly but none of the `cardtap/` endpoints seem to be included in that `.well-known` response

app/src/bcsc-theme/hooks/useBCSCApiClient.tsx
bryce-mcmath: Here it is just a string

app/src/bcsc-theme/api/hooks/useTokens.tsx
bryce-mcmath: Should be changed, but it is currently used in this form in over 20 different places, so will leave that for a later PR to keep this one focused on the reactive api client

app/src/bcsc-theme/api/hooks/useAuthorizationApi.tsx
MacDeluca: yea exactly, that would make testing useConfigApi much easier as we would be able to inject a mock client

jleach:
> I'm seeing a few issues with the date picker on android. It still sometimes selects the wrong date (off by one) but also now doesn't 'settle' like it should. I'll stop the picker on a date and it will jump to different day once my finger is off the device. I'll try and capture a video of that

@al-rosenthal This is a known issue [2577](https://github.com/bcgov/bc-wallet-mobile/issues/2577). I too have also seen it on Android. You  could add to 2577.

bryce-mcmath:
> > I'm seeing a few issues with the date picker on android. It still sometimes selects the wrong date (off by one) but also now doesn't 'settle' like it should. I'll stop the picker on a date and it will jump to different day once my finger is off the device. I'll try and capture a video of that
> 
> @al-rosenthal This is a known issue [2577](https://github.com/bcgov/bc-wallet-mobile/issues/2577). I too have also seen it on Android. You could add to 2577.


Was able to repro the same bug, found a simple fix, will push up as part of this PR in a bit here

app/src/bcsc-theme/api/hooks/usePairingApi.tsx
bryce-mcmath: There's a lot of other missing JSDoc comments too, I'll add all of them in a subsequent PR to prevent this one from getting too beastly

bryce-mcmath: (The function isn't net new from this PR)

jleach (approved):

Issue #2718, opened 2025-09-29
Update build strategy to reduce reliance on patching


jleach:
### Description of feature / user story

A few of us have had to rebuilt the patches for the single app which can be frustrating and time consuming. To reduce reliance on patching we propose creating an environment with a DEFAULT_BLAH that will determing if the default behaviour is BCSC or BCWALLET. This will take care of TypeScript code only to start and we weill rely on patching for native mechanics only: Android Studio and XCode config.

### Acceptance Criteria

- [ ] Only have two patches in the app that patch project mechanics only;
- [ ] A new variable in .env that is respected by the app
- [ ] The CICD has the proper changes to account for the new .env setting.

### Wireframes or relevant image assets / links

PR #2737, opened 2025-10-03
fix: replace git patch for build_target env #2718

MacDeluca:
# Summary of Changes

This PR removes a Git patch [0003-chore-common-default-mode-to-bcsc.patch](https://github.com/bcgov/bc-wallet-mobile/pull/2737/files#diff-4891102f62f8829b7a68a1821e939bcb7cc6e650f8ef5dafcbc0f18c1ea76c16) and replaces with a new env BUILD_TARGET. This value is used to dynamically set the initial mode and theme of the application.

### Testing notes
- Add the new env BUILD_TARGET to .env
- Set to either: bcsc or bcwallet (set to bcsc to see changes)
- Remove the application and re-build
- Application should initialize as the env you set for BUILD_TARGET.

- Background the application, should be the same theme and mode when coming back into foreground

# Related Issues

# Pull Request Checklist


MacDeluca:
Found one issue, currently fixing.

app/src/store.tsx
```diff
@@ -249,7 +250,7 @@ export const initialState: BCState = {
   developer: developerState,
   dismissPersonCredentialOffer: dismissPersonCredentialOfferState,
   bcsc: bcscState,
-  mode: Mode.BCWallet,
+  mode: Config.BUILD_TARGET === Mode.BCSC ? Mode.BCSC : Mode.BCWallet,
```
jleach, line 253: Does this need to be updated when in BC Wallet one switches mode from Settings?

MacDeluca: This value is just for the initial state. So when we run a fresh build it will grab this value and then the store tracks and manages the value after.

jleach (approved):

PR #2775, opened 2025-10-10
feat(bcsc): account nickname

KjartanE:
# Summary of Changes

Core account creation and application startup sequence to enable multi-account support via user-defined nicknames.

image

image

image

image

# Related Issues

# Pull Request Checklist

- [ ] All commits contain a DCO `Signed-off-by` line (we use the [DCO GitHub app](https://github.com/apps/dco) to enforce this)
- [ ] Related issues are included under the Related Issues section above
- [ ] If applicable, screenshots, gifs, or video are included for UI changes


app/src/bcsc-theme/features/account/components/NicknameForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+    secondButton: {
+      marginTop: Spacing.sm,
+    },
+  })
+
+  const handleChangeText = (text: string) => {
+    setAccountNickname(text)
+  }
+
+  const handleContinuePressed = () => {
```
jleach: I think you'll want to wrap these in `useCallback`.

app/src/bcsc-theme/features/account/components/NicknameForm.tsx
```diff
@@ -0,0 +1,137 @@
...
+    if (accountNickname.length < 1) {
```
jleach: These'll get flagged as [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)) better to use minimumLength etc.

app/src/bcsc-theme/features/verify/SetupStepsScreen.tsx
```diff
@@ -215,11 +238,25 @@ const SetupStepsScreen: React.FC<SetupStepsScreenProps> = ({ navigation }) => {
...
+            navigation.navigate(BCSCScreens.NicknameAccount)
+          }}
+        />
+
+        <View style={styles.itemSeparator} />
+
+        {/* SETUP STEP 2: Identification submission */}
+
+        <SetupStep
+          title={'Step 2'}
```
jleach: Let's start adding these to translations.

app/src/bcsc-theme/features/account/RenameAccountScreen.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const navigation = useNavigation()
+  const { agent } = useAgent()
+
+  const onCancel = useCallback(() => {
+    navigation.goBack()
+  }, [navigation])
+
+  const onSubmitSuccess = useCallback(
+    (name: string) => {
+      agent.config.label = name
```
MacDeluca: question: What is the purpose of this?

KjartanE: unused, removed

app/src/store.tsx
```diff
@@ -322,6 +328,20 @@ const bcReducer = (state: BCState, action: ReducerAction<BCDispatchAction>): BCS
...
+      PersistentStorage.storeValueForKey<BCSCState>(BCLocalStorageKeys.BCSC, bcsc)
+      return newState
+    }
+    case BCSCDispatchAction.SELECT_ACCOUNT: {
+      const selectedAccountIndex = (action?.payload || []).pop() ?? 0
+      const bcsc = { ...state.bcsc, selectedAccountIndex }
+      const newState = { ...state, bcsc }
+      // don't persist, should be assigned every app start
+      return newState
+    }
```
MacDeluca, line 347: suggestion: if we used a Set for the nicknames we wouldn't need to use an index for the selected account.

We could select the account by nickname instead of index

KjartanE: Great suggestion I like this as a solution alot! Implemented.

app/src/localization/en/index.ts
MacDeluca, line 1: nit: can you add the same content to the french and portuguese files. They don't need to be translated, but can be tagged instead. Will make it easier when we audit these values before release.

ie: "AccountName": "Account Nickname (FR)"

MacDeluca (changes requested):
bug fix: Entering no content for the nickname works correctly (renders help text). Entering an empty space " " crashes the app.

error: `state.bcsc.nicknames.has is not a function (it is undefined)`

Probably just needs to be trimmed before validating.

app/src/localization/en/index.ts
```diff
@@ -269,7 +269,17 @@ const translation = {
...
       "Step3": "Step 3",
-      "Step4": "Step 4"
+      "Step4": "Step 4",
+      "Step5": "Step 5",
+    },
+    "NicknameAccount": {
+      "AccountName": "Account Nickname",
+      "CreateAccountName": "Create a nickname for your account on this device. It identifies your account:",
+      "AccountNameDescription1": "• When you open the app",
+      "AccountNameDescription2": "• In notifications",
```
MacDeluca: suggestion: I think we should use HTML for the bullets instead of text

MacDeluca (approved):

PR #2822, opened 2025-10-28
fix: refresh tokens race condition v2

MacDeluca:
# Summary of Changes

Missed a race condition with the first PR.

This PR fixes the race condition with requests fetching a new access token while other requests simultaneously requesting data from IAS.

Senario:

1. Call 1: We call getTokensForRefreshToken - fetches token and sets internal state
2. Call 2: While Call 1 is resolving but before it has set internal state we make a request to IAS with stale access token.
3. Call 2: Fails with invalid_access_token, even though the class has the correct access token in state (race condition)

The fix:
If any outgoing request to IAS has tokens refreshing or resolving, await tokens promise and use as current tokens.

# Testing Instructions

Refreshing the application, or removing the running app and restarting should eventually trigger the race condtion. 

remove these lines client.ts line 156-159 or swap with this to trigger race condition.
```ts
      if (this.tokensPromise) {
        // return the existing promise if currently refreshing
        return this.tokens as any // this.tokensPromise
      }

```

# Acceptance Criteria

Invalid access token race condition should never display in the logs.

# Related Issues
https://github.com/orgs/bcgov/projects/108?pane=issue&itemId=135457496&issue=bcgov%7Cbc-wallet-mobile%7C2811
# Pull Request Checklist

### DIFF
--- a/app/src/bcsc-theme/api/client.ts
+++ b/app/src/bcsc-theme/api/client.ts
@@ -46,7 +46,6 @@ class BCSCApiClient {
   endpoints: BCSCEndpoints
   config: BCSCConfig
   baseURL: string
-  // TODO (MD): Persist tokens securely using PersistentStorage ie: loadTokens(), saveTokens()
   tokens?: TokenResponse // this token will be used to interact and access data from IAS servers
   tokensPromise: Promise<TokenResponse> | null // to prevent multiple simultaneous token fetches
 
@@ -152,16 +151,26 @@ class BCSCApiClient {
     }
   }
 
-  async fetchAccessToken(): Promise<TokenResponse> {
+  async ensureValidTokens(): Promise<TokenResponse> {
     return withAccount(async () => {
+      if (this.tokensPromise) {
+        // Return the existing promise if currently refreshing
+        return this.tokensPromise
+      }
+
       if (!this.tokens) {
         // initialize tokens using `getTokensForRefreshToken`
-        throw new Error('Client tokens not initialized')
+        throw new Error('Client missing tokens')
       }
 
       if (this.isTokenExpired(this.tokens.refresh_token)) {
         // note: refresh tokens should not expire
         throw new Error('Refresh token expired')
+      }
+
+      if (!this.isTokenExpired(this.tokens.access_token)) {
+        // access token is still valid don't refresh
+        return this.tokens
       }
 
       // access token is expired, fetch new tokens using refresh token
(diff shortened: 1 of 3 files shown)


app/src/bcsc-theme/features/settings/ContactUsScreen.tsx
MacDeluca, line 1: This file must have squeaked through on a previous manual merge?

app/src/bcsc-theme/api/client.ts
```diff
@@ -152,18 +151,28 @@ class BCSCApiClient {
     }
   }
 
-  async fetchAccessToken(): Promise<TokenResponse> {
+  private async ensureValidTokens(): Promise<TokenResponse> {
     return withAccount(async () => {
+      if (this.tokensPromise) {
+        // return the existing promise if currently refreshing
+        return this.tokensPromise
+      }
```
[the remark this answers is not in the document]
MacDeluca: I think this is missing context. It's not taking into account line: 221 or how the interceptor works.

app/src/bcsc-theme/api/client.ts
```diff
@@ -152,18 +151,28 @@ class BCSCApiClient {
...
+
       if (!this.tokens) {
         // initialize tokens using `getTokensForRefreshToken`
-        throw new Error('Client tokens not initialized')
+        throw new Error('Client missing tokens')
```
jleach, line 164: Do we need to log.error?

app/src/bcsc-theme/api/client.ts
```diff
@@ -188,46 +197,43 @@ class BCSCApiClient {
       return config
     }
 
-    if (!this.tokens || !this.isTokenExpired(this.tokens.access_token)) {
-      this.tokens = await this.fetchAccessToken()
-    }
+    const tokens = await this.ensureValidTokens()
```
jleach, line 202: try/catch/log?

jleach (commented):
Just a few questions.

app/src/bcsc-theme/api/client.ts
MacDeluca: Probably doesnt hurt. 
This is an error that the developer will see if they try and make a request before properly configuring the tokens with `getTokensForRefreshToken`.

It's called in initialization, and the tokens will always be defined after.

app/src/bcsc-theme/api/client.ts
MacDeluca: Similar to the above comment, I think we wan't this to explode if ensureValidTokens is unable to retrieve tokens but the developer didnt explicitly skip this step with `skipBearerAuth`

I'll include some error logs inside ensureValidTokens

jleach:
@MacDeluca One Sonarqube issue.

jleach (approved):

Issue #2724, opened 2025-09-29
Fix camera focus


jleach:
### Description of problem

The application camera isn't focusing correctly during evidence collection. Fix the camera focus implementation so the app can reliably capture clear images/video. Pickup the user touch and use it to focus the camera on the ID / Card. The React Native Vision Camera docs have an example of how to do this. Check in Bifold if you like to see if it does it also.

### Environment
- [x] Occurs on Android
- [x] Occurs on iOS

Build #:

Android Device Model: 

iOS Device Model: 

### Severity
- [ ] High
- [ ] Medium
- [ ] Low

PR #2898, opened 2025-11-14
feat: add tap to focus to camera

travkhl:
# Summary of Changes

This PR adds tap to focus to the following areas of the app
- ScanSerialScreen
- TransferQRScannerScreen

# Testing Instructions

1. Navigate to a screen with the camera component (scan serial in verify or transfer qr in account setup).
2. Tap on the camera view
3. Verify if the camera properly changes focus based on the area tapped.

# Acceptance Criteria

User's can tap to focus on the camera everywhere it is present in the app.

[scan-barcode.webm](https://github.com/user-attachments/assets/45d85246-66ef-471e-8660-819fd493ac46)

[transfer-account.webm](https://github.com/user-attachments/assets/5ac44f9f-1963-4c45-bf36-8f0c7a3db926)

# Related Issues

https://github.com/bcgov/bc-wallet-mobile/issues/2724

# Pull Request Checklist


bryce-mcmath (dismissed):
Looks good! One non-blocking question, what made you decide to use useCallback for some functions but not others?

travkhl:
> Looks good! One non-blocking question, what made you decide to use useCallback for some functions but not others?

@bryce-mcmath I'm pretty sure that useCallback on focus() is an artifact from development when there was something in that dependency array. I don't think it's necessary anymore. Do you think it should be removed?

bryce-mcmath:
If there's no reactive dependencies and the linter is good with it, removing it would be fine I think. But I'm good with it either way

bryce-mcmath:
Maybe throw in upgrading the setup-java action to v5 to see if it fixes the build issue?

bryce-mcmath:
Could try disabling setup java cache? Seems to be a resource issue (probably cause we're doing four separate builds simultaneously

app/src/bcsc-theme/components/CodeScanningCamera.tsx
```diff
@@ -204,8 +267,27 @@ const CodeScanningCamera: React.FC<CodeScanningCameraProps> = ({
...
+              opacity: focusOpacity,
+              transform: [{ scale: focusScale }],
+            },
+          ]}
+        />
+      )}
       {/* scan area cutout */}
-      <View style={styles.overlayContainer}>
+      <View style={styles.overlayContainer} pointerEvents="none">
         <View style={styles.overlayTop} />
```
al-rosenthal, line 291: It might be a good idea to replace some of the overlay views here with the SVGOverlay, just so we keep things consistent within the app. `TransferQRScannerScreen` shows a good example of that and it would clean up this component a bit.

travkhl: Good suggestion, it would be a nice simplifying change but I do not think it can be done in this PR. SVGOverlay is incompatible with this screen. Either it needs to be updated to accept a size instead of always using the window dimensions or the design of this screen will be changed completely to stray from the wireframe.

bryce-mcmath (approved):
woo the builds work again

Issue #2932, opened 2025-11-19
Crash (OOM) when reading large files using RNFSManager.readFile


jleach:
### Description of problem

The Android application crashes with an Out of Memory (OOM) exception when attempting to read large files (approximately 124 MB) into memory using RNFSManager.readFile. This occurs because the method attempts to load the entire file content into the heap at once, exceeding the available memory limits of the device. As a result, the operation fails with a "Failed to allocate" error, causing the app to force close.

Note the top line in the screen shot `RNFSManager.readFile`. It's probably trying to read the entire file into memory to send to the API. When it base64 encodes it that probably greatly expands the file size and thus memory footprint.

### Steps to reproduce

On my Pixel 7a I can record videos of 5-8 seconds no problem. When I create them around the 10+ second mark the app either quits or shows the error in the screen shot below.

### Screenshots and/or log output

<img width="200" alt="Image" src="https://github.com/user-attachments/assets/ab0a07c8-7182-4ce7-8455-9ad14aaa67ce" />

### Environment
- [x] Occurs on Android
- [ ] Occurs on iOS

Build #:

Android Device Model: 

Pixel 7a

iOS Device Model: 

### Severity
- [ ] High
- [ ] Medium
- [ ] Low

travkhl:
Cannot recreate on ios or android emulator with 128MB ram. It seems fixed.

PR #2969, opened 2025-11-25
fix(send video): invalid state and cache cleanup

MacDeluca:
# Summary of Changes

Waiting on #2932 

This PR fixes some state issues when going through the "Send Video" verification process.

# Testing Instructions

- Verify by send video
- Send video and photo to IAS
- Click "Choose another way to verify" and cancel verification
- Send video and photo to IAS (should complete)

# Acceptance Criteria

# Related Issues

https://github.com/orgs/bcgov/projects/108/views/8?pane=issue&itemId=139237176&issue=bcgov%7Cbc-wallet-mobile%7C2895


app/src/bcsc-theme/features/verify/VerificationMethodSelectionScreen.tsx
```diff
@@ -34,17 +36,35 @@ const VerificationMethodSelectionScreen = ({ navigation }: VerificationMethodSel
   const handlePressSendVideo = useCallback(async () => {
     try {
       setSendVideoLoading(true)
+
       const { sha256, id, prompts } = await evidence.createVerificationRequest()
+
+      await Promise.all([
```
bryce-mcmath: Question: would it better to use Promise.allSettled so we can do as much as possible and see if more than one of the promises failed, rather than fail early if any fail?

bryce-mcmath (commented):
Question about the behavior of this new cache: does it result in videos being persisted between full app exits + restarts? Even if unseen to the user?

app/ios/AriesBifold.xcodeproj/project.pbxproj
```diff
@@ -550,13 +550,13 @@
 					"$(inherited)",
 					"$(SDKROOT)/usr/lib/swift",
 				);
-				MARKETING_VERSION = 1.0.1;
+				MARKETING_VERSION = 3.15.0;
```
al-rosenthal: not sure we want this stuff pushed up

MacDeluca: already removed (added by mistake)

app/src/bcsc-theme/features/verify/VerificationMethodSelectionScreen.tsx
MacDeluca: Yea I think that would probably be a better approach.

Changed.

MacDeluca:
@bryce-mcmath 
> Question about the behavior of this new cache: does it result in videos being persisted between full app exits + restarts? Even if unseen to the user?

No it does not. This just helps with reducing disk fetching during this workflow.

It does prevent an additional disk fetch before sending the media metadata off to IAS. For a 15 second video this resulted in a time savings of 8ish seconds for the user.

bryce-mcmath:
> For a 15 second video this resulted in a time savings of 8ish seconds for the user.

😍

bryce-mcmath (dismissed):
🏎💨

al-rosenthal (commented):
ran this locally and it works as expected

al-rosenthal (approved):

PR #2992, opened 2025-12-01
fix: back button hit slop

MacDeluca:
# Summary of Changes

This PR creates a new component for the back button which now includes hit slop.

# Testing Instructions

Header back buttons should work as expected.

# Acceptance Criteria

All back buttons work (navigate back)
No back buttons missing

# Related Issues

https://github.com/orgs/bcgov/projects/108/views/1?filterQuery=-is%3Apr+-status%3A%22Eternal+Icebox%22+uat&pane=issue&itemId=141417979&issue=bcgov%7Cbc-wallet-mobile%7C2968


app/src/bcsc-theme/navigators/MainStack.tsx
```diff
@@ -64,6 +65,8 @@ const MainStack: React.FC = () => {
           title: '',
           headerBackTestID: testIdWithKey('Back'),
           headerShadowVisible: false,
+          headerTitleContainerStyle: { maxWidth: HEADER_TITLE_MAX_WIDTH },
```
MacDeluca: note: This change prevents the header title from overriding the header left hit slop

previous default (bifold): 
```ts
      headerTitleContainerStyle: {
        flexShrink: 1,
        maxWidth: '68%',
        width: '100%',
      }

```

app/src/utils/os-utils.ts
bryce-mcmath, line 1: This is kind of reimplementing Platform.select, no?

MacDeluca: Yea this is pretty much the same... I didn't know that existed, neat!

I will swap to use Platform.select

app/src/bcsc-theme/navigators/MainStack.tsx
bryce-mcmath: Makes sense, but are we sure we don't want the flexShrink still?

MacDeluca: Very possible, when I was tweaking this I set all the background colours to see what was happening with the flex boxes. I will do it one more time to see if we need it or not.

This was a little annoying to troubleshoot, it's hard to see the interaction of the flexboxes / container widths and the invisible hitSlop. So a lot was just trial and error until I got it working.

MacDeluca: Looks like it behaves the same with or without. Adding back the flexShrink property to keep it consistent with before.

bryce-mcmath (approved):
🔙

MacDeluca:
Aproximate hit slop.

Screenshot_20251203-093755

Issue #2997, opened 2025-12-02
Update Google Services file patching to use full exact file


bryce-mcmath:
### Description
Currently our CICD patches BC Wallet to become BCSC v4, part of that patching is edits to our google-services.json and GoogleService-Info.plist files. But these edits don't match the actual files from the IAS Firebase console. The patches should be updated to match correctly so that we properly receive and are able to register for push notifications

### Related issues
#2986 
#2752 
#2720

Issue #2986, opened 2025-11-27
Login with Pairing code is not as per v3.15.1


RaulConstan:
### Description of problem
There is a difference between v4.0 and 3.x versions.  See the two different flow video clips.

### Steps to reproduce
See videos.

### Screenshots and/or log output
v4.0 actual:

https://github.com/user-attachments/assets/db2d7700-1358-4270-b364-7b7c3d9fd82c

v3.x actual:

https://github.com/user-attachments/assets/70f240d8-b1d3-40e0-acf0-1af2a6fb8330

### Environment

Build #:
3504

Android Device Model:

iOS Device Model:
14 Plus

jeznorth:
@jleach to follow up with external dependencies to confirm.

bryce-mcmath:
Marcos believes this is a side effect of push notifications not being registered properly, most likely due to incorrect firebase / google services config, which there is now a PR to update (linked by mention above)

RaulConstan:
As of build 3527, the fix has not been merged yet.

bryce-mcmath:
Even with the updated Firebase / Google config files, push notifications are still not being received for some reason. Moving back to backlog

gvodden:
looks good in 3789

PR #3002, opened 2025-12-02
fix(cicd): update patches with exact google service config

bryce-mcmath:
# Summary of Changes

This PR just updates the patch files for BCSC iOS and Android to use the exact Google Services files from the Firebase Console.

Note: I noticed some oddities in the patch, we are kind of switching from AriesBifold to BCWallet in some places but not all? There's a good chance that could cause unforeseen issues as well. Will deal with in a later PR.

# Testing Instructions

Nothing to test from this branch, only from Testflight / Beta

# Acceptance Criteria

Testflight and Beta builds maybe start getting push notifications

N/A

# Related Issues

#2997
#2986


al-rosenthal (dismissed):
🔥

al-rosenthal (approved):

PR #3005, opened 2025-12-03
fix: skip onboarding notifications screen if previously granted

MacDeluca:
# Summary of Changes

This PR bypasses the notifications screen in the onboarding workflow if the user has previously accepted them.

It will show the screen if the user has denied them previously.

Note: There is an error in react-native when accepting android permissions 3+ times. The request permissions function will block and never resolve. I added a temporary timeout for that request to prevent that behaviour. 

If the user is seeing the notifications screen for the 3rd time, the continue button will hang for 2 seconds and then continue. Previously this would block the user, and they would need to fresh install the application.

Issue: https://github.com/facebook/react-native/pull/53898
Fixed in: https://github.com/reactwg/react-native-releases/issues/1151 V0.81.x
Commit: https://github.com/facebook/react-native/commit/447a7a3527e3e38e4c3ceb330d12d77afe7bc0b4

# Testing Instructions

### First onboarding pass
- Fresh install the app
- Go through onboarding
- Deny notifications system prompt
- Continue to setup steps
- Press reset app setup steps button

### Second onboarding pass
- Go through onboarding
- Should see notifications screen
- Accept notifications prompt
- Continue to setup steps
- Press reset app setup steps button

### Third onboarding pass
- Go through onboarding
- Should NOT see notifications screen
- Continue to setup steps

# Acceptance Criteria

Follow testing instructions.

Note: 
First prompt: select deny
Second prompt: select accept

https://github.com/user-attachments/assets/1601b48e-13ab-4057-aa4b-6faf5e2123cc

# Related Issues


bryce-mcmath:
Should we add a loading spinner and state to the continue button for that 2 second timeout?

MacDeluca:
> Should we add a loading spinner and state to the continue button for that 2 second timeout?

Up to you. This will only affect us until we upgrade react-native, so this will be automatically fixed before release. Technically this patch won't even need to be removed.

Its also a super specific edgecase for this to occur.

ie: react-native !== V0.81.x && User must have seen that onboarding screen twice, denied the notifications prompt both times and is viewing that onboarding screen for the 3rd+ time.

Only then will it load for 2 seconds before continuing onto the next screen.

bryce-mcmath (approved):
Good either way wrt the loading spinner

PR #3031, opened 2025-12-10
feat: snowplow analytics integration

MacDeluca:
# Summary of Changes

This PR introduces Snowplow analytics. 

- Created a wrapper class around the snowplow tracker API ie: AnalyticsTracker
- Tied analytics into navigation to automatically track screen view events (current and previous)
- Created new native alert function `showAlert` to render a typed alert and track analytics events
- Created `AnalyticsSystemCheck` to initialize analytics on startup
- Modified existing analytics screen to initialize analytics during onboarding

#### `trackAlertActionEvent`: tracks the action of an alert, to spec using existing iglu schema

# Testing Instructions

Note: Snowplow micro dev server only setup for Android currently.

Note: Currently there is a schema issue with the alert events. I have messaged Marcos about it, he is waiting to hear from someone.

- On Android build application and install packages
- Run `yarn snowplow:start` to start the dev server
- Navigate through onboarding
- Accept analytics
- View snowplow micro server `localhost:9090/micro/ui`
- Verify screen events are registering

# Acceptance Criteria

- Can see screen view events in the Snowplow micro server

# Related Issues


al-rosenthal (changes requested):
Toggling analytics in the settings doesn't work as expected. If toggling from on -> off, analytics will continue to work until the app is restarted. The same is true for going from off -> on, the app needs to be restarted before changes are applied.

app/src/utils/analytics/analytics-tracker.ts
```diff
@@ -0,0 +1,184 @@
+import { AlertEvent, AlertInteractionEvent } from '@/events/alertEvents'
+import { newTracker, ReactNativeTracker } from '@snowplow/react-native-tracker'
+import { getBuildNumber, getBundleId, getIpAddress, getUniqueId, getVersion } from 'react-native-device-info'
+import { getPlatformContextProperties, getPlatformContextRetriever } from './platform-context-retriever'
+
+export const ANALYTICS_SINGLEAPP_NAMESPACE = 'singleapp_client'
+export const ANALYTICS_SINGLEAPP_ENDPOINT = __DEV__ ? 'localhost:9090' : '' // TODO (MD): Add production endpoint
+export const ANALYTICS_MOBILE_ERROR_EVENT_SCHEMA = 'iglu:ca.bc.gov.idim/mobile_error/jsonschema/1-0-0'
```
al-rosenthal: I think these should be in our const file

app/src/utils/analytics/platform-context-retriever.ts
```diff
@@ -0,0 +1,170 @@
...
+    getPhysicalMemory: safeAsync(getTotalMemory),
+    getAppAvailableMemory: safeAsync(_getAppAvailableMemory),
+    getBatteryLevel: safeAsync(getBatteryLevel),
+    getBatteryState: safeAsync(_getBatteryState),
+    getLowPowerMode: safeAsync(_getLowPowerMode),
+    isPortrait: safeAsync(_isPortrait),
+    getResolution: safeAsync(_getResolution),
+    getScale: safeAsync(_getScale),
+    getLanguage: safeAsync(_getLanguage),
+    // getAppleIdfa?: () => Promise<string | undefined>;
```
al-rosenthal, line 50: do we need to wait until we have access to a development server for snowplow to tackle these?

al-rosenthal (commented):
ran locally and analytics were being sent to micro server.

app/src/utils/analytics/platform-context-retriever.ts
MacDeluca: Honestly not entirely sure, I went to capture as much metadata as I could but couldnt find the correct functions to retrieve this data.

I just left it as comments so we can easily see some are missing for the future.

app/src/utils/analytics/analytics-tracker.ts
MacDeluca: I removed the exports to make it a little more clear that these values aren't shared. They are specific to the analytics implementation and I feel more useful being assigned in the file they are used.

If these values were used across modules, I would move them into the shared constants.ts file.

MacDeluca: ie: I could've written the schemas inline where they are being used, instead I assigned them at the top of the file to prevent hardcoding.

To me this feels different than a constant that is shared accross multiple files and modules

MacDeluca:
> Toggling analytics in the settings doesn't work as expected. If toggling from on -> off, analytics will continue to work until the app is restarted. The same is true for going from off -> on, the app needs to be restarted before changes are applied.

This has been resolved :)

al-rosenthal (approved):
ran this locally, works as expected

PR #3047, opened 2025-12-15
fix: prevent script injections in github actions

travkhl:
# Summary of Changes

This PR refactors several github actions files to remove expression expansions from shell commands. Instead, data is now assigned to environment variables which are handled differently by actions.

> ### How does this work?

>The key difference lies in when and how the untrusted data is processed:

> * Direct interpolation (vulnerable): When using ${{ github.event.pull_request.title }} directly in a shell command, GitHub Actions processes the workflow file and expands the expression before the shell script runs. The expansion happens at the workflow processing stage:
>   1.  GitHub Actions reads the workflow YAML file
>   2. It evaluates ${{ github.event.pull_request.title }} and replaces it with the actual PR title value
>   3. The resulting string (which may contain shell metacharacters) is inserted into the script
>   4. The shell then receives and executes the already-expanded script
> * Environment variable approach (safe): When the untrusted data is assigned to an environment variable using the env key, GitHub Actions still expands the expression, but the expansion happens in a different context:
>   1. GitHub Actions expands ${{ github.event.pull_request.title }} and assigns the value to an environment variable
>   2. The shell script references this variable using standard shell syntax ($PR_TITLE)
>   3. The shell treats the environment variable’s contents as literal data, not as executable code

>When the shell evaluates $PR_TITLE, it performs variable expansion (not expression expansion), which treats the value as a string literal. Even if the value contains shell metacharacters, they remain as literal characters and are not interpreted as commands. This separation ensures that malicious shell syntax in the untrusted data cannot be interpreted as commands, effectively neutralizing injection attempts.

# Testing Instructions

# Acceptance Criteria

# Related Issues

https://github.com/bcgov/bc-wallet-mobile/issues/3062


.github/workflows/actions/create-ios-sig-creds/action.yml
```diff
@@ -27,8 +27,9 @@ runs:
       shell: bash
       env:
         CERTIFICATE: ${{ inputs.certificate }}
+        certificate_password: ${{ inputs.certificate_password }}
```
MacDeluca: nit: I would suggest we use screaming snake case for our env variables. I know these values are slightly different, but for consistency I think they should be the same casing as the other values.

ie: CERTIFICATE_PASSWORD

travkhl: changed 🐍🗣️

lucabolzonello (approved):
LGTM

bryce-mcmath (approved):

PR #3081, opened 2025-12-19
fix: navigation error

jleach:
# Summary of Changes

Fixed a navigation error where `navigation` as used before it was fully initialized.

```
 ERROR  The 'navigation' object hasn't been initialized yet. This might happen if you don't have a navigator mounted, or if the navigator hasn't finished mounting. See https://reactnavigation.org/docs/navigating-without-navigation-prop#handling-initialization for more details.
```

### DIFF
--- a/app/App.tsx
+++ b/app/App.tsx
@@ -70,7 +70,8 @@ const App = () => {
   // Wrap navigationRef.navigate to avoid accessing it before initialization
   const navigate = (stack: never, params: never) => {
     if (navigationRef.isReady()) {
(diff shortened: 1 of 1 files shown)


bryce-mcmath:
This way it will fail silently and maybe intentional navigation will be missed (like to a startup modal on app expiry). I wonder if it would be better to catch the navigation failing where it is attempted and retry after a short delay? 

edit: also, did you find out what was actually using this navigation ref immediately on startup?

jleach:
I don't see the error on main any longer. Closing.

Issue #3091, opened 2025-12-23
iOS migration from v3 to v4


bryce-mcmath:
### Description of problem
User should be able to migrate from an existing v3 verified account or an existing v3 in progress verification to v4 and pick up where they left off with the same auth method.

Currently auth, tokens, account, account flags, are all migrated correctly but not the credential or authorization request. These should migrate successfully as well.

### Steps to reproduce
Create v3 account with Testflight v3, in progress or fully verified. Install v4, unlock your app and be shown the correct screen with correct data

### Environment
iOS
(Android migration has not yet been tested, we will leave for a separate ticket as it requires more finesse to setup the test)

PR #3092, opened 2025-12-24
fix: ios migration

bryce-mcmath:
# Summary of Changes

This PR fixes migration from v3 from both a fully verified account and an account with an in progress verification. The changes are are almost entirely in native Swift code, adjusting the object unwrapping and adding more logging.

The two TS changes I made were:
1) adjusting the nickname setting effect in RootStack as the state load was overwriting the set nickname
2) making failed registration deletions in the factoryReset non-blocking, so the rest of the reset process continues to complete

Note:
- `requiringSecureCoding: false` is required because v3 uses deprecated archive methods that have requiringSecureCoding set to false by default

# Testing Instructions

Scenario 1:
- Fresh install previous Testflight build, 3.15.1, stop once you get past the serial number + birthdate step
- Do an iOS build from this branch
- Check that you can use the same PIN / device auth to log in and that you have the same access to setup steps with an in progress setup steps

Scenario 2:
- Fresh install previous Testflight build, 3.15.1, get fully verified
- Do an iOS build from this branch
- Check that you can use the same PIN / device auth to log in and that you have the same access to your full account

# Acceptance Criteria

The above two scenarios work

Here is a v3 Testflight build, fully verified, being migrated to a build from this branch:

https://github.com/user-attachments/assets/c17ed85f-036a-4492-98d9-a1058c64cf0e

# Related Issues

#3091

### DIFF
--- a/packages/bcsc-core/ios/ClientRegistration.swift
+++ b/packages/bcsc-core/ios/ClientRegistration.swift
@@ -26,16 +26,7 @@ class ClientRegistration: NSObject, NSSecureCoding {
     _ = coder.decodeObject(forKey: .keys)
 
     let decodedCredential = coder.decodeObject(forKey: .credential)
-    print(
-      "ClientRegistration.init: Decoded credential object type: \(type(of: decodedCredential)), value: \(String(describing: decodedCredential))"
-    )
     self.credential = decodedCredential as? Credential
-    if self.credential == nil, decodedCredential != nil {
-      print(
-        "ClientRegistration.init: Failed to cast credential to Credential type, actual type: \(type(of: decodedCredential))"
-      )
-    }
-
     self.accessToken = coder.decodeObject(forKey: .accessToken) as? String
     self.accessTokenID = coder.decodeObject(forKey: .accessTokenID) as? String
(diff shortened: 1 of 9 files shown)


packages/bcsc-core/ios/ClientRegistration.swift
```diff
@@ -19,14 +19,22 @@ class ClientRegistration: NSObject, NSSecureCoding {
...
+    // V3 has a keys array we don't need in v4 - decode and ignore it
+    _ = coder.decodeObject(forKey: .keys)
 
-    // Decode credential with all required Foundation types for secure coding
-    self.credential = coder.decodeObject(
-      of: [Credential.self, NSString.self, NSDate.self, NSNumber.self, NSArray.self],
-      forKey: .credential
-    ) as? Credential
+    let decodedCredential = coder.decodeObject(forKey: .credential)
+    print(
```
KjartanE: Is print the best logger for swift?

bryce-mcmath: Ah these are leftover debugging logs, will remove

bryce-mcmath: Done

KjartanE (approved):
Looks good! 🐱

PR #3115, opened 2026-01-08
feat(fcm): trigger device invalidation check on status notification

jleach:
# Summary of Changes

This PR fixes an issue where users had to restart the app to see the device invalidation screen after their device was canceled. The app now detects this immediately when receiving a push notification.

When a status notification comes in, the app refreshes the OAuth tokens and checks the new token data. If it finds the device was canceled, it shows the invalidation screen right away instead of waiting for the next app launch.
Testing Checklist

# Testing Instructions

1. Crate a new test account via the management UI
2. Add the newly created test account to your device
3. Verify the account (in person, send video, etc)
4. With the app setup and ready:
5. go back to the management UI and cancel the card (**WARNING: THIS CANNOT BE UNDONE**). 

You should receive a push notification and the device should automatically detect the card was canceled and show the invalidation screen.

-  Push notification received while device is active
-  DeviceInvalidated screen appears immediately (no restart needed)
-  Account data updates correctly after token refresh

# Acceptance Criteria

https://github.com/user-attachments/assets/110f7a3d-b1dc-4acb-92ae-85da74f0871f

# Related Issues

### DIFF
--- a/app/src/bcsc-theme/features/modal/DeviceInvalidated.tsx
+++ b/app/src/bcsc-theme/features/modal/DeviceInvalidated.tsx
@@ -19,7 +19,7 @@ type DeviceInvalidatedProps = StackScreenProps<BCSCMainStackParams, BCSCModals.D
 export const DeviceInvalidated = ({ route }: DeviceInvalidatedProps): JSX.Element => {
   const { t } = useTranslation()
   const [store] = useStore<BCState>()
-  const invalidationReason = route.params.invalidationReason
+  const invalidationReason = route.params?.invalidationReason
   const factoryReset = useFactoryReset()
   const [logger] = useServices([TOKENS.UTIL_LOGGER])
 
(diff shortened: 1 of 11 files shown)


app/src/bcsc-theme/features/modal/DeviceInvalidated.tsx
```diff
@@ -29,6 +29,10 @@ export const DeviceInvalidated = ({ route }: DeviceInvalidatedProps): JSX.Elemen
   const handleFactoryReset = useCallback(async () => {
     const factoryResetParams: Partial<Record<BCSCReason, Partial<BCSCState>>> = {
       // Can add more cases here for different BCSCReason types in the future
+      [BCSCReason.Cancel]: {
```
bryce-mcmath, line 32: If we have CancelledByUser and CancelledByAgent, what is this new Cancel from?

app/src/bcsc-theme/utils/id-token.ts
```diff
@@ -25,10 +25,10 @@ export enum BCSCReason {
   ApprovedByAgent = 'Approved by Agent',
   Renew = 'Renewed by Card Renew',
   Replace = 'Replaced by Card Replace',
-  Cancel = 'Canceled by Card Cancel',
+  Cancel = 'Canceled by Card Cancel', // i.e agent via web portal.
   ExpiredBySystem = 'Expired by System',
-  CanceledByAgent = 'Canceled by Agent',
-  CanceledByUser = 'Canceled by User',
+  CanceledByAgent = 'Canceled by Agent', // i.e automatic, too many cards.
```
bryce-mcmath, line 30: Seeing this, I think there may be a name mixup. Cancel should really be called CanceledByAgent, and CancelledByAgent should really be called something like CancelledByNewAccounts or something

bryce-mcmath: Look like the misnomer already existed but since we're working on the same area might be good to fix now

jleach: @bryce-mcmath I also thought that to and checked with Kjartan. They're correct 🤷🏻

bryce-mcmath (approved):

Issue #3121, opened 2026-01-08
IOS: Masked camera error


MacDeluca:
### Description of problem
Masked camera emitting error during non bcsc flow
### Screenshots and/or log output

<img width="1179" height="2556" alt="Image" src="https://github.com/user-attachments/assets/4fa44803-8a74-47b6-82b0-d891914bda26" />

Camera error: [unknown/unknown]: Error Domain=AVFoundationErrorDomain Code=-11800 "The operation could not be completed" UserInfo={NSLocalizedFailureReason=An unknown error occurred (-16401), NSLocalizedDescription=The operation could not be completed, NSUnderlyingError=0x14227dad0 {Error Domain=NSOSStatusErrorDomain Code=-16401 "(null)"}} (caused by {"code":-11800,"details":{"NSUnderlyingError":null,"NSLocalizedDescription":"The operation could not be completed","NSLocalizedFailureReason":"An unknown error occurred (-16401)"},"message":"Error Domain=AVFoundationErrorDomain Code=-11800 \"The operation could not be completed\" UserInfo={NSLocalizedFailureReason=An unknown error occurred (-16401), NSLocalizedDescription=The operation could not be completed, NSUnderlyingError=0x14227dad0 {Error Domain=NSOSStatusErrorDomain Code=-16401 \"(null)\"}}","domain":"AVFoundationErrorDomain"}
### Environment

Build #:

Android Device Model:

iOS Device Model:

Issue #3117, opened 2026-01-08
iPhone - Verify by send video approval not being received by app.


gvodden:
### Description of problem
iPhone 15  iOS 18.7.3
App build 3926

The approval from the verify by send video is not being received by the phone app, and the Check Status button does not respond.

The app is stuck in an unapproved state, and selecting another method does not work:"
In Person results in " Confirmation number already approved - view mobile card activity for more information" on IDCheck

### Expected behavior
App should receive notification that the verification was approved
App should be approved and show the "You're All Set" screen

### Steps to reproduce
Fresh install of build 3926
register BCSC card and submit Send Video request.
Close app
Wait 5 minutes (not critical, but I did it)

Approve request on ID Check

<should get push notification, but they are not working right now>
Open App 
<should get You're all Set message, but that is not happening>
tap Check Status
<should show either still in progress, or results, but doesn't do anything>

### Environment  SIT

Build #: 3926

Android Device Model:

iOS Device Model: 15+

### Workaround
User must reset data and start again using In Person or Live video as IAS Thinks that the in person code has already been approved, so the app data is out of sync with IAS

bryce-mcmath:
Fixed in 3125, to be clear just fixing the check status button not working, the push notification etc will be in a different ticket

gvodden:
Check Status works in build 4134.
Really looking forward to the rest.

PR #3125, opened 2026-01-09
fix: two persistence / storage issues

bryce-mcmath:
# Summary of Changes

Two issues:
1) We were not persisting or properly fetching the verification request ID, which was causing verifications to get lost between lockouts / app restarts
2) We were overwriting the NSArray class, which was blowing up the camera (and who knows what else) (my b lol) 

Another good example of where native code tests would help, taking stuff from v3 doesn't always work with some adaptation

# Testing Instructions

Go through non-BCSC flow
Go through send video, when you have a pending request, restart the app, check you can still "Check status" of it

# Acceptance Criteria

You can use the camera in non-BCSC flow
You can have a send video request pending, restart the app, and still check it successfully

N/A

# Related Issues

#3121 
#3117


app/src/bcsc-theme/features/verify/non-photo/EvidenceCaptureScreen.tsx
```diff
@@ -66,6 +67,16 @@ const EvidenceCaptureScreen = ({ navigation, route }: EvidenceCaptureScreenProps
       setCaptureState(CaptureState.CAPTURING)
       setCurrentPhotoPath(undefined)
       setCapturedPhotos([])
+
+      // Delay camera mount until after navigation transition completes
```
jleach, line 71: `useFocusEffect`?

jleach (approved):

PR #3158, opened 2026-01-15
feat: alerts 1-10 + client error policies

MacDeluca:
# Summary of Changes

This PR introduces a new error class "AppError", api endpoint error handling and custom alert actions. 

- Centralizes the client error handling to a single entry point for endpoint specific errors.
- Simplifies the existing ErrorAlert context API and integrates with the AppError class
- Ties Analytics into the new AppError class
- Updates the Axios client error handling transform AxiosErrors into AppErrors
- Adds new concept clientErrorPolicies. Custom error handlers that have endpoint specific context. ie: execute an action when a specific AppError shows on endpoint '/token'

# Testing Instructions

I extended the ErrorAlertTest.tsx component to inject error codes into the axios interceptor. This is to simulate a failed request.

In settings/developer/Error & Alert Testing theres a new section: Server error codes

1. no_internet: noop
2. unsecured_network: all endpoints
3. server_timeout: all endpoints
4. server_error: all endpoints
5. ios_app_update_required: evidence endpoints
6. android_app_update_required: evidence endpoints
7. no_tokens_returned: token endpoint (navigates to the remove account screen)

# Acceptance Criteria

# Related Issues


app/src/screens/Splash.tsx
bryce-mcmath, line 1: This screen is only for BC Wallet, no? I wonder if it makes sense to leave this out

MacDeluca: The ErrorAlertContext API changed slightly, this is just a refactor.

MacDeluca: The change being: error analytics now happens in the constructor of the AppError. So the showModal property became obsolete (previously used to track the error in analytics without showing the error dialog).

I also decoupled UI concerns (showModal) from the ErrorDefinition type so it only represents domain-level error semantics.

app/src/bcsc-theme/api/clientErrorPolicies.ts
```diff
@@ -0,0 +1,107 @@
...
+  },
+  handle: (error, context) => {
+    context.emitErrorAlert(error, {
+      actions: [
+        {
+          // QUESTION (MD): The docs suggest using "Update" for android, do we want to differentiate here?
+          text: context.translate('Alerts.Actions.GoToAppStore'),
+          onPress: async () => {
+            const appStoreUrl = getBCSCAppStoreUrl()
+            await context.linking.openURL(appStoreUrl)
```
KjartanE: The openURL could fail silently, wrap in a try catch

app/src/errors/errorHandler.ts
```diff
@@ -79,3 +79,25 @@ export function logError(
...
+ * @returns The corresponding ErrorDefinition or null if not found.
+ */
+export const getErrorDefinitionFromAppEventCode = (appEvent?: string): ErrorDefinition | null => {
+  if (!appEvent) {
+    return null
+  }
+
+  const definitions = Object.values(ErrorRegistry)
+
+  for (const definition of definitions) {
```
KjartanE: iterating over every definition on every API error. Seems like possible poor performance

app/src/screens/ErrorAlertTest.tsx
```diff
@@ -115,21 +137,31 @@ const ErrorAlertTest: React.FC<ErrorAlertTestProps> = ({ onBack }) => {
...
     })
   }
 
+  const injectErrorCodeIntoAxiosResponse = async (client: BCSCApiClient, errorCode: string, endpoint?: string) => {
+    const id = client.client.interceptors.request.use((config) => {
+      client.client.interceptors.request.eject(id)
+      throw new AxiosError('Injected error message', errorCode, config)
+    })
+
+    await client.get(endpoint ?? '/any-endpoint')
```
KjartanE: I know its part of the developer screen. But this request is also not wrapped with try catch

app/src/bcsc-theme/contexts/BCSCApiClientContext.tsx
```diff
@@ -65,7 +123,11 @@ export const BCSCApiClientProvider: React.FC<{ children: React.ReactNode }> = ({
           !BCSC_API_CLIENT_SINGLETON ||
           BCSC_API_CLIENT_SINGLETON.baseURL !== store.developer.environment.iasApiBaseUrl
         ) {
-          newClient = new BCSCApiClient(store.developer.environment.iasApiBaseUrl, logger as RemoteLogger)
+          newClient = new BCSCApiClient(
+            store.developer.environment.iasApiBaseUrl,
+            logger as RemoteLogger,
+            handleApiClientError
```
KjartanE: handleApiClientError is only passed to the client once at creation. If navigation or emitErrorAlert change later, the client will use stale references. Add a useEffect to sync BCSC_API_CLIENT_SINGLETON.onError when handleApiClientError changes.

app/src/errors/errorHandler.ts
MacDeluca: Yea I can fix that, the list is still quite small for it to matter yet, but for future proofing I can export a map.

app/src/bcsc-theme/contexts/BCSCApiClientContext.tsx
MacDeluca: Nice catch. I was so distracted from the memory leak issue that I totally missed this.

I have updated the BCSCApiClientContext to update the client reference when the handleApiClientError changes now.

bryce-mcmath (dismissed):
handled 🏖️🧉

bryce-mcmath (approved):

PR #3192, opened 2026-01-26
feat: add AGENTS.md file

jleach:
# Summary of Changes

Moves our github copilot instructions file to the more command and standardized  [AGENTS.md](https://agents.md/) to help get better aligned and quality results from AI agents used by various teams and Copilot reviews. This should work better with editors that are not VSCode.

# Testing Instructions

N/A

# Acceptance Criteria

N/A

N/A

# Related Issues

N/A


AGENTS.md
```diff
@@ -0,0 +1,231 @@
...
+    UserProfileViewModel.ts
+  /views            # UI components
+    /components     # Reusable UI components
+    /screens        # Full screen views
+  /adapters         # Data transformation layers
+    UserAdapter.ts
+  /services         # External service integrations
+    ApiService.ts
+  /utils            # Utility functions
+```
```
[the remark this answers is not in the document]
jleach: Fixed.

AGENTS.md
```diff
@@ -0,0 +1,275 @@
+# Project Context & AI Persona
+
+You are an expert mobile developer specializing in clean architecture, performance optimization, and robust UI implementation. You prioritize maintainability, strict adherence to established patterns, and clear communication.
```
bryce-mcmath: Kind of a nit but could we mention React Native here? Since that encompasses a lot of specific domain knowledge

AGENTS.md
```diff
@@ -0,0 +1,275 @@
...
+  }
+
+  private static getInitials(name: string): string {
+    /* ... */
+  }
+  private static generateAvatarUrl(id: string): string {
+    /* ... */
+  }
+}
+```
```
MacDeluca: question: Are we actually using this Adapter pattern?

AGENTS.md
```diff
@@ -0,0 +1,275 @@
+# Project Context & AI Persona
+
+You are an expert mobile developer specializing in clean architecture, performance optimization, and robust UI implementation. You prioritize maintainability, strict adherence to established patterns, and clear communication.
+
+## Architecture Patterns
+
+### MVVM (Model-View-ViewModel)
+
+This project follows the MVVM architecture pattern. Always structure code according to these principles:
+
```
MacDeluca, line 10: suggestion: It looks like the below examples and sections are quite different than the actual implementation we are using. I think we currently use a hybrid approach to this pattern, with two distinct pieces. 

1. `model` - react hook which wraps reactive logic into a single location 
example: useSetupStepsModel
2. `view` - react component which injests the useModel hook and renders the content 
example: SetupStepsScreen.tsx

I think we should update the examples and sections to be aligned with the current pattern we use in the codebase.

AGENTS.md
jleach: Some might argue the ViewModel is an adapter pattern on steroids because it bridges the M and V layers. But I'm happy to just call it a ViewModel if that's more clear.

AGENTS.md
jleach: It's a good point. I think you're describing the React-idiomatic implementation where `hooks` are the ViewModel. I'll update to reflect it to be less of an academic doc and more about how we apply the concept.

MacDeluca (approved):

Issue #3183, opened 2026-01-22
Birthdate attempt lockout


bryce-mcmath:
### Description of feature / user story
In BCSC v3, if you attempt the incorrect birthdate for your entered serial number four consecutive times, you are locked out from further attempts for 10 minutes.

### Acceptance Criteria
- Like-for-like with v3

### Wireframes or relevant image assets / links
- https://citz-cdt.atlassian.net/browse/IASP-4814

jeznorth:
Review UI updates/changes with UX team before completion.

bryce-mcmath:
In 4269

jeznorth:
I've observed the following when testing this...

- After 4 attempts of entering a birthdate, I'm met with the 'Sorry, you've tried too many times' interface.
- Tapping 'Close' redirects me back to the setup steps, with step 2 highlighted.
- I can continue to proceed into step 2 again, select a card, scan the barcode and proceed to the birthdate screen.
- After 1 attempt, I'm met with the 'Sorry you've tried too many times' interface.

Verified this is consistent with V3.

Issue #3180, opened 2026-01-22
birthdate picker often reads incorrect year.


gvodden:
https://github.com/user-attachments/assets/e58ec7fe-bdc0-44e4-9ea2-eee64e53c225

### Description of problem
when selecting your birthdate using the rollers, the last roller adjusted is often not updated correctly, and record an incorrect value.

This gives the user a "Card Not Found" error.

### Expected behavior
The date pickers should not require a delay to pick the correctly entered date.

### Steps to reproduce
enter card info.
select date and submit (somewhat quickly)
See video for playback...

### Environment  SIT

Build # 4134

Android Device Model: Galaxy S8

iOS Device Model: 15Plus

bryce-mcmath:
in 4269

gvodden:
looks good in 4289!

PR #3210, opened 2026-02-01
feat: birthdate lockout and bug

bryce-mcmath:
# Summary of Changes

This PR adds the birthdate lockout feature and fixes a bug with the birthdate picker. To do so, it includes a small refactor of the error handling around device authorization

# Testing Instructions

Go through the BCSC flow, quickly changing the birthdate and hitting the continue button and ensuring the date sent to the API matches the date you chose on the picker.

Attempt the wrong birthdate four consecutive times, and ensure on the fourth attempt you are navigated to the new lockout screen. Ensure when you back navigate or click close, you are take directly to setup steps.  

# Acceptance Criteria

Feature works as expected, rapid picker choice doesn't cause incorrect date.

Current v3 lockout:
https://github.com/user-attachments/assets/820028fa-f4d9-4578-9ce0-2473f55e546c

v4 lockout from this PR:
https://github.com/user-attachments/assets/07200e4c-7dd9-4afd-98dd-fe91f56d0cf7

# Related Issues

#3183
#3180


app/src/bcsc-theme/api/client.ts
```diff
@@ -19,12 +19,13 @@ import { withAccount } from './hooks/withAccountGuard'
 declare module 'axios' {
   export interface AxiosRequestConfig {
     skipBearerAuth?: boolean
+    skipPolicyHandlers?: boolean
```
MacDeluca: question: is this used?

bryce-mcmath: yes, I added usage of this in this PR for the authorizeDevice call in the TransferQRScannerScreen. That screen doesn't want any navigation when the device is already registered etc.

bryce-mcmath: after the refactor this was no longer needed

app/src/errors/appError.ts
```diff
@@ -96,7 +98,18 @@ export class AppError extends Error {
...
+
+/**
+ * Check if an error is an AppError that has already been handled by an error policy.
+ *
+ * @param error - The error to check
+ * @returns True if the error is an AppError and has been handled
+ */
+export function isHandledAppError(error: unknown): boolean {
+  return error instanceof AppError && error.handled
+}
```
MacDeluca, line 115: suggestion: Modifing this into a type-guard would enhance typing

[Typescript: User defined type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

al-rosenthal (commented):
I ran this locally on iOS and found that, when entering card details, if the date picker is still spinning when you hit the continue button, it doesn't update the date as expected. 

On android the picker and screen work as expected.

bryce-mcmath:
> I ran this locally on iOS and found that, when entering card details, if the date picker is still spinning when you hit the continue button, it doesn't update the date as expected.
> 
> On android the picker and screen work as expected.

That is a limitation of the iOS date picker unfortunately (and is the same in v3.) Android has the `onStateChange` API to detect whether it is still spinning or not, but iOS has no equivalent. The important thing is that when it does stop spinning and you hit submit, is the date correct?

app/src/errors/appError.ts
bryce-mcmath: done

al-rosenthal (approved):

PR #3211, opened 2026-02-02
feat: non prod overlay

bryce-mcmath:
# Summary of Changes

Simple non-prod overlay to indicate which environment is being used. 

# Testing Instructions

On iOS and Android, run the BCSC app. Switch environments and confirm prod and preprod don't show the overlay and the other environments show the correct name. 

# Acceptance Criteria

Passes the above test

https://github.com/user-attachments/assets/87a1a255-d3ae-4b9d-a86f-4ec8b470e1a0

# Related Issues

N/A


app/src/components/NonProdOverlay.tsx
```diff
@@ -0,0 +1,62 @@
...
+const DIAGONAL = 1.414 // sqrt(2)
+const HALF_DIAGONAL = DIAGONAL / 2
+
+const NonProdOverlay: React.FC = () => {
+  const [store] = useStore<BCState>()
+  const { Spacing } = useTheme()
+  const envName = store.developer.environment.name.split(' ')[0].toUpperCase()
+
+  // Only show if not production (or preprod)
+  const isProd = envName.includes('PROD')
```
MacDeluca: question: Do we want to consider 'preprod' as PROD? Do we ship the application as 'preprod' for users?

MacDeluca:
Love this so much 🥇 

opinion: I think reducing the triangle overlay size and increasing the transparency slightly would help to make this less distracting for the devs.

MacDeluca:
question: I wonder if we need a way to enable and disable the overlay for demos?

bryce-mcmath:
preprod


@MacDeluca how's this?

app/src/bcsc-theme/components/NonProdOverlay.tsx
```diff
@@ -0,0 +1,62 @@
...
+      fontSize: 14,
+      width: TRIANGLE_SIZE,
+      top: TRIANGLE_SIZE / DIAGONAL - Spacing.md,
+      right: TRIANGLE_SIZE / DIAGONAL,
+      textAlign: 'center',
+      transform: [{ rotate: '-90deg' }],
+    },
+  })
+
+  if (isProd) {
```
bryce-mcmath, line 47: @MacDeluca easiest way to disable for demos. If we want, we can add an additional developer toggle for hiding this overlay / indicator in every environment, but that might be more than necessary for this PR

```suggestion
  if (true) {
```

MacDeluca: Yea I like that, we can add a dev toggle later if it becomes annoying.

MacDeluca:
Tested on iOS: ✅ 
Tested on Android: ✅

MacDeluca (approved):

Issue #3191, opened 2026-01-26
Check for, and update if, any remaining dependencies blocking the 16kb requirement


bryce-mcmath:
### Description
Use the Android Studio bundle analyzer to verify the 16kb requirement, and if any dependencies are still blocking it, update them.

Related, as part of this ticket, update the minimum iOS deployment target to 15 as that is recommended for our new version of React Native.

### Developer notes
The remaining dependencies blocking 16kb:
- A one liner resolution to an older version of a native Android barcode scanning dependency (`implementation 'com.google.mlkit:barcode-scanning:17.2.0'`)
- react-native-keychain needs a major version update
- react-native-argon2 needs a major version update

Issue #2664, opened 2025-09-12
Update app to ensure it uses 16KB memory page size


jleach:
The app must use 16KB memory page size by November 1, 2025 or we will not be able to upload new version to the Google Play store. Existing versions should continue to work past this date.

### What does this entail (!)
The following libraries must be upgraded to versions that support 16kb page sizes:
- **React Native** (min 79, worth just going to 81 if it's not too much trouble)
- **indy-vdr-react-native, anoncreds-react-native, aries-askar-react-native must all be upgraded** (work is in progress by the OSS community)
- react-native-vision-camera (can be done pretty painlessly with RN upgrade)
- **react-native-argon2** must be replaced, as it has been abandoned. Tim from Ontario is a good person to talk to about alternatives

![](https://chat.developer.gov.bc.ca/file-upload/jvBGuBqtetHDWyoYH/play_console.png)

jleach:
Out of an abundance of caution I applied for the extension:

<img width="411" height="145" alt="Image" src="https://github.com/user-attachments/assets/09b9ea01-854c-458e-895c-8587c4cb2d13" />

TimoGlastra:
> react-native-argon2 must be replaced, as it has been abandoned. Tim from Ontario is a good person to talk to about alternatives

Regarding argon2, berend and i were talking about adding this directly to askar. That way if you have askar, you have argon2, and we don't need another potentially outdated lib.

@berendsliedrecht since argon2 seems to become an issue with 16kb page size, maybe we can add it to askar in the coming month?

berendsliedrecht:
Yeah, I think adding it to askar is good. Do we want to expose the API via credo crypto API or just directly from @owf/askar react native package?

TimoGlastra:
I think let's start with OWF Askar, since we now also integrate it directly, and it would unblock us. Since it's very specific to mobile I'm not sure if there's a need to expose it directly in Credo (but we can always add it at a later stage).

berendsliedrecht:
argon2 is now available directly from askar using v0.4.3

https://www.npmjs.com/package/@openwallet-foundation/askar-react-native 

API is quite simple 

```typescript
Argon.derivePassword(Argon2Parameters.Interactive, passwordBytes, saltBytes)
```

PR #3229, opened 2026-02-03
feat(deps): 16kb requirement

bryce-mcmath:
# Summary of Changes

This PR fixes the remaining issues blocking the Android 16kb requirement and improves performance. This was done with the following changes:
- updated Bifold packages
- removed react-native-argon2 (Bifold handles using Askar's Argon2 method used in it's place)
- updated to 0.5.x Askar which has the correct Argon2 ability
- removed an old resolution in build.gradle that pinned the project to an old version of the barcode scanning library (scanning is moderately more effective on Android now)
- updated react-native-keychain from 8.1.3 to 10.0.0
- removed unneeded intl polyfills (faster TTI)
- disabled bundle compression on Android (faster TTI on Android)

# Testing Instructions

On both iOS and Android, do a BC Wallet build and go through onboarding, get a credential. **On iOS you may notice that labels for credentials / proofs are incorrect or missing, this is an existing issue tracked elsewhere and is not caused by this PR**

# Acceptance Criteria

App still works on both OS

N/A

# Related Issues

#2664 
#3191


KjartanE:
- [x] **Android:** Build and run; check QR/barcode scanning (transfer or verify flows).
- [x] **iOS:** Build and run; check keychain still works and microphone permission is correct.
- [x] **Both:** Spot-check i18n (dates/numbers in EN/FR/pt-BR) and run the test suite.

al-rosenthal (approved):
tested locally on both android and iOS. I was able to get credentials, respond to proofs and on iOS I was able to get my person credential. 

✅

PR #3237, opened 2026-02-03
chore: wait for navigation readiness and refactor system checks

MacDeluca:
# Summary of Changes

Previously we were using a temp hack to prevent calling code when navigation that wasn't ready (navigation container not mounted).

With a recent addition of the NavigationContainerContext we can now access the state of the container via a custom hook (useNavigationContainer.isNavigationReady). This allows us to display a loading screen while navigation is mounting and a prevent issues with undefined navigation downstream.

This also includes a refactor of the systemchecks to use navigation readiness vs awaiting custom timeout function. Also seperates the system checks react hook into two parts:

1. useSystemChecks - calls the system checks based on the scope

2. useCreateSystemChecks - organizes dependencies and returns system checks for scope, also includes "readiness" state

# Testing Instructions

App should run and behave the same, all previous system checks should still run.

# Acceptance Criteria

System checks should run for startup and main stack.
Logs should be improved to include list of system checks that ran and their status (runCheck result)

# Related Issues


jleach (approved):
Runs fine, no navigation issue reported in the logs. New account setup. System check ran:

```
[useSystemChecks]: Ran 3 system checks on scope: STARTUP {
  "AnalyticsSystemCheck": false,
  "ServerStatusSystemCheck": true,
  "UpdateAppSystemCheck": true
} 
```

Issue #3218, opened 2026-02-02
Login to service seems to be circular and not actually log in


RaulConstan:
### Description of problem
The login for the BC Parks seems to be circular without actually getting to the login confirmation screen.

### Expected behavior
The v4 app should be triggering a confirmation login request from the service - it does not.

### Steps to reproduce
See video.

### Screenshots and/or log output

https://github.com/user-attachments/assets/513e81dc-7262-4569-b841-288d706b7db3

### Environment

Build #:
4224

Android Device Model:

iOS Device Model:
iPhone 14 Plus

jleach:
Fixed in 4.0.0-4344.

RaulConstan:
@jleach It's using the incorrect flow to log in...  4353 is using the log in from a computer flow, instead of log in on the mobile device flow.  (look to Government Gateway OIDC flow - that's how it should work).

jleach:
Video from 4.0.0-3704. I think it's working as expected based on the video above. I'm not sure I understand the comment above. 

https://github.com/user-attachments/assets/48e5b472-d74f-4609-a2de-5e45ad98cb74

RaulConstan:
Found additional issues with the flow:
1. v4 has Close button (v3 does not) see 15/34 seconds timestamp in Jason's vid.
2. v4 does not have the app switching button (back to Safari) in the top left corner (v3 does)
3. wording and formatting are different between v3 and v4
Note: could change the Close button to function as the app switcher to the browser...

@jleach returning to needs refinement.

RaulConstan:
Providing additional v3 footage of the flow:
see timestamp 00:46.

https://github.com/user-attachments/assets/593a7ade-d6be-424e-8595-24a719a354d5

jleach:
Test in 4.0.0-4756 or higher.

RaulConstan:
Currently blocked by:

<img width="462" height="1128" alt="Image" src="https://github.com/user-attachments/assets/c1a65fb4-4476-49ec-9d99-02ee7c616956" />

gvodden:
I was able to test this on build 4892 in SIT, but the v4 flow still leaves the user in the BCSC App, and not on the web browser.
The v3.15 app has a tappable link to close the app and return to the browser :

<img width="417" height="409" alt="Image" src="https://github.com/user-attachments/assets/478ad0ce-e76b-401a-9a73-36ad7d0a3a94" />

[...]

PR #3240, opened 2026-02-04
fix: circular navigation on external service link

jleach:
# Summary of Changes

Fix navigation when doing a deep link from the a services screen. See video in #3218 for more information.

# Testing Instructions

1. Open the app with a setup account
2. Go to the services screen / tab
3. Select BC Parks (needs to Link to an external service, this one I know does)
4. Open the external link
5. Logon with BC Services App. This will deep link to the app
6. The app should open and show you the correct "Do you want to logon ..." screen. It should **not** be showing you the screen with the link anymore (what you saw when you left the app).

# Acceptance Criteria

Instructions pass

n/a

# Related Issues

#3218

### DIFF
--- a/app/src/bcsc-theme/features/services/hooks/useServiceLoginState.test.ts
+++ b/app/src/bcsc-theme/features/services/hooks/useServiceLoginState.test.ts
@@ -1,4 +1,4 @@
-import { renderHook, waitFor } from '@testing-library/react-native'
+import { act, renderHook, waitFor } from '@testing-library/react-native'
 
 import { useServiceLoginState } from './useServiceLoginState'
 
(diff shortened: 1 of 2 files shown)


MacDeluca:
Android: ✅ 
iOS: ❌ 

Was able to go through the flow on Android no problem. 

On iOS I ran into an error "Safari cannot open the page because the address is invalid" (with nothing in the logs)

MacDeluca:
After applying the patch ios-local-dev.patch the deep linking worked correctly on iOS.

Probably will need to investigate why Android doesn't need a patch but iOS does for local development.

MacDeluca (approved):

MacDeluca (approved):

PR #3244, opened 2026-02-04
feat: terms of use content gradient

bryce-mcmath:
# Summary of Changes

I missed this screen when adding content gradients in my last PR

# Testing Instructions

Go through onboarding, check out the terms of use

# Acceptance Criteria

There is a lovely content gradient that makes it easier for users to recognize there is more scrollable content

https://github.com/user-attachments/assets/45679138-bfff-427e-b54e-2f3af964cdf4

# Related Issues

N/A


MacDeluca:
suggestion: Should we also always show the scroll bar?

bryce-mcmath:
> suggestion: Should we also always show the scroll bar?

~~I think that could be a good idea for this screen in particular, but we'll have to update the props of ScreenWrapper to allow for that kind of customization~~ waaait, this is a webview. There doesn't appear to be a way to make it permanently visible :(

jleach:
Confirmed on Android and iOS.

jleach (approved):

PR #3269, opened 2026-02-11
chore: use alerts hook

MacDeluca:
# Summary of Changes

WIP PR to wrap alerts into a reusable hook for easier configuration and injestion.

Prerequisite to #3061. Broad goal to un-couple alerts and errors (error registry / definitions)

ref: https://github.com/bcgov/bc-wallet-mobile/pull/3272#discussion_r2800899287

# Testing Instructions

Replace this text with detailed instructions on how to test the changes included in this PR.

# Acceptance Criteria

Replace this text with the acceptance criteria that must be met for this PR to be approved.

# Screenshots, videos, or gifs

Replace this text with embedded media for UI changes if they are included in this PR. If there are none, simply enter N/A

# Related Issues

Replace this text with tagged issue #'s that are relevant to this PR. If there are none, simply enter N/A

PR #3272, opened 2026-02-11
fix(service): surface keypair generation errors to user

jleach:
# Summary of Changes

Standardizes keypair generation error handling across iOS and Android native modules so failures are surfaced to the user via the existing error modal system.

# Testing Instructions

1. Add the following at like 1151 in `BcscCoreModule.kt`:
```
            // DEBUG: Force keypair generation error to test UI error handling
            throw KeypairGenerationException("DEBUG: Simulated keypair generation failure for UI testing")
```
2. Setup a new account.
3. You will see the error as expected.

# Acceptance Criteria

n/a

# Related Issues

### DIFF
--- a/app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
+++ b/app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
@@ -3,15 +3,18 @@ import { useCallback, useMemo } from 'react'
 import { Platform } from 'react-native'
 import {
   AccountSecurityMethod,
+  BcscNativeErrorCodes,
   getAccount,
   getAccountSecurityMethod,
   getDeviceId,
   getDynamicClientRegistrationBody,
+  isBcscNativeError,
   setAccount,
 } from 'react-native-bcsc-core'
 
 import useSecureActions from '@/bcsc-theme/hooks/useSecureActions'
 import { getNotificationTokens } from '@/bcsc-theme/utils/push-notification-tokens'
+import { useErrorAlert } from '@/contexts/ErrorAlertContext'
 import { BCState } from '@/store'
 import { TOKENS, useServices, useStore } from '@bifold/core'
 import BCSCApiClient from '../client'
@@ -60,6 +63,7 @@ const useRegistrationApi = (apiClient: BCSCApiClient | null, isClientReady: bool
   const [store] = useStore<BCState>()
   const [logger] = useServices([TOKENS.UTIL_LOGGER])
   const { updateTokens } = useSecureActions()
+  const { emitError } = useErrorAlert()
   /**
    * Retrieves platform-specific attestation for device verification.
    *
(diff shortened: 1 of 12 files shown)


packages/bcsc-core/src/index.ts
```diff
@@ -15,6 +15,35 @@ export type {
...
+  KEYPAIR_RETRIEVAL_FAILED: 'E_KEYPAIR_RETRIEVAL_FAILED',
+} as const;
+
+/**
+ * Shape of errors thrown by React Native native module promise rejections.
+ * The `code` property corresponds to the first argument passed to `reject()` / `promise.reject()`.
+ */
+export interface BcscNativeError extends Error {
+  code: string;
+  message: string;
```
MacDeluca: nit: message is already included in Error

app/src/errors/errorRegistry.ts
MacDeluca, line 691: @al-rosenthal 

thought (no action required): Talked to Alfred about this previously, but I feel the ErrorRegistry coupling UI alert properties with error properites feels strange. 

I think a future change would be to remove the UI properties and make error definitions specific to just errors.

This would prevent generic titles and descriptions being directly coupled to errors. ie: KEYPAIR_GENERATION_ERROR.title = "Problem with app"

WIP PR
https://github.com/bcgov/bc-wallet-mobile/pull/3269/changes#diff-a0f536b829b0f4c277760857e12a96edd88c3dd6d4868369393994845dfb257b

```ts
type ErrorDefinition = {
    statusCode: number
    appEvent: AppEventCode
    category: ErrorCategory
}

const alerts = useAlerts() // hook to organize alert callbacks

try {
    await someBcscCorePackageFn()
} catch (error) {
     // generate error with error registry + track in analytics automatically (or explicitly)
    new AppError('Developer message - problem while generating keypair', ErrorRegistry.KEYPAIR_GENERATION_ERROR, { cause: error, track: true })
     // use shared alert + track in analytics
    alerts.problemWithAppAlert()
}





```

app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
```diff
@@ -140,7 +144,15 @@ const useRegistrationApi = (apiClient: BCSCApiClient | null, isClientReady: bool
...
-      const body = await getDynamicClientRegistrationBody(fcmDeviceToken, deviceToken, attestation)
+      let body: string | null
+      try {
+        body = await getDynamicClientRegistrationBody(fcmDeviceToken, deviceToken, attestation)
+      } catch (error) {
+        if (isBcscNativeError(error) && error.code === BcscNativeErrorCodes.KEYPAIR_GENERATION_FAILED) {
+          emitError('KEYPAIR_GENERATION_ERROR', { error })
+        }
+        throw error
+      }
```
MacDeluca, line 155: I think emitError/emitAlert might be better triggered externally when the caller catches this error. This would help keep the hook callback focused on a single responsibility.

The caller already has to catch the error and action accordingly. Mostly just thinking about if there are times when we don't want to emit an error modal ie: injecting `register` into something else etc.

What do you think?

jleach: Done as part of PR #3279

MacDeluca (approved):
Tested locally and works as expected :)

PR #3318, opened 2026-02-25
fix: suppress expected 404 errors during attestation polling

jleach:
## Summary

- Added `suppressStatusCodeLogs: [404]` to the attestation status GET call in `useDeviceAttestationApi` so that 404 responses are not logged as errors during polling.
- Added `attestationPollingErrorPolicy` to `ClientErrorHandlingPolicies` so that 404s on the attestation endpoint are recognized as expected (attestation not yet consumed by `verifyAttestation`) and handled with an info-level log instead of being treated as unhandled errors.

These changes prevent noise in logs and error reporting from 404s that are a normal part of the attestation polling flow. Rather than red error logs / exceptions from Metro, you see the more graceful logging below:

```console
[BCSCApiClient] Applying error handling policy for: {
  "endpoint": "https://idsit.gov.bc.ca/device/attestations/79caeb5e-45db-49cf-bcab-ef1df7125db6",
  "appEvent": "unknown_server_error"
}
[AttestationPollingErrorPolicy] 404 expected during polling — attestation not yet consumed
```

## Test plan

- [ ] On a device with an account setup, transfer to another device from Account tab
- [ ] No error on HTTP 404 — just the logging noted above

### DIFF
--- a/app/src/bcsc-theme/api/clientErrorPolicies.test.ts
+++ b/app/src/bcsc-theme/api/clientErrorPolicies.test.ts
@@ -6,6 +6,7 @@ import { BCSCScreens } from '../types/navigators'
 import {
   alreadyRegisteredErrorPolicy,
   alreadyVerifiedErrorPolicy,
+  attestationPollingErrorPolicy,
   AxiosAppError,
   birthdateLockoutErrorPolicy,
   cardExpiredErrorPolicy,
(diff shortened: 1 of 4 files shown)


bryce-mcmath (approved):
Error logs not showing up ✅

PR #3321, opened 2026-02-26
chore(ci): build variants

cvarjao:
# Summary of Changes

Replace this text with a high-level summary of the changes included in this PR.

# Testing Instructions

# Acceptance Criteria

# Related Issues


.github/workflows/main.yaml
```diff
@@ -130,35 +132,309 @@ jobs:
             echo "result=false" >> $GITHUB_OUTPUT
           fi
 
+  # ─── Shared Preparation ─────────────────────────────────────────────
+  # Builds the JS bundle once for all variants, saving ~2-3 min per
+  # variant build. Also warms the Metro bundler cache so that any
```
bryce-mcmath: Love this

.github/workflows/main.yaml
```diff
@@ -130,35 +132,309 @@ jobs:
...
       matrix:
         include:
-          - build_type: 'bc-wallet'
-            app_version: '1.0.27'
-            export_options: 'export-options-bcw.plist'
-            release_config: 'release-bcw.xcconfig'
-            artifact_name: 'ios-bc-wallet-artifact'
+          - variant: 'bcwallet-prod'
+            build_target: 'bcwallet'
+            export_method: 'app-store'
```
bryce-mcmath, line 376: What is the difference between the `app-store-connect` used elsewhere and the `app-store` used here?

.github/workflows/main.yaml
```diff
@@ -130,35 +132,309 @@ jobs:
...
+  # ─── Shared Preparation ─────────────────────────────────────────────
+  # Builds the JS bundle once for all variants, saving ~2-3 min per
+  # variant build. Also warms the Metro bundler cache so that any
+  # residual re-bundling in iOS xcodebuild is near-instant.
+  #
+  # Future optimizations:
+  #   - Use dynamic matrix (fromJSON) to build only bcwallet-prod +
+  #     bcsc-dev on pull requests, all variants on main/dispatch.
+  #   - Upload pre-built iOS JS bundle and use SKIP_BUNDLING=1 in
+  #     xcodebuild to skip Metro entirely for iOS too.
```
bryce-mcmath: Is this a stale comment? I see SKIP_BUNDLING=1 further below in this workflow and the caching seems to be loading the JS bundle

.vscode/settings.json
```diff
@@ -73,5 +73,14 @@
...
     "typescript": "typescriptreact"
+  },
+  "chat.tools.terminal.autoApprove": {
+    "uniq": true,
+    "diff": true,
+    "paste": true,
+    "git rev-parse": true,
+    "git add": true,
+    "git commit": true,
+    "git push": true
```
bryce-mcmath: I don't think we want to include this in main

bryce-mcmath (changes requested):
Some really good stuff in here, love the caching and time saving. Couple things to double check, one or two things that should probably be adjusted or taken out

.github/workflows/main.yaml
jleach: @cvarjao This should be `app-store-connect`. I checked in XCode 16.4 and in the export options plist Apple uses `app-store-connect`.

cvarjao: I am not entirely sure, but here is what I found:

>The difference between app-store and app-store-connect in export-options.plist:

> * app-store — Exports a locally-signed IPA. The IPA is NOT uploaded during export. Upload to App Store Connect happens separately via the ship-to-itunes action (which uses codemagic-cli-tools / app-store-connect publish).
> * app-store-connect — Xcode handles both export AND upload to App Store Connect in one step during xcodebuild -exportArchive. Requires the authentication keys to be passed.

variants using `app-store-connect` cause `xcodebuild -exportArchive` to also attempt an upload, meaning the build could be uploaded twice (once by `xcodebuild`, once by `ship-to-itunes`).

AI is suggesting that I should use `app-store` since we ship the bundle separately. I have not verified that `app-store-connect` actually does the shipping as well. But I will switch to `app-store` as this is the current method being used. we can review/update later

.vscode/settings.json
cvarjao: removed

.github/workflows/main.yaml
cvarjao: updated comments/documentation

bryce-mcmath (dismissed):
🚢

bryce-mcmath (approved):

Issue #3282, opened 2026-02-17
Android - Under 12 year-olds are blocked from BCSC Mobile Cards


RaulConstan:
### Description of problem
Under 12 years of age BC residents are blocked on Primary Identification intake.

### Expected behavior
Under 12 years of age are not blocked from having a BCSC mobile app.

### Steps to reproduce
1. Have a 011 file loaded in IAS (under twelve years of age: mine is RC0000083, Jan 1, 2026)
2. Complete step 1
3. Start step 2, and select primary as Passport
4. Fill in the Passport intake form
5. Enter DoB, as per BCSC data in 011: in my case Jan 1, 2026 using the date picker.
6. Change focus out of the DoB field.

Actual:
A red warning is shown on DOB suggesting under 12 are not allowed to do this process.

### Screenshots and/or log output

https://github.com/user-attachments/assets/ec98612d-7ad9-4d09-84c3-f2814b3129c2

### Environment
SIT

Build #:
4495

Android Device Model:
Pixel 6 (Android 16)

iOS Device Model:

### Workaround
None

jleach:
# Spec: Age-Based Verification Method Restrictions

## Overview

When a user adds their BC Services Card to the app, the server determines which identity verification methods are available based on the user's date of birth (and potentially other factors). The client does not perform age calculations — it displays whatever the server returns.

There are two distinct age gates with different thresholds and behaviors:

| Age Gate | Threshold | Behavior |
|----------|-----------|----------|
| Mobile card setup | Under 12 | **Blocked entirely** — cannot set up a mobile card |
| Person credential | Under 19 | **Blocked entirely** — cannot create a Person credential |

For users 12+ who pass initial card setup, the server controls which verification methods are offered (e.g., under a certain age the server may only return `counter`).

---

## Flow: Device Authorization

### Request

`POST` to the provider's device authorization endpoint with a URL-encoded body:

```
client_id={clientId}&scope={scope}&response_type=device_code&card_serial_number={csn}&birth_date={ISO8601Date}
```

- `birth_date` format: ISO 8601 date only (e.g. `2010-05-15`), using the device's local timezone
- `Content-Type: application/x-www-form-urlencoded`

### Success Response

The response includes a `verification_options` field — a **space-separated string** of allowed verification method identifiers:

```
"counter"
"video_call counter"
"back_check video_call counter"
"back_check counter"
```

Other relevant response fields:

| Field | Type | Description |
|-------|------|-------------|
[...]

PR #3322, opened 2026-02-26
fix: show correct evidence form based on card process

jleach:
**TL;DR** This may be an incremental fix to #3282. First step is to get it showing the correct forms under the correct circumstances. This is PR addresses this by showing the abbreviated data collection (document ID only) for NPC. Next PR will address age restrictions.

## Key Changes
- Show abbreviated evidence form (document number only) for NPC 2nd ID flow (`BCSCNonPhoto`)
- Show full form (document number + name + DOB) for "Other ID" flow (`NonBCSC`)
- Replace `additionalEvidenceData.length` check with `cardProcess`-based check to correctly distinguish flows
- Cancel button now removes evidence from store and navigates back to evidence type list
- Added `removeEvidenceByType` to `useSecureActions` for targeted evidence cleanup

## Test plan
- [ ] NPC flow: select "card without a photo" → enter serial + DOB → select 2nd ID (e.g. passport) → verify only document number field is shown
- [ ] "Other ID" flow: select "Other ID" → select passport → verify full form (document number, name, DOB) is shown
- [ ] Cancel on evidence form returns to evidence type list in both flows
- [ ] After cancel, previously selected evidence type is available for re-selection


app/src/bcsc-theme/features/verify/non-photo/EvidenceIDCollectionScreen.tsx
```diff
@@ -123,38 +127,23 @@ const EvidenceIDCollectionScreen = ({ navigation, route }: EvidenceIDCollectionS
...
-  const validateEvidence = (values: EvidenceCollectionFormState, additionalEvidenceRequired: boolean) => {
+  const validateEvidence = (values: EvidenceCollectionFormState) => {
     const errors: EvidenceCollectionFormErrors = {}
 
     if (!isDocumentNumberValid(values.documentNumber)) {
       errors.documentNumber = t('BCSC.EvidenceIDCollection.DocumentNumberError')
     }
 
-    if (!additionalEvidenceRequired) {
+    if (!personalInfoRequired) {
```
jleach, line 159: The old name was a little misleading. Additional evidence is always required, but some times it's the abbreviated form (document ID only) and some times it needs to be the full personal information (first / last / middle / dob / etc)

bryce-mcmath (changes requested):
I think this will stop showing the form fields for personal info whenever the flow isn't BCSC. But there is the case in non-BCSC where two additional IDs are required and we only want the personal info form to show up for the first of the two additional IDs. Which is the reason we had the length check

jleach:
> I think this will stop showing the form fields for personal info whenever the flow isn't BCSC. But there is the case in non-BCSC where two additional IDs are required and we only want the personal info form to show up for the first of the two additional IDs. Which is the reason we had the length check

I tried it as is Passport first, then Citizenship Card. It asked all details for #1, then auto-filled for #2. Will fix. V3 just asks for document number.

bryce-mcmath (approved):
Tested, works as expected for both non-Photo and non-BCSC

Issue #3208, opened 2026-01-30
Video call does not terminate properly


jleach:
### Description of problem

I was testing a Send Video verification scenario but mistakenly pressed Video Call instead. I didn't notice until after I took the photo and the video call started. As soon as I saw the call connecting screen, I cancelled and returned to select Send Video. While I was watching the countdown timer for the video recording, I heard the recorded message from the Video Call saying no agent was available.

### Expected behavior

The video call should have been terminated 100%.
Auto inactivity lockout during livecall should be prevented

### Steps to reproduce

As per the description.

### Screenshots and/or log output

Notice in this screen shot that the green dot is visible. This is active when the mic is active.

![Image](https://github.com/user-attachments/assets/410bc3f7-2bda-4fdc-ac0d-3bc6a60dabc3)

### Environment

Build #:

Android Device Model:

iOS Device Model:

bryce-mcmath:
4659

RaulConstan:
Here is a video with audio to showcase how Cancelling a call does not actually cancel it.

https://github.com/user-attachments/assets/aa5a8148-0937-455f-aaf5-a77715d2ec3a

bryce-mcmath:
4785

PR #3327, opened 2026-02-27
fix: live call cleanup and activity monitor pause

bryce-mcmath:
# Summary of Changes

This PR does two things for live call:
- pauses activity monitoring during live call so users don't get kicked out of a call due to not touching the screen for a given time
- fixes cleanup logic so there isn't a race condition, allowing all calls to be cancelled properly

There's also a three character change to a button position to get closer to wireframe

# Testing Instructions

See the reproduction steps in the ticket linked below

# Acceptance Criteria

All calls that should be ended, are fully ended

N/A

# Related Issues

#3208


jleach (approved):
Had to test this one on Android because my iOS setup is not working. Call ended clean when I swiped back.

jleach (approved):

PR #3340, opened 2026-03-02
fix: alert styling

KjartanE:
# Summary of Changes

- **Error modal:** New `BCSCErrorModal` with white background, dark text, and improved layout. "Report this problem" is first (primary), "Okay" is second (outlined).
- **Analytics:** Report button uses Snowplow only (no remote logger). Error tracked once on display; Report sends an alert action event.
- **Other:** Theme-based colors, fixed button/Show Details placement, plain payload from context. Unit tests added/updated.

# Testing Instructions

1. Trigger an error modal (e.g. Error Alert Test screen).
2. Confirm: white card, dark text, Report then Okay, Show Details when a technical message exists.
3. Tap Report → shows "Reported" and is disabled; tap Okay → modal closes.
4. Run: `jest src/components/ErrorModal.test.tsx src/errors/errorHandler.test.ts src/contexts/ErrorAlertContext.test.tsx`

# Acceptance Criteria

- [ ] White background, dark text (no red/white-on-dark).
- [ ] Report first, Okay second; Report uses Snowplow only.
- [ ] Show Details visible when technical message exists.
- [ ] Unit tests pass.

image
image
image

# Related Issues

N/A


bryce-mcmath (changes requested):
App.tsx affects both BC Wallet *and* BCSC - the ErrorModal chosen should depend on the mode the app is in to keep BC Wallet the same as it was

KjartanE:
> App.tsx affects both BC Wallet _and_ BCSC - the ErrorModal chosen should depend on the mode the app is in to keep BC Wallet the same as it was

added a conditional `{appMode === Mode.BCSC ? <BCSCErrorModal enableReport /> : <ErrorModal enableReport />}`

Leaving BCWallet utilizing the Bifold ErrorModal, if we are wanting to update styling for Bifold we will either create another modal or update the bifold component

app/src/components/ErrorModal.tsx
```diff
@@ -0,0 +1,354 @@
...
+    const removeHandler = DeviceEventEmitter.addListener(EventTypes.ERROR_REMOVED, () => {
+      setError(null)
+      setVisible(false)
+    })
+
+    return () => {
+      addHandler.remove()
+      removeHandler.remove()
+    }
+  }, [])
```
MacDeluca: question: do we need the device event emitters?

Could we not just conditionally render this error modal in the ErrorAlertContext/Provider component?

app/src/components/ErrorModal.tsx
```diff
@@ -0,0 +1,354 @@
...
+
+  const handleReport = useCallback(() => {
+    if (!error) {
+      return
+    }
+
+    if (error.appEvent) {
+      const definition = getErrorDefinitionFromAppEventCode(error.appEvent)
+      if (definition) {
+        trackErrorInAnalytics(definition, AlertInteractionEvent.ALERT_ACTION, t('Error.ReportThisProblem'))
```
bryce-mcmath: I don't know if it makes sense for the report a problem button to popup an alert that also has a report a problem button

app/src/components/ErrorModal.tsx
KjartanE: We definitely could rewire this component fully and move modal state into the `ErrorAlertProvider`. But my initial intention was to follow the `BifoldError` as close as possible which allows for a simple swap of `ErrorModal`s into the App level. If there is value to a broader refactor and merging of `ErrorModal` and the `ErrorAlertProvider` happy to add to this PR.

app/src/components/ErrorModal.tsx
KjartanE: This is just tracking that the user has submitted the error for a report. Not triggering another alert. :D

bryce-mcmath: icic, ty

app/src/components/ErrorModal.tsx
MacDeluca: Yea I dunno, imo it feels like this is overly complicated using the `DeviceEventEmitter` when it probably could be simplified.

I'm easy though, whatever you think feels best.

KjartanE: There was a case if the ErrorBoundary could emit to the same listener. But didn't work as expected so moved to common components with ErrorInfoCard and rewired the `ErrrorAlertProvider` to utilize the `ErrorModal`

app/src/contexts/ErrorAlertContext.test.tsx
```diff
@@ -1,54 +1,60 @@
...
+}))
+
 jest.mock('react-native', () => ({
-  DeviceEventEmitter: {
-    emit: jest.fn(),
-  },
   Alert: {
     alert: jest.fn(),
   },
+  Modal: 'Modal',
```
bryce-mcmath, line 19: What does this effectively do?

bryce-mcmath (approved):
Tested, works as expected, nicely done!

Issue #3203, opened 2026-01-29
BCSC v4 and BC Wallet should answer to different deep link protocols


jleach:
### Description of problem

BCSC v4 and BC Wallet v1 should not answer to the same deep link protocols. For example, BC Wallet should respond to `ca.bc.gov.BCWallet` and `didcomm` where as BCSC v4 should answer to `ca.bc.gov.id.servicecard`. If they both answer to the same protocol then Android (maybe iOS) will prompt the user to open both apps which can be confusing. 

### Expected behavior

Android and iOS should only prompt to open a deep link with one app.

### Environment

Build #:

Android Device Model:

iOS Device Model:

bryce-mcmath:
I think this ticket should be done after https://github.com/bcgov/bc-wallet-mobile/pull/3321 gets merged or else it's going to create some wasted effort

bryce-mcmath:
4873

PR #3341, opened 2026-03-02
fix: deep link protocols

bryce-mcmath:
# Summary of Changes

This PR makes it so only BC Wallet has didcomm and bcwallet inbound deeplinks, and only BCSC variants have the same inbound deeplinks as v3 (ias-ios and ias-android) including some v2 schemes we were missing. Not sure what those are used for but best to keep parity I figure.

The only difference between v3 and v4 now is that our non-prod environments accept all non-prod inbound schemes, not just their own, to account for the environment switching we have enabled. So qa, dev, sit, test, can all be tested with BCSC Dev still. Prod cannot, and prod does not accept non-prod inbound schemes.

# Testing Instructions
Do a BC Wallet build, try a BC Wallet deeplink from the showcase (the functionality doesn't work but just verify BC Wallet is opened)

Do a BCSC bcsc-dev variant build, try a BCSC demo deeplink

# Acceptance Criteria

Deeplinks work as expected and BC Wallet doesn't open BCSC stuff, BCSC doesn't open BC Wallet stuff

N/A

# Related Issues

#3203

### DIFF
--- a/variants/bcsc-dev/variant.env
+++ b/variants/bcsc-dev/variant.env
@@ -13,8 +13,8 @@ IOS_PRODUCT_NAME='$(TARGET_NAME)'
 IOS_PBXPROJ_TARGET_NAME='BCWallet'
 
 # URL schemes (comma-separated) — replaces base bcwallet/didcomm schemes
-IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.iddev.servicescard.v2'
-ANDROID_URL_SCHEMES='ca.bc.gov.iddev.servicescard.v2,ca.bc.gov.id.servicescard.dev'
+IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.idqa.servicescard,ca.bc.gov.idtest.servicescard'
+ANDROID_URL_SCHEMES='ca.bc.gov.id.servicescard.dev,ca.bc.gov.id.servicescard.qa,ca.bc.gov.id.servicescard.test'
 
 # Base variant to inherit shared assets from
 BASE_VARIANT='_bcsc-base'
--- a/variants/bcsc-dev/variant.env
+++ b/variants/bcsc-dev/variant.env
@@ -12,8 +12,9 @@ IOS_BUNDLE_ID='ca.bc.gov.iddev.servicescard'
 IOS_PRODUCT_NAME='$(TARGET_NAME)'
 IOS_PBXPROJ_TARGET_NAME='BCWallet'
 
-# URL schemes to add to Info.plist (comma-separated)
-IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.idqa.servicescard,ca.bc.gov.idtest.servicescard,ca.bc.gov.idsit.servicescard'
+# URL schemes (comma-separated) — replaces base bcwallet/didcomm schemes
+IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.iddev.servicescard.v2'
+ANDROID_URL_SCHEMES='ca.bc.gov.iddev.servicescard.v2,ca.bc.gov.id.servicescard.dev'
 
 # Base variant to inherit shared assets from
 BASE_VARIANT='_bcsc-base'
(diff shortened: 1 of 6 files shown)


variants/bcsc-dev/variant.env
```diff
@@ -12,8 +12,9 @@ IOS_BUNDLE_ID='ca.bc.gov.iddev.servicescard'
 IOS_PRODUCT_NAME='$(TARGET_NAME)'
 IOS_PBXPROJ_TARGET_NAME='BCWallet'
 
-# URL schemes to add to Info.plist (comma-separated)
-IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.idqa.servicescard,ca.bc.gov.idtest.servicescard,ca.bc.gov.idsit.servicescard'
+# URL schemes (comma-separated) — replaces base bcwallet/didcomm schemes
+IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.iddev.servicescard.v2,ca.bc.gov.idqa.servicescard,ca.bc.gov.idqa.servicescard.v2,ca.bc.gov.idtest.servicescard,ca.bc.gov.idtest.servicescard.v2'
```
jleach, line 16: Should this be limited to just "ca.bc.gov.idev.*` after this PR so that the DEV BCSC v4 only responds to DEV deep links? I'm thinking that each app should only respond to it's own deep links and not others.

jleach (changes requested):

variants/bcsc-dev/variant.env
jleach: Oh. NM. This build can change environments so it's probably needed.

variants/bcsc-qa/variant.env
```diff
@@ -12,8 +12,9 @@ IOS_BUNDLE_ID='ca.bc.gov.idqa.servicescard'
 IOS_PRODUCT_NAME='$(TARGET_NAME)'
 IOS_PBXPROJ_TARGET_NAME='BCWallet'
 
-# URL schemes to add to Info.plist (comma-separated)
-IOS_URL_SCHEMES='ca.bc.gov.id.servicescard,ca.bc.gov.iddev.servicescard,ca.bc.gov.idqa.servicescard,ca.bc.gov.idtest.servicescard,ca.bc.gov.idsit.servicescard'
+# URL schemes (comma-separated) — replaces base bcwallet/didcomm schemes
+IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.iddev.servicescard.v2,ca.bc.gov.idqa.servicescard,ca.bc.gov.idqa.servicescard.v2,ca.bc.gov.idtest.servicescard,ca.bc.gov.idtest.servicescard.v2'
```
jleach, line 16: In BCSC v3 QA cannot change back-end environments. This might be fine for now since they're not going to be around for much longer.

variants/bcsc-test/variant.env
```diff
@@ -12,8 +12,9 @@ IOS_BUNDLE_ID='ca.bc.gov.idtest.servicescard'
 IOS_PRODUCT_NAME='$(TARGET_NAME)'
 IOS_PBXPROJ_TARGET_NAME='BCWallet'
 
-# URL schemes to add to Info.plist (comma-separated)
-IOS_URL_SCHEMES='ca.bc.gov.id.servicescard,ca.bc.gov.iddev.servicescard,ca.bc.gov.idqa.servicescard,ca.bc.gov.idtest.servicescard,ca.bc.gov.idsit.servicescard'
+# URL schemes (comma-separated) — replaces base bcwallet/didcomm schemes
+IOS_URL_SCHEMES='ca.bc.gov.iddev.servicescard,ca.bc.gov.iddev.servicescard.v2,ca.bc.gov.idqa.servicescard,ca.bc.gov.idqa.servicescard.v2,ca.bc.gov.idtest.servicescard,ca.bc.gov.idtest.servicescard.v2'
```
jleach, line 16: In BCSC v3 TEST cannot change back-end environments. This might be fine for now since they're not going to be around for much longer.

jleach (changes requested):
Two comments about how I think BCSC v3 works vs 4. In QA and TEST you cannot switch environments in BCSC v3 so I would think it can only respond to its own deep link protocols. Thoughts?

bryce-mcmath:
> Two comments about how I think BCSC v3 works vs 4. In QA and TEST you cannot switch environments in BCSC v3 so I would think it can only respond to its own deep link protocols. Thoughts?

This is true, but in v4 we can change environments so if we want that to work still, we have to let all non-prod bundles handle eachother's deeplinks. Prod I've kept separate - prod can only open its own and only prod can open its own. Seeing  options when you press a BCSC deeplink as a tester is not a problem, it's only a problem if you're a user.

jleach (approved):

PR #3348, opened 2026-03-03
feat(service): surface ERR_111 missing JWK error to user

jleach:
## Summary

- `useJwksApi.getFirstJwk` now throws `AppError` (ERR_111_UNABLE_TO_VERIFY_MISSING_JWK) when the JWKS response contains no keys, instead of returning `null`
- `useQuickLoginUrl` catches ERR_111 and calls `alerts.missingJwkAlert()` to surface a "Problem with App" alert to the user
- `FcmViewModel` constructs an `AppError` for the missing-JWK case and marks it `handled = true` to trigger analytics tracking without showing an alert
- `useAlerts` gets `missingJwkAlert` wired up via `_createBasicAlert` (error code 111), following the pattern from ERRs 105, 109, 110
- Tests updated across `useJwksApi`, `useQuickLoginUrl`, `FcmViewModel`, and `useAlerts`

## Context

ERR_111 was already defined in the error enum and registry but was never thrown or surfaced. This follows the established pattern from #3272 / #3279: API hooks throw, service hooks catch and show alerts.

## Test plan

- [ ] No JWK returned from server during quick login flow shows "Problem with App (error 111)" alert
- [ ] Analytics event ERR_111 is tracked in both `useQuickLoginUrl` and `FcmViewModel` paths
- [ ] All existing tests pass (`yarn test`)

### How to trigger ERR_111 for manual testing

Add the following debug override to `app/src/bcsc-theme/api/hooks/useJwksApi.tsx` at the top of `getJwks()` to simulate an empty JWKS response:

```typescript
const getJwks = useCallback(async (): Promise<Keys> => {
  // DEBUG: simulate empty JWKS response for ERR_111 testing
  return []

  const {
    data: { keys },
  ...
```

Then trigger a quick login via ServiceLoginScreen. The "Problem with App (error 111)" native alert should appear. **Remove the debug line after testing.**

https://github.com/user-attachments/assets/a9e5d02c-15f5-46ec-a1ef-978e08780302


app/src/bcsc-theme/features/fcm/FcmViewModel.ts
```diff
@@ -192,7 +194,12 @@ export class FcmViewModel {
         this.lastJwkBaseUrl = apiClient.baseURL
         this.logger.info(`[FcmViewModel] Server JWK fetched successfully from ${apiClient.baseURL}`)
       } else {
-        this.logger.warn(`[FcmViewModel] No keys found in JWKS response`)
+        // Construct AppError (auto-tracked on creation) and mark as handled since this error is intentionally consumed
+        const appError = AppError.fromErrorDefinition(ErrorRegistry.MISSING_JWK_ERROR)
+        appError.handled = true
```
MacDeluca: question: do we need to mark this as `handled` if we arent throwing it again?

MacDeluca:
praise: code looks great, I would have implemented this the same way.

Just gonna run it locally now

app/src/bcsc-theme/features/fcm/FcmViewModel.ts
jleach: Fixed.

KjartanE (approved):
Tested locally, works as intended!

Issue #3357, opened 2026-03-04
Add JWT signature verification to decodePayload and consolidate with decodeLoginChallenge


jleach:
## Context

The legacy native apps (Android `JwtCoordinator`, iOS `BcscError`) performed a two-step pipeline: **decrypt JWE → verify JWS signature → extract claims**. ERR_114 was thrown when claims extraction failed after both decrypt and verify succeeded.

In the React Native implementation, the native `BcscCore` module has two separate methods:

### `decodePayload()`
- Decrypts JWE → extracts JWT payload via base64 decode
- **Does NOT verify the inner JWT signature**
- Used by `useUserApi.tsx` and `id-token.ts`

### `decodeLoginChallenge()`
- Parses signed JWT → extracts claims → optionally verifies signature
- Claims are extracted **before** and **independently** of verification
- Verification result is returned as a boolean flag, not used as a gate
- Used by `FcmViewModel.ts`

## Proposal

1. **Add signature verification to `decodePayload()`** so the JWE decrypt path also verifies the inner JWT's signature before returning claims — matching the legacy behavior.
2. **Consolidate** `decodePayload()` and `decodeLoginChallenge()` if possible, since both decode JWTs and extract claims. Consider a unified native method that handles: decrypt (optional, for JWE) → verify signature (optional, with provided key) → extract claims.
3. Ensure distinct native error codes so TypeScript can differentiate decrypt failures from verification failures from claims extraction failures.
4. At least make sure `decodePayload()` and `decodePayload()` return a similar structure if feasible.

## Files

- `packages/bcsc-core/ios/BcscCore.swift` — `decodePayload()` (line ~1033), `decodeLoginChallenge()` (line ~1084)
- `packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt` — `decodePayload()` (line ~1417), `decodeLoginChallenge()` (line ~3755)
- `packages/bcsc-core/src/NativeBcscCore.ts` — TypeScript interface

jleach:
Closing — the substantive part of this shipped in #3988 (merged 2026-06-24).
**Done:**
- `decodePayload` takes an optional JWK and returns `{ verified, claims }` on both iOS and Android, sharing a verify helper with `decodeLoginChallenge`.
[...]

PR #3358, opened 2026-03-04
feat(bcsc): implement ERR_114 failed to get claims set after decrypt and verify

jleach:
## Summary

- Added `failedToGetClaimsSetAlert` (ERR_114) alert definition in `useAlerts.tsx`
- Added null/type guard after `JSON.parse` in `useUserApi.tsx` and `id-token.ts` to throw ERR_114 when decoded payload is not a valid claims object
- Wrapped `decodeLoginChallenge` failure in `FcmViewModel.ts` with ERR_114 `AppError` for consistent structured error logging
- Added ERR_114 catch and alert dispatch in `useUserService.tsx` and `useTokenService.tsx`
- Added tests covering all new branches (5 new tests across 6 test files)
- No native changes — all TypeScript

Related: #3357 (JWT signature verification in native `decodePayload`)

## Test plan

- [x] Verify ERR_114 alert is shown when `getUserInfo` returns a non-object payload
- [x] Verify ERR_114 alert is shown when `getIdTokenMetadata` decodes a null payload
- [ ] Verify ERR_114 alert is shown in `useTokenService` on claims set error
- [x] Verify ERR_114 error is logged (not surfaced to user) in `FcmViewModel` when `decodeLoginChallenge` fails
- [ ] Run `yarn test` and confirm all new tests pass
- [ ] Confirm no native changes required

## Manual Testing

  Add a temporary throw right before line 116 in id-token.ts:

```
  // DEBUG: Force ERR_114 for visual verification
  throw AppError.fromErrorDefinition(ErrorRegistry.CLAIMS_SET_ERROR)
```

  Place it right after the JSON.parse succeeds (after line 114), so it always triggers when any ID token is decoded.


MacDeluca (approved):
Works great!

Issue #3343, opened 2026-03-03
Terms of Use momentarily blank on iphone 6s plus


gvodden:
https://github.com/user-attachments/assets/55146dcb-dc75-4370-8b50-5a5cdb77b723

### Description of problem
when setting up v4 (4857) to test QR transfer, I noticed that the ToU was not populating properly.

### Expected behavior
the ToU should be fully loaded before it is displayed to the user.

### Steps to reproduce
Fresh app install
(I entered dev mode and switched to QA env.
l progressed through the setup and captured the attached video.

### Environment
QA

Build #:  4857

Android Device Model: 

iOS Device Model: iPhone 6S Plus

bryce-mcmath:
4941

gvodden:
looks good in 4941!

PR #3359, opened 2026-03-04
fix: use terms api as v3 does

bryce-mcmath:
# Summary of Changes

v3 actually uses the API to get terms in onboarding, they only use the static URL in settings - we were using the static URL in both. It works by the API returning HTML which the app formats and prepends with the message you see in the screenshots below. I also added retry functionality incase the API call fails for any reason

I also got the edges of the terms to line up properly with the edges of the button 🤓 

# Testing Instructions

Look at terms on iOS / Android

# Acceptance Criteria

Terms works and matches v3 (aside from colouring)

terms
retry_terms

# Related Issues

#3343


al-rosenthal (approved):
ran this locally and it works as expected 📋

PR #3374, opened 2026-03-05
feat: add error alert 201-300 codes

KjartanE:
# Summary of Changes

Implements handling for IAS error codes 201–300. When the backend returns one of these codes (e.g. `add_card_server_configuration`, `err_206_missing_or_null_values_in_json_response`), the app now shows the right alert instead of failing silently.

- **useAlerts** — 15 new alerts (Problem with Service for 201–204, Add Card Not Available for 205, Problem with App for 206–213, 299, 300). For 202, 203, 204, OK is intended to return the user to Start Setup when the error occurs in the real setup flow (not when triggered from the dev screen).
- **clientErrorPolicies** — New `iasErrorPolicy` that matches these app events and triggers the corresponding alert. Wired into the existing policy list.
- **i18n** — New keys: ProblemWithService, DynamicRegistrationError, AddCardNotAvailable (en, fr, pt-br).
- **Tests** — Table-driven tests for all 15 codes (policy match + correct alert), plus `getAppErrorFromAxiosError` tests for the IAS → AppError mapping.
- **ErrorAlertTest** — IAS 201–300 added to the “Server error codes” list so QA can tap and verify each alert.

# Testing Instructions

1. **Automated:** `yarn test --testPathPattern="clientErrorPolicies|error-utils"` (all 128 tests should pass).
2. **Manual:** Open the app → Developer → Error & Alert Testing → scroll to “Server error codes”. Tap each of the new IAS entries (e.g. `add_card_server_configuration`, `err_206_missing_or_null`, …) and confirm the expected alert appears. (The dev screen does not route you back to Start Setup; it only fires the alert so you can verify copy.)

# Acceptance Criteria

- [ ] All new unit tests pass.
- [ ] Each IAS error code 201–300 shows the correct alert when triggered from ErrorAlertTest.
- [ ] 202, 203, 204 alerts show the correct copy; in the real setup flow, OK navigates to Start Setup (not testable from the dev screen).

N/A

# Related Issues

N/A

### DIFF
--- a/app/src/bcsc-theme/api/client.ts
+++ b/app/src/bcsc-theme/api/client.ts
@@ -138,11 +138,15 @@ class BCSCApiClient {
       }
 
       // 4. Invoke onError callback if provided which marks as handled
-      this.onError?.(appError as AxiosAppError, {
-        endpoint: String(error.config?.url),
-        statusCode: error.response?.status ?? 0,
-        apiEndpoints: this.endpoints,
-      })
+      try {
+        this.onError?.(appError as AxiosAppError, {
+          endpoint: String(error.config?.url),
+          statusCode: error.response?.status ?? 0,
+          apiEndpoints: this.endpoints,
+        })
+      } catch (handlerError) {
+        this.logger.error('[BCSCApiClient] Error handler threw', handlerError as Error)
+      }
 
       return Promise.reject(appError)
     })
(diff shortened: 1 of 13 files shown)


app/src/bcsc-theme/api/client.ts
```diff
@@ -138,11 +138,15 @@ class BCSCApiClient {
...
       // 4. Invoke onError callback if provided which marks as handled
-      this.onError?.(appError as AxiosAppError, {
-        endpoint: String(error.config?.url),
-        statusCode: error.response?.status ?? 0,
-        apiEndpoints: this.endpoints,
-      })
+      try {
+        this.onError?.(appError as AxiosAppError, {
+          endpoint: String(error.config?.url),
+          statusCode: error.response?.status ?? 0,
```
jleach, line 144: Probably `0` is a magic number here. Not going to nitpick at this point.

jleach (approved):

Issue #3095, opened 2025-12-29
Alerts with Error Codes (111 - 120)


jeznorth:
Version 4 needs to handle all alerts with associated error codes. Resolve error codes **111 through 120** as described in the following confluence documentation.

- https://citz-cdt.atlassian.net/wiki/spaces/BMS/pages/301574122/Mobile+App+Alerts#MobileAppAlerts-Alertswitherrorcodes

### Acceptance Criteria

- [x] err_111_unable_to_verify_missing_jwk
- [x] err_112_jws_verification_failed
- [ ] err_113_unable_to_verify_jose_exception (see comment below)
- [x] err_114_failed_to_get_claims_set_after_decrypt_and_verify
- [x] err_115_failed_to_serialize_json
- [ ] err_116_provider_unexpectedly_null (see comment below)
- [x] err_117_failed_to_parse_jws
- [ ] err_118_failed_to_retrieve_token_expiry (not applicable to v4 architecture. See comment below)
- [x] err_119_token_unexpectedly_null
- [x] err_120_client_registration_failure

jleach:
Skipping error 113

TL;DR: That is specific to a Android library. 

  Android — JwsVerifier.java:70
  } catch (JOSEException e) {
      throw new BcscException(AlertKey.ERR_113_UNABLE_TO_VERIFY_JOSE_EXCEPTION, e.getMessage());
  }
  Thrown when jws.verify(verifier) throws a JOSEException (as opposed to returning false, which is err_112). This is the catch block right after the signature verification attempt in the same method.

  - Enum: AlertKey.java:70

  iOS — Defined as enum case in BcscError.swift:117 but never actually thrown/used anywhere in the iOS Swift codebase. It appears to be a dead code path — iOS doesn't have an equivalent catch for JOSE exceptions.

  Alert config (both platforms) — bcsc_alerts.json:
  - iOS: "error 113"
  - Android: "error 113a"
  - Same message as err_112: "The app does not appear to be installed correctly. Please remove the app from your device and add it again."

  Root cause: A JOSEException during RSA signature verification — typically a cryptographic/key format issue rather than a simple signature mismatch (which is err_112).

jleach:
## ERR_115 Findings: `err_115_failed_to_serialize_json`

### Legacy codebase analysis (bcsc-native-v3)

[...]

PR #3378, opened 2026-03-06
feat: implement ERR_119 token unexpectedly null error handling

jleach:
## Summary

- Throw `AppError` with `ERR_119_TOKEN_UNEXPECTEDLY_NULL` when tokens are expected but null in `BCSCApiClient.ensureValidTokens()` and `useTokenApi.getCachedIdTokenMetadata()`
- Add `tokenUnexpectedlyNullAlert` to `useAlerts` hook (error code 119)
- Catch and surface ERR_119 in `useTokenService` service layer
- Add tests for alert, API hook, and service layer

Related #3095 (ERR_119 item)

## Test plan

- [ ] Open Developer Settings → Error Alert Test → scroll to useAlerts Hook section → tap `tokenUnexpectedlyNullAlert` → verify "Problem with App" alert with error 119
- [ ] Verify `yarn typecheck` passes
- [ ] Verify `yarn test` passes

### Via Developer Settings (visual only)
Developer Settings → Error Alert Test → useAlerts Hook section → tap **tokenUnexpectedlyNullAlert**

### Via real error path
Temporarily add this line in `app/src/bcsc-theme/api/client.ts` inside `ensureValidTokens()` (~line 205), before the `if (!this.tokens)` check:

```typescript
this.tokens = undefined // DEBUG: force ERR_119
```

Then trigger any authenticated API call (e.g. navigate to a screen that fetches user info). The error will flow through the real path: interceptor → `AppError` → service layer catch → alert.

Screenshot_20260306-101103


jleach:
Addressed Copilot review feedback in 9c98b56a — added early return in response error interceptor to pass through `AppError` instances without re-wrapping them as `UNKNOWN_SERVER_ERROR`. This ensures ERR_119 (and any future `AppError` thrown from the request interceptor) preserves its `appEvent` code through the interceptor chain.

bryce-mcmath (approved):

PR #3385, opened 2026-03-06
chore: updated v4 screen names to match v3

MacDeluca:
# Summary of Changes

This PR aligns V4 screen names with V3 screen names (to my best ability).

Added some fixmes inline where I was uncertain which screens they map to.

Note: Added this ticket https://github.com/orgs/bcgov/projects/108?pane=issue&itemId=164589903&issue=bcgov%7Cbc-wallet-mobile%7C3409 that requires a refactor of the existing screens. Need to determine if the effort is worth it.

# Testing Instructions

App should behave the same and tests pass.

# Acceptance Criteria

V4 screens map to correct V3 screen names.

# Related Issues

https://github.com/orgs/bcgov/projects/108?pane=issue&itemId=162611114&issue=bcgov%7Cbc-wallet-mobile%7C3370


app/src/bcsc-theme/types/navigators.ts
```diff
@@ -18,102 +18,120 @@ export enum BCSCModals {
...
+  OnboardingOptInAnalytics = 'Analytics Opt In', // NOTE: New V4 screen, not in V3
+  OnboardingWebView = `${BCSCStacks.Onboarding} Web view`,
+  OnboardingDeveloper = `${BCSCStacks.Onboarding} Developer`,
+  MainLoading = `${BCSCStacks.Main} Loading`,
+  MainSettings = `${BCSCStacks.Main} In App Settings`,
+  MainWebView = `${BCSCStacks.Main} Web view`,
+  MainContactUs = `${BCSCStacks.Main} Contact Us`,
+  MainPrivacyPolicy = `${BCSCStacks.Main} Privacy Information`,
+  MainDeveloper = `${BCSCStacks.Main} Developer`,
   MainAutoLock = 'BCSCMainAutoLock',
```
KjartanE, line 110: Nit: Few different pages maintained their "BCSC" Prefix names. Are these new pages to v4? Should they also get a more readable name?

app/src/bcsc-theme/types/navigators.ts
```diff
@@ -18,102 +18,120 @@ export enum BCSCModals {
...
-  MainAppSecurity = 'BCSCMainAppSecurity',
-  MainChangePIN = 'BCSCMainChangePIN',
-  VerifySettings = 'BCSCVerifySettings',
-  VerifyWebView = 'BCSCVerifyWebView',
+  MainAppSecurity = `${BCSCStacks.Main} App Security Setting`,
+  MainChangePIN = `${BCSCStacks.Main} Change PIN`,
+  VerifySettings = `${BCSCStacks.Verify} In App Settings`,
+  VerifyWebView = `${BCSCStacks.Verify} Web view`,
   VerifyAutoLock = 'BCSCVerifyAutoLock',
```
KjartanE, line 115: ^

app/src/bcsc-theme/types/navigators.ts
```diff
@@ -18,102 +18,120 @@ export enum BCSCModals {
...
-  VerifyContactUs = 'BCSCVerifyContactUs',
-  VerifyPrivacyPolicy = 'BCSCVerifyPrivacyPolicy',
-  VerifyDeveloper = 'BCSCVerifyDeveloper',
-  VerifyChangePIN = 'BCSCVerifyChangePIN',
+  VerifyAppSecurity = `${BCSCStacks.Verify} App Security Setting`,
+  VerifyContactUs = `${BCSCStacks.Verify} Contact Us`,
+  VerifyPrivacyPolicy = `${BCSCStacks.Verify} Privacy Information`,
+  VerifyDeveloper = `${BCSCStacks.Verify} Developer`,
+  VerifyChangePIN = `${BCSCStacks.Verify} Change PIN`,
   AccountExpired = 'BCSCAccountExpired',
```
KjartanE, line 121: ^

app/src/errors/components/ErrorModal.tsx
```diff
@@ -56,25 +54,15 @@ export const BCSCErrorModal: React.FC<BCSCErrorModalProps> = ({
...
-      if (definition) {
-        trackErrorInAnalytics(definition, AlertInteractionEvent.ALERT_ACTION, t('Error.ReportThisProblem'))
-      } else {
-        Analytics.trackErrorEvent({ code: String(error.code), message: error.message })
-      }
-    } else {
-      Analytics.trackErrorEvent({ code: String(error.code), message: error.message })
-    }
-  }, [error, t])
+    Analytics.trackAlertActionEvent(error.appEvent as AppEventCode, 'Report this problem')
```
KjartanE, line 64: Should this be hardcoded? or update to translation file

app/src/contexts/NavigationContainerContext.tsx
```diff
@@ -35,23 +36,26 @@ export const NavigationContainerProvider = ({ children }: PropsWithChildren): Re
         ref={navigationRef}
         theme={NavigationTheme}
         onReady={() => {
```
KjartanE, line 38: The old onReady used to capture the initial screen name into routeNameRef, so the first onStateChange would report it as the "previous" screen. The new code skips that, so the first transition will have previousScreenName = undefined. Every transition after that works fine. Just flagging in case the analytics dashboard expects a "from" screen on that first event — if not, no action needed.

MacDeluca: I think when I was testing this was firing as undefined initially and only onStateChange was the previous screen being correctly registered. I will re-test once again.

app/src/errors/components/ErrorModal.tsx
MacDeluca: Hardcoded / english is preferred. It's a value that's only used in our analytics dashboard.

app/src/bcsc-theme/types/navigators.ts
MacDeluca: I will take another look. I probably just missed them.

I can add a comment if they are new or need additonal context.

app/src/contexts/NavigationContainerContext.tsx
MacDeluca: Aight so I just double checked what the values were on initial startup, and we can leave it as is.

```ts
OnReadyScreenName: undefined
OnStateChangeScreenName: "Account Select"
```

KjartanE (approved):
Tested locally, works as expected. All Comments resolved. 🐱

Issue #3389, opened 2026-03-09
Handle suspended/deactivated API error during Person Credential creation


jleach:
## Description

When the Person Credential creation flow is implemented in RN, the app needs to handle HTTP 400 errors from the server when the account is suspended or deactivated.

### Server Response
- HTTP 400 with `error: "unauthorized_client"`, `error_description: "suspended"` or `"deactivated"`

### Expected Behavior (matching native apps)
Both suspended and deactivated should show an error screen:
- **Title:** "Cannot create Person credential"
- **Reason:** "There is a problem with your account."
- **Resolution:** "You need to remove your account from the app and set it up again. Then you can create your Person credential."
- **Actions:** Remove Account button, Close button

### Notes
- Suspend/unsuspend are NOT delivered via push notification or ID token — they only surface as API errors during Person Credential creation
- Emergency mode (allowing suspended credentials) is server-side only, no client logic needed
- Both `suspended` and `deactivated` show the same UI on native (iOS: `PersonCredentialUnavailableView`, Android: `AccountProblemFragment`)
- Analytics should track them separately: "Account problem app suspended" vs "Account problem app deactivated"

### Related
- #2748

anosha-ash:
Content update for error:

Deactivated account

- Title: Digital Services Card unavailable
- Body: Your BC Services Card account may be suspended or deactivated. It must be active before a Digital Services Card can be issued.

Suspended account (suspension normally happens due to bad actor, so better to provide no more info other than problem with account), as discussed with Marcos C.

- Title: Digital Services Card unavailable
- Body: There is a problem with your account.

al-rosenthal:
Decision point: swap screens and use a more generic "Something went wrong" modal with a report a problem button

PR #3395, opened 2026-03-09
feat: handle CanceledByCardTypeChange and CanceledDueToInactivity in DeviceInvalidated modal

### LINKED ISSUE #2748: Integrate and Map Credential Lifecycle Status
### Description of feature / user story

The application needs to properly ingest, interpret, and handle the "card lifecycle" items received within the verified app token response. This ensures the app accurately reflects the current status of the digital credential (e.g., active, cancelled, expired, suspended, revoked) in the user interface and business logic.

Eg. If the BC Services Card account is cancelled or expired the user should not be treated as verified - see Credential.swift in ias-ios codebase

Ref [Card Status Change](https://citz-cdt.atlassian.net/wiki/spaces/BMS/pages/301574382/State+Model#State-Transitions-Triggered-by-Card-Status-Change)

### Developer notes
**We will need to take special care that we are handling the Credential object that gets persisted to native storage correctly, as that is used to determine if a user is verified or not in both v3 and v4, although v4's implementation has maybe been simplified too much. Remember the UAT pointed out that in "emergency mode", a suspended credential may be allowed to work.**

### Related
- #3247 
- #3364

jleach:
## Summary

Audited how iOS and Android legacy apps handle `BCSCReason` values and found 4 gaps in the RN layer. This PR addresses 2 of them:

- Add handling in `DeviceInvalidated` modal for `CanceledByCardTypeChange` and `CanceledDueToInactivity` reasons (both factory reset with empty params, matching the Cancel-group pattern)
- Add `CanceledByCardExpire` enum value to `BCSCReason` (marked legacy/unused — physical card expiry no longer cancels the mobile app)
- Add `CARD_CANCELED_DUE_TO_INACTIVITY` analytics event code to `AppEventCode`
- Add EN/FR/PT-BR translation keys for the 2 new reason display strings
- Add `DeviceInvalidated` component tests
- Add `isUserVerified` and `getCredentialVerificationStatus` utilities to `bcsc-credential.ts` with co-located tests
- Narrow `CredentialInfo.bcscEvent` type to a union of known string literals

Fixed #2748
Related: #3389

## Test plan

- [ ] Trigger a `CanceledByCardTypeChange` reason — verify `DeviceInvalidated` modal renders the correct message and factory reset proceeds
- [ ] Trigger a `CanceledDueToInactivity` reason — verify same
- [ ] Run `yarn test` — all new unit tests pass
- [ ] Verify no regression on existing Cancel-group reasons (`CanceledByUser`, `CanceledByAdditionalCard`, `CanceledByAgent`)

### DIFF
--- a/app/src/bcsc-theme/utils/id-token.ts
+++ b/app/src/bcsc-theme/utils/id-token.ts
@@ -40,6 +40,7 @@ export enum BCSCReason {
   CanceledByAdditionalCard = 'Canceled by Additional Card',
   CanceledByCardTypeChange = 'Canceled by Card Type Change',
   CanceledDueToInactivity = 'Canceled due to Inactivity',
+  CanceledByCardExpire = 'Canceled by Card Expire', // Legacy: physical card expiry no longer cancels the mobile app
 }
 
 /**
--- a/app/src/events/appEventCode.ts
+++ b/app/src/events/appEventCode.ts
@@ -27,7 +27,7 @@ export enum AppEventCode {
   ANDROID_DEVICE_PROTECTION_REQUIRED = 'android_device_protection_required',
   CARD_EXPIRED_WILL_REMOVE = 'card_expired_will_remove',
   CARD_STATUS_ADDITIONAL_CARD = 'card_status_additional_card',
-  CARD_CANCELED_DUE_TO_INACTIVITY = 'card_canceled_due_to_inactivity',
+  CARD_CANCELLED_DUE_TO_INACTIVITY = 'card_cancelled_due_to_inactivity',
   CARD_STATUS_CANCELLED = 'card_status_cancelled',
   CARD_STATUS_CANCELLED_BY_USER = 'card_status_cancelled_by_user',
   CARD_STATUS_EXPIRED = 'card_status_expired',
--- a/app/src/events/appEventCode.ts
+++ b/app/src/events/appEventCode.ts
@@ -27,6 +27,7 @@ export enum AppEventCode {
   ANDROID_DEVICE_PROTECTION_REQUIRED = 'android_device_protection_required',
   CARD_EXPIRED_WILL_REMOVE = 'card_expired_will_remove',
   CARD_STATUS_ADDITIONAL_CARD = 'card_status_additional_card',
+  CARD_CANCELED_DUE_TO_INACTIVITY = 'card_canceled_due_to_inactivity',
   CARD_STATUS_CANCELLED = 'card_status_cancelled',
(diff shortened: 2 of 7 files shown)


app/src/events/appEventCode.ts
```diff
@@ -27,6 +27,7 @@ export enum AppEventCode {
   ANDROID_DEVICE_PROTECTION_REQUIRED = 'android_device_protection_required',
   CARD_EXPIRED_WILL_REMOVE = 'card_expired_will_remove',
   CARD_STATUS_ADDITIONAL_CARD = 'card_status_additional_card',
+  CARD_CANCELLED_DUE_TO_INACTIVITY = 'card_cancelled_due_to_inactivity',
```
KjartanE, line 30: This `CARD_CANCELLED_DUE_TO_INACTIVITY` looks like its never used anywhere else

app/src/bcsc-theme/utils/id-token.ts
```diff
@@ -40,6 +40,7 @@ export enum BCSCReason {
   CanceledByAdditionalCard = 'Canceled by Additional Card',
   CanceledByCardTypeChange = 'Canceled by Card Type Change',
   CanceledDueToInactivity = 'Canceled due to Inactivity',
```
KjartanE, line 42: EventReasonAlertsSystemCheck – No tests for the new cancel reasons (CanceledByCardTypeChange, CanceledDueToInactivity).

app/src/events/appEventCode.ts
jleach: Yes, I added it mostly for me to remember it exists in the docs. It's defined in the legacy code. We catch it as part of a grouping effect on all "Canceled" events. They all trigger the same workflow under the hood because AFAIK we don't have specific logic in legacy for the different cancelation reasons. Thoughts? Happy to remove it - it's dead code.

app/src/bcsc-theme/utils/id-token.ts
jleach: Ya, but no.

  - runCheck() (line 55) in `EventReasonAlertsSystemCheck.ts` returns on BCSCEvent.Cancel — it doesn't care about the specific BCSCReason. All "Cancel" reasons take the same path.
  - onFail() (line 74)  in the same file switches on `BCSCEvent.Cancel ` to navigates to DeviceInvalidated passing this.reason through — but the system check  itself doesn't behave differently per reason.

  So the existing Cancel tests  cover the system check behaviour. Pretty trivial to add more though. LMK.

KjartanE (approved):
Good to go, tested locally. 🐱

PR #3399, opened 2026-03-10
fix: ios scaling webview content

KjartanE:
# Summary of Changes

Adjusts in-app webview scaling on iOS so content fits better on screen and remains readable without extra zooming.

# Testing Instructions

- On an iOS device or simulator, open screens that use the in-app webview and confirm the page scales correctly on first load (no clipped content, no excessive zoom).
- Rotate the device and navigate between a few webview screens to ensure scaling stays consistent and text remains readable.

# Acceptance Criteria

- iOS webview content fits within the visible area on initial load without requiring manual zooming.
- Scaling remains stable across navigation and orientation changes.

image

# Related Issues

N/A

### DIFF
--- a/app/src/bcsc-theme/utils/webview-utils.ts
+++ b/app/src/bcsc-theme/utils/webview-utils.ts
@@ -127,16 +127,17 @@ export const createWebViewJavascriptInjection = (colorPalette: IColorPalette, fo
 
       const style = document.createElement('style');
       style.textContent = \`
+        html, body, body * {
+          ${fontSizeCss}
+        }
         body, body * {
           background-color: ${colorPalette.brand.primaryBackground} !important;
           color: ${colorPalette.brand.secondary} !important;
-          ${fontSizeCss}
         }
         a, a *, a:visited, a:visited *, a:hover, a:hover *, a:active, a:active * {
           color: ${colorPalette.brand.link} !important;
           text-decoration-color: ${colorPalette.brand.link} !important;
           border-color: ${colorPalette.brand.link} !important;
-          ${fontSizeCss}
         }
       \`;
       document.head.appendChild(style);
--- a/app/src/bcsc-theme/utils/webview-utils.ts
+++ b/app/src/bcsc-theme/utils/webview-utils.ts
@@ -89,32 +89,38 @@ ${bodyContent}
 </html>`
 }
 
-export const createFontScalingScript = (): string => {
-  const fontScale = Dimensions.get('window').fontScale
-  if (Platform.OS === 'ios') {
-    return `
-      const fontScale = ${fontScale};
-      document.documentElement.style.fontSize = (16 * fontScale) + 'px';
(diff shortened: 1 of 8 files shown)


app/src/bcsc-theme/utils/webview-utils.ts
```diff
@@ -89,38 +89,47 @@ ${bodyContent}
...
-      ${createFontScalingScript()}
+      ${createFontScalingScript(baseFontSizePx)}
       document.querySelectorAll('footer, header, h1, nav[aria-label="breadcrumb"]').forEach(el => el.remove());
       document.body.style.backgroundColor = '${colorPalette.brand.primaryBackground}';
       document.body.style.color = '${colorPalette.brand.secondary}';
 
       const style = document.createElement('style');
       style.textContent = \`
+        html, body, body * {
+          ${fontSizeCss}
```
jleach: I think font size is applied 2x. Once on like 123 and again on line 131. No?

app/src/bcsc-theme/utils/webview-utils.ts
```diff
@@ -89,38 +89,47 @@ ${bodyContent}
...
+ * This includes applying iOS font scaling (Android uses WebView textZoom), setting the background
+ * color, text color, and link colors to match the app theme, and removing page chrome.
  *
- * @param {IColorPalette} colorPalette - The color palette object containing brand colors
- * @returns {*} {string} JavaScript string to be injected into the WebView
+ * @param colorPalette - The color palette object containing brand colors
+ * @param fontScale - Device font scale (e.g. from useWindowDimensions().fontScale)
  */
-export const createWebViewJavascriptInjection = (colorPalette: IColorPalette): string => {
+export const createWebViewJavascriptInjection = (colorPalette: IColorPalette, fontScale: number): string => {
```
jleach, line 132: I like passing in `fontScale` here but should we also apply the pattern to `createTermsOfUseHtml` and pass in fontScale? It reads it from `Dimensions.get`?

jleach (changes requested):

app/src/bcsc-theme/utils/webview-utils.ts
KjartanE: the Dynamic implementation from `createFontScalingScript` did not cover all text in screen. But removing that dynamic and only having font scaling on the `body, body *` covers all text.

app/src/bcsc-theme/utils/webview-utils.ts
KjartanE: great idea, I will look into updating 👍

app/src/bcsc-theme/utils/webview-utils.ts
KjartanE: Found a solution that covers all possible inputs for web views and scales accordingly. Only applied once!

jleach (approved):

PR #3405, opened 2026-03-10
fix: prevent verification loop by persisting account status and improving auth routing

MacDeluca:
# Summary of Changes

  - Add useInitializeAccountStatus hook to handle account hydration with retry logic, replacing inline logic in RootStack
  - Move hasAccount from volatile bcscSecure (native-only) to persistent bcsc (AsyncStorage-backed) to prevent account status loss on app wake/background
  - Replace falsy (!value) routing checks in RootStack with strict equality (=== false / === true) so that undefined or null state no longer silently matches a condition
  - Add explicit fallback to AuthStack when verification state is indeterminate, preventing users from landing on MainStack without proper authentication
  - Fix retryAsync dropping retryIfNullish flag after error recovery

  ## Problem

  In rare cases, store values like verified or hasAccount can be undefined (e.g. state loss, incomplete hydration). The previous falsy checks treated undefined the same as false, which could route users to the wrong stack — including showing MainStack as the implicit default without an auth gate. Additionally, hasAccount was only stored in native secure storage, so a flaky getAccount() call on app wake could reset users back to onboarding.

# Testing Instructions
  - Fresh install → lands on OnboardingStack
  - Existing account, not authenticated → lands on AuthStack
  - Authenticated, not verified → lands on VerifyStack
  - Authenticated and verified → lands on MainStack
  - Factory reset → returns to OnboardingStack
  - Lock phone screen, wait 30+ seconds, wake → app should NOT show onboarding
  - Force-kill and relaunch → app should show auth screen, not onboarding

# Acceptance Criteria

# Related Issues

### DIFF
--- a/app/src/bcsc-theme/api/hooks/useInitializeAccountStatus.tsx
+++ b/app/src/bcsc-theme/api/hooks/useInitializeAccountStatus.tsx
@@ -1,0 +1,62 @@
+import { retryAsync } from '@/bcsc-theme/utils/retry'
+import { BCDispatchAction, BCState } from '@/store'
+import { TOKENS, useServices, useStore } from '@bifold/core'
+import { useCallback, useEffect, useState } from 'react'
+import { getAccount } from 'react-native-bcsc-core'
+
+const ACCOUNT_FETCH_MAX_RETRIES = 3
+const ACCOUNT_FETCH_RETRY_DELAY_MS = 500
+
+/**
+ * Hook to initialize the account status in the global store on app startup.
+ *
+ * @returns An object containing the initializingAccount boolean which indicates if the account status is being initialized.
+ */
+export const useInitializeAccountStatus = () => {
+  const [store, dispatch] = useStore<BCState>()
+  const [logger] = useServices([TOKENS.UTIL_LOGGER])
+  const [hasCheckedAccount, setHasCheckedAccount] = useState(Boolean(store.bcsc.hasAccount))
+
+  const initializingAccount = store.stateLoaded && !hasCheckedAccount
+
+  /**
+   * Checks for an existing BCSC account and updates the global store accordingly.
+   * Note: Will try to fetch the account multiple times with a delay in case of transient errors
+   *
+   * @returns Promise that resolves when the account status has been initialized
+   */
+  const initializeAccountStatus = useCallback(async () => {
+    try {
+      // TEMP (MD): Patch until BcscCore package correctly bubbles errors https://github.com/bcgov/bc-wallet-mobile/issues/3372
+      const account = await retryAsync(getAccount, ACCOUNT_FETCH_MAX_RETRIES, ACCOUNT_FETCH_RETRY_DELAY_MS, true)
+
+      dispatch({ type: BCDispatchAction.SET_HAS_ACCOUNT, payload: [Boolean(account)] })
+
+      if (account?.nickname && !store.bcsc.nicknames.includes(account.nickname)) {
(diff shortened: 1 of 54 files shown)


app/src/bcsc-theme/api/hooks/useInitializeAccountStatus.tsx
```diff
@@ -0,0 +1,62 @@
...
+      const account = await retryAsync(getAccount, ACCOUNT_FETCH_MAX_RETRIES, ACCOUNT_FETCH_RETRY_DELAY_MS, true)
+
+      dispatch({ type: BCDispatchAction.SET_HAS_ACCOUNT, payload: [Boolean(account)] })
+
+      if (account?.nickname && !store.bcsc.nicknames.includes(account.nickname)) {
+        dispatch({ type: BCDispatchAction.ADD_NICKNAME, payload: [account.nickname] })
+      }
+    } catch (error) {
+      logger.error('[useInitializeAccountStatus] Error checking for existing account:', error as Error)
+      dispatch({ type: BCDispatchAction.SET_HAS_ACCOUNT, payload: [false] })
```
jleach: Not sure how `retryAsync` works but, if it fails a few times (since it retries) would it trigger this dispatch making it seem like the user is trapped in a onboarding loop? Maybe either not dispatching on error or dispatching only if `hasAccount` was already false in the store.

app/src/store.tsx
```diff
@@ -577,14 +583,8 @@ const bcReducer = (state: BCState, action: ReducerAction<BCDispatchAction>): BCS
...
     case BCSCDispatchAction.SUCCESSFUL_AUTH: {
-      const bcscSecure = { ...state.bcscSecure, hasAccount: true }
       const authentication = { ...state.authentication, didAuthenticate: true }
-      return { ...state, bcscSecure, authentication }
-    }
-    case BCSCDispatchAction.SET_HAS_ACCOUNT: {
-      const hasAccount = (action?.payload || []).pop() ?? false
-      const bcscSecure = { ...state.bcscSecure, hasAccount }
-      return { ...state, bcscSecure }
+      return { ...state, authentication }
```
jleach: The comment above the reducer batched `hasAccount` and `didAuthenticate` to prevent re-render. This change might be de-batching which could cause a flicker - no? If nothing else a comment update for clarity?

jleach (changes requested):

app/src/store.tsx
MacDeluca: Yea I agree that it's better batched. Updated.

app/src/bcsc-theme/api/hooks/useInitializeAccountStatus.tsx
MacDeluca: `retryAsync` will just retry the callback n times (if throws or returns null) with a delay in between. 

This section of the code is mostly trying to circumvent that strange transient error that UAT is getting where `getAccount` is returning null on device wakeup (very rarely...)

Updated to drop the on error dispatch call.

jleach (approved):
- [x] Fresh install → lands on OnboardingStack
- [x] Existing account, not authenticated → lands on AuthStack
- [x] Authenticated, not verified → lands on VerifyStack
- [x] Authenticated and verified → lands on MainStack
- [x] Factory reset → returns to OnboardingStack
- [x] Lock phone screen, wait 30+ seconds, wake → app should NOT show onboarding
- [x] Force-kill and relaunch → app should show auth screen, not onboarding

PR #3408, opened 2026-03-11
fix: don't settle promise on intermediate biometric failure (Android)

### LINKED ISSUE #3281: Biometric unlock fails after initial fingerprint rejection
## Description

On Android, when using fingerprint biometrics to unlock the app via the "Continue as" flow, a failed fingerprint attempt (e.g. unregistered finger) causes the subsequent successful attempt to dismiss the biometric prompt without actually unlocking the app. The user must re-activate the biometric prompt and succeed on the first attempt to unlock successfully.

https://github.com/user-attachments/assets/bd9abfc9-6bc2-493f-bf48-77d14c65e58f

## Steps to Reproduce

1. Start the Android app with a BCSC account already set up
2. On the lock screen, tap **Continue as \<account\>**
3. When the fingerprint prompt appears, scan an **unregistered** finger — the prompt will show a failure/error state and the screen glitches
4. Without dismissing the prompt, scan the **correct registered finger**
5. The biometric prompt dismisses (indicating success), but the app remains locked

## Expected Behaviour

After a failed attempt followed by a successful fingerprint scan, the app should unlock and proceed normally.

## Actual Behaviour

The biometric prompt dismisses as though authentication succeeded, but the app does not unlock. The user must retry the entire flow and succeed on the **first** fingerprint scan attempt.

## Environment

- Platform: Android
- Build target: BCSC (v4)
- Biometric method: Fingerprint

## Notes

- Make sure the root cause is understood first. Is this a RN re-rendering or loss of state?
- The issue does not reproduce if the correct finger is used on the first attempt
- The screen glitch on the failed attempt may be a related symptom worth investigating
- Likely related to how the biometric `onAuthenticationError` / `onAuthenticationFailed` callback state is managed — a failed attempt may leave the authentication state in an inconsistent condition that causes a subsequent success to be mishandled

jleach:
## Summary
- On Android, `BiometricPrompt.onAuthenticationFailed()` is an intermediate callback — the prompt stays open for retry. The code was treating it as terminal by immediately resolving the React Native promise with `{ success: false }`, causing subsequent successful attempts to be silently ignored.
- Stop forwarding `FAILED` to the callback in `DeviceAuthenticationService.kt` and make the `FAILED` branch a no-op (log only) in both `performDeviceAuthentication` and `unlockWithDeviceSecurity` in `BcscCoreModule.kt`.

Fixes #3281

## Manual Testing

- [ ] On Android, trigger biometric prompt via "Continue as" flow
- [ ] Scan wrong finger, then scan correct finger — app should unlock
- [ ] Scan correct finger on first attempt — app should unlock (no regression)
- [ ] Cancel biometric prompt — should navigate back (no regression)
- [ ] Verify iOS biometric flow is unaffected

https://github.com/user-attachments/assets/ac220e23-a813-42ea-8313-8c8d63ce1e1c


al-rosenthal (approved):
testing this locally and works as expected

Issue #3388, opened 2026-03-07
Refreshing tokens too frequently, causing 302 errors


bryce-mcmath:
### Description

The UAT folks seem to be encountering 302 errors (the BCSC 302 error not HTTP 302) in a few different flows and Marcos tracked it down to refreshed tokens that were too new. He also found that token refresh was happening multiple times for a single user in some cases within the span of four seconds. We should match the token refresh strategy of v3 (ias-ios)

@marcos-carretero feel free to add any additional notes here as needed

jleach:
Dev Notes:
- Look at the TTL of the tokens, don't do it earlier than needed
- Could be a react hook related issue (re-render calling refresh).

bryce-mcmath:
5183

Issue #3382, opened 2026-03-06
302 server error after navigating to Account tab in app


gvodden:
### Description of problem
When I navigate to the Account tab, a server error (302 - problem with the service) pops up

### Expected behavior
no server errors

### Steps to reproduce
Set up and verify a BC Services Card (eg: gv0000001 1/1/2000)
Tap the Account tab
wait a moment, and the error pops up

### Screenshots and/or log output Press question mark to see available shortcut keys

<img width="476" height="1033" alt="Image" src="https://github.com/user-attachments/assets/a4c29923-cbe3-4d43-bc08-1a6e22c0d7eb" />

### Environment: SIT

Build #: 5012

Android Device Model: S24, Pixel 8

iOS Device Model: 15+, 6SPlus

KjartanE:
The "Server Error (302)" shown on the Account tab is coming from the API, not from the mobile app. The app is surfacing the error returned by the backend when it requests account data.

KjartanE:
Couldn’t reproduce this on our side. Sending back to UAT for another pass. If it still happens, please share any steps that might differ from the original ones.

gvodden:
If you open the v4 app and navigate to the account tab first, the error pops up.
If you open the app and navigate to the Services tab first, wait for the list to load, and THEN navigate to the Account tab, the error Does Not appear. 
If you close and re-open the app again, and then navigate to Account, the error returns.

I suspect that when you wait for the list of services to load, something else also loads that is not loading when you go to Account first...

see video:

https://github.com/user-attachments/assets/076b47ef-5a0b-4dcc-b3bb-2ec27cc414ee

KjartanE:
I have attempted multiple times, I never get any error. 

https://github.com/user-attachments/assets/63996413-a10a-4b78-ad15-eda58dd3a98f

gvodden:
@KjartanE  Are you using a physical device, or a virtual one?

gvodden:
https://github.com/user-attachments/assets/fcc15b32-523b-48e3-aea0-0f9b47ea0d26
@KjartanE  check the sequence here.
If I don't go to services, the error persists.
If I let Services load first, the error does not persist.

If you are using a virtual device, it is probably loading fast enough in the back end to avoid whatever I am seeing on the physical devices.

KjartanE:
> [@KjartanE](https://github.com/KjartanE) Are you using a physical device, or a virtual one?

I am using a physical device. 
[...]

Issue #3232, opened 2026-02-03
Invalid QR Code message when transferring account


gvodden:
### Description of problem
Transferring results in "server Error 302" and Invalid QR code"

### Expected behavior
Account should transfer 

### Steps to reproduce
Set up BC Services card on one device
Use the Transfer feature on another

After scanning the QR code, error messages will pop up

### Screenshots and/or log output

<img width="362" height="782" alt="Image" src="https://github.com/user-attachments/assets/98486ba5-d628-47a0-b34a-e4f87b2e28ba" />

### Environment: SIT

Build #: 4289

Android Device Model: S24

iOS Device Model: 15+

bryce-mcmath:
Updated ticket description and title, I've seen this going both ways now: iOS <-> Android

gvodden:
For what it's worth, there is a defect for this issue in Jira for the IAS R3.20 testing in the QA environment:  https://citz-cdt.atlassian.net/browse/IASP-18216

gvodden:
The Jira ticket above has been resolved in IAS R3.20 RC5

jleach:
Fixed in 4.0.0-4678.

gvodden:
Just reproduced the error in build 4843, and again in 4857.  Tried to capture it on video but was initially unable to.
After a couple transfers back and forth I managed to capture a few things.

Invalid QR
302 errors on both devices
the ToU is not populating before being displayed (will cap that and submit on its own).

Devices: iPhone 6SPlus (iOS 15.8.6) - build 4857
Galaxy S24 (android 15) - build 4857

Did fresh installs of build 4857 and captured the video below.
https://github.com/user-attachments/assets/35b49f83-24d8-4953-9357-f4f1808332c2

bryce-mcmath:
I am unable to reproduce in build 5012. @gvodden @RaulConstan can you confirm this still exists as an issue? I am able to transfer from iOS to Android, from Android to iOS without errors.

gvodden:
<img width="821" height="855" alt="Image" src="https://github.com/user-attachments/assets/e286a667-6209-474d-997d-a5058b60c748" />

going from android to iOS, I got the invalid message, then I went back and scanned again (it worked on second pass)
Going the other way (iOS to Android), it threw the error, but the other device things it was successful (see screenshot).

In the end, the account actually IS active.

RaulConstan:
[...]

PR #3410, opened 2026-03-12
fix: refresh token issues

bryce-mcmath:
# Summary of Changes

This PR addresses two issues:
- Prevents unnecessary double-fetch of refresh tokens
- Ensures tokens are refreshed a minute before they expire rather than the moment they expire

# Testing Instructions

Go through verification, log in and out, kill the app and restart. Nothing breaks. Feel free to `console.warn('blah')` in the fetchTokens call to confirm it is only called once as-needed

# Acceptance Criteria

Refresh token calls are only made as needed

N/A

# Related Issues

#3388 
#3232 (potentially)
#3382 (potentially)

### DIFF
--- a/app/src/bcsc-theme/api/client.ts
+++ b/app/src/bcsc-theme/api/client.ts
@@ -15,9 +15,9 @@ import { JWK, JWKResponseData } from './hooks/useJwksApi'
 import { TokenResponse } from './hooks/useTokens'
 import { withAccount } from './hooks/withAccountGuard'
 
-// Refresh tokens 60 seconds before they actually expire to avoid
+// Refresh tokens 30 seconds before they actually expire to avoid
 // expiry-on-the-wire races when multiple requests fire near the boundary.
-const TOKEN_EXPIRY_BUFFER_MS = 60 * 1000
+const TOKEN_EXPIRY_BUFFER_MS = 30 * 1000
 
 // Extend AxiosRequestConfig to include skipBearerAuth
 declare module 'axios' {
--- a/app/src/bcsc-theme/api/client.ts
+++ b/app/src/bcsc-theme/api/client.ts
@@ -15,6 +15,10 @@ import { JWK, JWKResponseData } from './hooks/useJwksApi'
 import { TokenResponse } from './hooks/useTokens'
 import { withAccount } from './hooks/withAccountGuard'
 
+// Refresh tokens 60 seconds before they actually expire to avoid
+// expiry-on-the-wire races when multiple requests fire near the boundary.
+const TOKEN_EXPIRY_BUFFER_MS = 60 * 1000
+
 // Extend AxiosRequestConfig to include skipBearerAuth
 declare module 'axios' {
   export interface AxiosRequestConfig {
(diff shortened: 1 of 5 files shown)


app/src/bcsc-theme/hooks/useSecureActions.tsx
```diff
@@ -691,6 +683,15 @@ export const useSecureActions = () => {
         logger.warn('[IsVerified] No credential found but refresh token exists — treating as not verified.')
       }
 
+      if (refreshToken && apiClient && isClientReady) {
+        try {
+          await apiClient.getTokensForRefreshToken(refreshToken)
+        } catch (error) {
+          const message = error instanceof Error ? error.message : String(error)
+          logger.error(`[hydrateSecureState] Failed to refresh tokens with stored refresh token: ${message}`)
```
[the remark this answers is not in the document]
bryce-mcmath: intended behaviour

app/src/bcsc-theme/api/client.ts
```diff
@@ -233,18 +237,25 @@ class BCSCApiClient {
...
-      // access token is expired, fetch new tokens using refresh token
-      return this.getTokensForRefreshToken(this.tokens.refresh_token)
+      // access token is expired or about to expire, fetch new tokens using refresh token
+      // Set tokensPromise immediately to prevent concurrent callers from starting duplicate refreshes
+      this.tokensPromise = this.fetchTokens(this.tokens.refresh_token)
+      try {
+        this.tokens = await this.tokensPromise
+      } finally {
+        this.tokensPromise = null
+      }
```
MacDeluca, line 247: suggestion: should we add `updateTokens` to the client?

```ts
updateTokens(tokens: Tokens | Promise<Tokens>)
```

Then downstream we could also use that instead of `apiClient.tokens = tokens`

app/src/bcsc-theme/api/hooks/useTokens.tsx
```diff
@@ -62,9 +62,8 @@ const useTokenApi = (apiClient: BCSCApiClient) => {
         })
 
         try {
-          // Pass both refreshToken and accessToken to avoid duplicate API call in updateTokens
-          await updateTokens({ refreshToken: data.refresh_token, accessToken: data.access_token })
           apiClient.tokens = data
+          await updateTokens({ refreshToken: data.refresh_token, accessToken: data.access_token })
```
MacDeluca, line 66: question: Did we previously set the client tokens within `updateTokens`?

Just wondering if that would be a better place to update the client tokens as well

bryce-mcmath: I think all the other useSecureActions updateBlah just do two things; update the secure store and persist to native storage. Have one update method that makes an extra API request, but only in certain scenarios seemed distracting. It was also causing refresh token fetch requests when they weren't necessary

app/src/bcsc-theme/api/client.ts
bryce-mcmath: updateTokens currently handles updating native storage and secure storage, like the rest of the updateBlah methods in useSecureActions. I'm leaning towards thinking it would be out of place putting it in the API client, but we can chat about it

KjartanE (approved):
Tested locally, functioning as expected. Looks great!

PR #3418, opened 2026-03-12
fix(verify): show BCSC serial number in Step 2 during NPC flow

jleach:
## Summary

- Fixes regression where Step 2 of the setup flow no longer displayed the BCSC serial number after scanning in the NPC (Non-Photo Card) workflow
- Removed the `!step2Completed` early return in `getStep2Subtext()` so the serial is shown as soon as it's collected, rather than waiting for the 2nd photo ID to also be provided. In a Photo-Card step 2 is done with 1 ID; For a Non-Photo Card there is the 2nd step of entering a 2nd ID.
- Restores v3 behaviour

Screenshot_20260312-140138

## Manual Testing

- [ ] Input a Photo Card, you will see step 2 completed (and others) and the ID Number;
- [ ] Input a NPC, you will see the card number and the app will be waiting for the 2nd ID.

### DIFF
--- a/app/src/hooks/useSetupSteps.test.tsx
+++ b/app/src/hooks/useSetupSteps.test.tsx
@@ -583,6 +583,26 @@ describe('useSetupSteps Hook', () => {
       expect(hook.result.current.id.subtext[0]).toContain('123456789')
     })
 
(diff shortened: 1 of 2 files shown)


al-rosenthal (approved):

PR #3444, opened 2026-03-13
fix: scaling webview

KjartanE:
# Summary of Changes

**Android Terms of Use:** Terms of Use is now rendered from a single HTML document with styling embedded in the markup (via `createTermsOfUseHtml`), so it displays correctly on Android without depending on JS injection timing.

**Where to use WebView:** Adds a `disableInjectedStyles` flag so the onboarding "Where to use" WebView can render without injected theme styling. When the flag is true, the page keeps its own layout and styles instead of having background/colors/links overridden and header/footer removed, fixing the odd button styling on that screen.

- **Navigator params:** `OnboardingWebView` now accepts optional `disableInjectedStyles?: boolean`.
- **WebViewContent:** New prop `disableInjectedStyles`; when true (and loading a URL), skips `injectedJavaScriptBeforeContentLoaded`.
- **IntroCarousel:** Navigates to "Where to use" with `disableInjectedStyles: true`. Other OnboardingWebView call sites unchanged.

# Testing Instructions

1. **Terms of Use:** On Android, open onboarding and go to the Terms of Use screen; confirm the page renders with correct theme colors and layout.
2. Run onboarding and open the "Where to use" slide; tap the button to open the WebView.
3. Confirm the "Where to use" page and its button look correct (page’s own styling).
4. Open another onboarding WebView (e.g. Privacy Policy or Secure App) and confirm theme styling still applies as before.

# Acceptance Criteria

- [ ] Terms of Use page renders correctly on Android (embedded HTML styling).
- [ ] "Where to use" WebView shows the page with correct styling (no broken button).
- [ ] Other onboarding WebViews still use injected theme styling.
- [ ] IntroCarousel test expects `disableInjectedStyles: true` in navigate call.

N/A (or add before/after of "Where to use" screen if helpful)

# Related Issues

N/A


al-rosenthal (approved):
tested on both iOS and android, works as expected

Issue #3384, opened 2026-03-06
Device is using a non-standard keyboard error on samsung phone using samsung keyboard (default)


gvodden:
### Description of problem
when using the app for the first time, an error comes up when entering your nickname

<img width="476" height="1033" alt="Image" src="https://github.com/user-attachments/assets/00b60a20-883c-41ce-9a24-d5291e399255" />

Ref: https://citz-cdt.atlassian.net/wiki/spaces/BMS/pages/301574122/Mobile+App+Alerts#Third-Party-Keyboard-Use-Warning-in-Add-Card-and-Card-Use

### Expected behavior
No error message about the keyboard. 

### Steps to reproduce
install app on android device.
set up app until nickname
tap nickname field 
observe error

### Environment

Build #:

Android Device Model:

iOS Device Model:

bryce-mcmath:
will be in 5266

gvodden:
no longer seeing the message in 5266

PR #3458, opened 2026-03-16
fix: third party keyboard detection

bryce-mcmath:
# Summary of Changes

This PR changes the third party keyboard detection logic from an allowlist approach to a flag approach that matches the existing v3 ias-android app logic. It also includes a new native unit test suite for the method. If an error is thrown during the method, it falls back to false.

# Testing Instructions

Using a Samsung device with default keyboard, enter a text input field and notice there is no more warning about third party keyboards.

# Acceptance Criteria

Samsung (and other standard Android distributors) don't trigger the third party keyboard warning unless using a third party keyboard.

N/A

# Related Issues

#3384


jleach (approved):
LGTM. Tried on a Pixel 7a no issues. I don't have a Samsung phone so will merge and get it into the hands of UAT who does.

Issue #3465, opened 2026-03-16
Lock time showing as unknown minutes


jleach:
### Description of problem

In Settings, the time until autolock is showing as "undefined". See screenshot.

### Steps to reproduce

1. Setup a card, view Settings.

### Screenshots and/or log output

<img width="256" alt="Image" src="https://github.com/user-attachments/assets/899a14f0-8942-417d-8919-182f33800302" />

### Environment

Build #:

Android Device Model:

iOS Device Model:

bryce-mcmath:
5319, verified by @jleach

Issue #3438, opened 2026-03-13
Bad text on background contrast


RaulConstan:
### Description of problem
See the video clip...

### Steps to reproduce
1. Go to contact center
2. click "How to set up the BC Services Card app"
3. Scroll down to the contrast issue: in "Step 2" content.

### Screenshots and/or log output

https://github.com/user-attachments/assets/fe3b7bfe-1e89-4c61-bde5-0cf264fb9462

### Environment

Build #:
5183

Android Device Model:

iOS Device Model:
14 Plus (iOS 26.3.1 with liquid screen on default).

jeznorth:
Related to #3436

bryce-mcmath:
5319

gvodden:
looks good on 5394

Issue #3436, opened 2026-03-13
Conflicting Styling in the Access Services Online WebViews


jeznorth:
When accessing external links from the 'Access Services Online' interface, it appears we're attempting to inject styling onto those pages. This is causing styling conflicts with those webpages and negatively impacts the usefulness of those links (e.g. accessibility issues, etc.)

## Replication Procedure
- Perform a fresh install of the application
- On the 'Access services online' onboarding screen, tap on the 'Where to use' link.
- Tap on any of the links provided on this page.

## Observed Behavior
- In most cases, the resulting pages appear broken, incorrectly styled and in some cases are completely illegible.

## Expected Behavior
- External links that are accessed through this WebView should not have styling injected.

## Devices
- Android 16
- Pixel 6

## Screenshots
<img width="180" height="400" alt="Image" src="https://github.com/user-attachments/assets/79136241-82ac-4eee-8b93-0f3060603482" />

<img width="180" height="400" alt="Image" src="https://github.com/user-attachments/assets/0b4fbe58-4e6d-4b2a-9c89-9fae48366266" />

<img width="180" height="400" alt="Image" src="https://github.com/user-attachments/assets/41468c2d-8bf5-4066-be8b-05e0fedc5f61" />

bryce-mcmath:
5319

bryce-mcmath:
There is a related UAT ticket, leaving that open, closing this one

PR #3467, opened 2026-03-17
fix: simplify webview usage, remove injections

bryce-mcmath:
# Summary of Changes

This PR removes script injection from webviews, simplifies their usage, and corrects missing webviews / webviews that should have been external links. 

It also fixes a small bug related to the auto lock menu text, no longer reading "undefined min", now using a fallback to the default lockout if a preferred lockout is not yet defined.

To define the current state of webview usage etc:
ContactUs is not a webview, Privacy is not a webview, TermsOfUse is a webview but made with our own html from the API. All other webviews should have the usual white backgrounds and as-is styles. The only removal of headers and footers comes from the usage of query param `?fromapp=1`. External link vs internal webview behaviour should match v3.

# Testing Instructions

Go through webviews

# Acceptance Criteria

Webviews no longer have injected styles, except for TermsOfUse (which we create ourselves)

N/A

# Related Issues
#3438 
#3436


app/src/bcsc-theme/features/webview/WebViewContent.tsx
```diff
@@ -117,12 +111,9 @@ const WebViewContent: React.FC<WebViewContentProps> = ({ url, html, onLoaded, di
...
       userAgent="Single App"
-      // Accessibility: Apply font scaling for dynamic text sizing
-      textZoom={Platform.OS === 'android' ? Math.round(fontScale * 100) : undefined}
-      injectedJavaScriptBeforeContentLoaded={
-        html || disableInjectedStyles ? undefined : createWebViewJavascriptInjection(ColorPalette, fontScale)
-      }
-      onMessage={() => {}} // Required for injectedJavaScript to work
+      // Accessibility: Apply font scaling for dynamic text sizing, currently only impacts Android
+      // For iOS we apply font scaling to the HTML content itself, so it scales appropriately
+      textZoom={Math.round(fontScale * 100)}
```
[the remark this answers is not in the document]
bryce-mcmath: textZoom prop is android-only

app/src/bcsc-theme/features/settings/SettingsContent.tsx
```diff
@@ -192,7 +198,7 @@ export const SettingsContent: React.FC<SettingsContentProps> = ({
   const showChangePIN = accountSecurityMethod !== AccountSecurityMethod.DeviceAuth && onChangePIN
   const showEditNickname = store.bcscSecure.verified && onEditNickname
   const analyticsOptInText = store.bcsc.analyticsOptIn ? 'ON' : 'OFF'
-  const autoLockTimeText = `${store.preferences.autoLockTime} min`
+  const autoLockTimeText = `${store.preferences.autoLockTime ?? DEFAULT_AUTO_LOCK_TIME_MIN} min`
```
jleach, line 201: This might fix #3465. Assigned to you..

app/src/bcsc-theme/navigators/MainStack.tsx
```diff
@@ -134,7 +132,6 @@ const MainStack: React.FC = () => {
           options={{
             headerShown: true,
             title: t('BCSC.Screens.Settings'),
-            headerBackTestID: testIdWithKey('Back'),
```
jleach, line 137: Why can we get rid of all these?

bryce-mcmath: Cause they are already applied by the Stack Navigator screen options above (L103):

```
<Stack.Navigator
        initialRouteName={initialRouteName}
        screenOptions={{
          // ...default screen options for this stack like `headerBackTestID: testIdWithKey('Back')` that all child screens inherit
        }}
```

app/src/bcsc-theme/navigators/MainStack.tsx
```diff
@@ -163,26 +158,22 @@ const MainStack: React.FC = () => {
...
           options={() => ({
             headerShown: true,
-            headerBackTitleVisible: false,
             headerRight: createMainHelpHeaderButton({ helpCentreUrl: HelpCentreUrl.HOME }),
           })}
         />
         <Stack.Screen
           name={BCSCScreens.MainWebView}
-          component={MainWebViewScreen}
+          component={WebViewScreen}
```
MacDeluca, line 173: question: Shouldn't we still use the MainWebViewScreen vs WebViewScreen?
Isn't this problematic with react navigation and the screen jumping stacks?

edit: Maybe thats my question, does the component reference have an affect on the stack react-navigation uses? Or is it only the screen.

I thought we found that the component reference and the screen name had to be different for it to "honour" the correct stack and not jump randomly.

bryce-mcmath: it's actually the screen name that needs to be distinct, not the component. I kept that

bryce-mcmath: "Reusing a component reference for multiple, distinctly named screens in React Navigation works without issue and is a common practice for reducing code duplication. React Navigation manages screen state based on the screen name, not the underlying component, so each screen behaves as an independent instance in the navigation stack."

app/src/constants.ts
```diff
@@ -49,7 +49,7 @@ export const ACCESSIBILITY_URL =
   'https://www2.gov.bc.ca/gov/content/governments/government-id/bcservicescardapp/accessibility'
 // appending param fromapp=1 to certain id.gov urls automatically removes header and footer and such
 export const HELP_URL = 'https://id.gov.bc.ca/static/help/topics.html?fromapp=1'
-export const SECURE_APP_LEARN_MORE_URL = 'https://id.gov.bc.ca/static/help/secure_app.html'
+export const SECURE_APP_LEARN_MORE_URL = 'https://id.gov.bc.ca/static/help/secure_app.html?fromapp=1'
```
MacDeluca, line 52: thought(no-action): Probaly should update these URL's to template strings that reference the host, I vaguely remember some discussions around environment specific URL's.

```ts
const getIASRootUrl = () => 'https://id.gov.bc.ca'

const SECURE_APP_LEARN_MORE_URL = `${getIASRootUrl()}/static/help/secure_app.html?fromapp=1`
```

jleach (approved):
Tested several web views within the app. All LGTM. Displaying as expected. Terms still has blue gb but this is expected. Ship It.

PR #3473, opened 2026-03-17
feat: add loading indicator to services list screen

### LINKED ISSUE #3433: Add loading indicator to services list
### Description
The services list is missing a loading animation.

### Requirement
- Add a loading indicator to the services list component (Services.tsx)
- Check if services tab loads on stack mount vs screen navigation and if it loads on stack mount possibly change `lazy: false` to `lazy: true` to prevent unnecessary API calls

jleach:
## Summary

- `Services.tsx`: Added an `ActivityIndicator` (large, centered) that renders while service clients are loading, using the `isLoading` value already returned by `useFilterServiceClients` but previously unused. The header and search bar remain visible during loading. Removed the existing `// TODO (MD): implement a loading UI` comment.
- `TabStack.tsx`: Changed `lazy: false` to `lazy: true` so the Services and Account tabs defer mounting until first navigated to, preventing unnecessary API calls on app startup.

Closes #3433

## Visual Testing

The services list is cached by `useDataLoader`, so the spinner will only be briefly visible on first load. For sustained visual testing, add an artificial delay in `useFilterServiceClients.ts` by replacing the `fetchData` argument:

```typescript
// In useFilterServiceClients.ts, replace the fetchData argument:
() => new Promise((resolve) => setTimeout(() => metadata.getClientMetadata().then(resolve), 3000)),
```

This will hold the loading state for 3 seconds, giving enough time to observe the spinner before the list appears.

https://github.com/user-attachments/assets/84f41dfd-5a37-4095-9748-f73cbcbd5986


MacDeluca:
curiosity: On Friday I noticed that we were requesting the "Services / Client Metadata" up to 4 times on app startup. How many times do we make that request now?

Out of scope of this PR, but if we are doing it more than once, we should probably wrap the useFilterServiceClients into a context/provider and fetch it only once to be shared with the components.

jleach:
> curiosity: On Friday I noticed that we were requesting the "Services / Client Metadata" up to 4 times on app startup. How many times do we make that request now?
> 
> Out of scope of this PR, but if we are doing it more than once, we should probably wrap the useFilterServiceClients into a context/provider and fetch it only once to be shared with the components.

I'm not sure. I did notice it seemed to load faster the 2nd time I opend the services screen. I assumed it was caching.

MacDeluca (approved):
Ran locally, works great. 🚢

Issue #3426, opened 2026-03-13
Determine which STUN / TURN servers to use for Live Call WebRTC


bryce-mcmath:
### Description
Title, we need to check with our org as to policy around WebRTC STUN / TURN servers. Check in with Clecio / Marcos before finishing implementation / changes if there are any. See line 321ish of connect.ts

bryce-mcmath:
increasing estimate to a 3 because the work actually involves using stun / turn servers returned from Pexip API which requires a lot of testing and some refactoring of some live call code

PR #3475, opened 2026-03-17
fix: use pexip-provided stun/turn servers

bryce-mcmath:
# Summary of Changes

This PR ensures we use the Pexip provided STUN / TURN servers, falling back to the public one we were already using. Under the hood, the Pexip conference actually happens to use the same STUN server we were already hardcoding, but now if Pexip changes that, it will already be handled.

# Testing Instructions

Do a live call, you still connect to the queue

# Acceptance Criteria

Live call still works

N/A

# Related Issues

#3426

### DIFF
--- a/app/src/bcsc-theme/features/verify/live-call/utils/connect.ts
+++ b/app/src/bcsc-theme/features/verify/live-call/utils/connect.ts
@@ -14,6 +14,26 @@ import { mediaDevices, MediaStream, RTCIceCandidate, RTCPeerConnection } from 'r
 import { RTCOfferOptions } from 'react-native-webrtc/lib/typescript/RTCUtil'
 import type { ConnectionRequest, ConnectResult } from '../types/live-call'
 
+interface PexipIceServerEntry {
+  url?: string
+  username?: string
+  credential?: string
+}
+
+interface PexipTokenResult {
+  token: string
+  participant_uuid: string
+  stun?: PexipIceServerEntry[]
+  turn?: PexipIceServerEntry[]
+  [key: string]: unknown
+}
+
+interface IceServer {
+  urls: string | string[]
+  username?: string
+  credential?: string
+}
+
 // WebRTC Events need handlers even if we don't do anything with some of them
 const noop = () => {}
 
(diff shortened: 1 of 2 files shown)


app/src/bcsc-theme/features/verify/live-call/utils/connect.ts
```diff
@@ -14,6 +14,26 @@ import { mediaDevices, MediaStream, RTCIceCandidate, RTCPeerConnection } from 'r
 import { RTCOfferOptions } from 'react-native-webrtc/lib/typescript/RTCUtil'
 import type { ConnectionRequest, ConnectResult } from '../types/live-call'
 
+interface PexipIceServerEntry {
+  url?: string
```
jleach: I think this needs to be `url?: string[]`. The entry in the SDK uses an array - no? Maybe use the type from the pexip package? Also URL and username are not optional.

```
  // Pexip SDK's actual Turn type (@pexip/infinity-api/dist/token/types.d.ts:134)
  export interface Turn {
    credential?: string
    urls: string[] 
    username?: string
  }
```

jleach (changes requested):

app/src/bcsc-theme/features/verify/live-call/utils/connect.ts
jleach: PS. Looks like the SDK exports all the types you need `Result`, `Stun`, and `Turn`.

jleach (changes requested):

app/src/bcsc-theme/features/verify/live-call/utils/connect.ts
```diff
@@ -314,14 +335,54 @@ const requestInfinityToken = async (request: ConnectionRequest): Promise<any> =>
...
+      }
+    }
+  }
+
+  // Use TURN servers from Pexip token response
+  if (Array.isArray(tokenResult.turn) && tokenResult.turn.length > 0) {
+    for (const entry of tokenResult.turn) {
+      if (
+        entry &&
+        typeof entry.url === 'string' &&
```
jleach: If you change your type above will this break?

app/src/bcsc-theme/features/verify/live-call/utils/connect.ts
```diff
@@ -314,14 +335,54 @@ const requestInfinityToken = async (request: ConnectionRequest): Promise<any> =>
...
+        typeof entry.username === 'string' &&
+        typeof entry.credential === 'string'
+      ) {
+        iceServers.push({
+          urls: entry.url,
+          username: entry.username,
+          credential: entry.credential,
+        })
+      } else if (entry && typeof entry.url === 'string') {
```
jleach: same

jleach (changes requested):

jleach (approved):
LGTM. Tested. Got in the queue and heard the recording.

Issue #3472, opened 2026-03-17
UTF-8 encoding needs a review


RaulConstan:
### Description of problem
Seems the v4 is sending the packages for Send Video in non-UTF-8 format resulting in garbled text for Nickname fields when entered on another language, like Chinese.  See screen shot.

### Steps to reproduce
1. Send request to ID Check where Nickname is Chinese
2. In ID Check, proceed to verify the request.
3. In the last step of verification the garbled text is seen for Nickname

### Screenshots and/or log output

<img width="1728" height="1080" alt="Image" src="https://github.com/user-attachments/assets/f05ed5d1-0b0a-48b1-9f6d-e9eb1ae44a5a" />

<img width="1259" height="551" alt="Image" src="https://github.com/user-attachments/assets/ca4c64d2-17c0-4e8e-b721-cb93ff55981a" />

### Environment
SIT/QA

Build #:
5305

Android Device Model:

iOS Device Model:
14 Plus (iOS 26.3.1 liquid default)

RaulConstan:
Confirmed fixed in 5415:

https://github.com/user-attachments/assets/b6f1819e-20e3-4551-96ef-fed55e32fa6c

PR #3484, opened 2026-03-18
fix: specify utf8 charset

bryce-mcmath:
# Summary of Changes

This PR simply specifies utf-8 as the charset for application/json requests and then adds a bunch of tests to confirm. It does appear to fix the jumbled data the backend is receiving even though that should be the default anyway according to axios docs. 

# Testing Instructions

Do a send video request with non-latin, non-numeric characters (eg. korean) in your nickname or other text fields and ensure they make it successfully through to IDCheck.

# Acceptance Criteria

any utf-8 is preserved

bobby

# Related Issues

#3472

### DIFF
--- a/app/src/bcsc-theme/api/client.ts
+++ b/app/src/bcsc-theme/api/client.ts
@@ -74,7 +74,7 @@ class BCSCApiClient {
     this.logger = logger
     this.client = axios.create({
       headers: {
-        'Content-Type': 'application/json',
+        'Content-Type': 'application/json; charset=utf-8',
       },
     })
 
--- a/app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
+++ b/app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
@@ -187,7 +187,6 @@ const useRegistrationApi = (apiClient: BCSCApiClient | null, isClientReady: bool
       logger.info('Generated dynamic client registration body')
 
       const { data } = await apiClient.post<RegistrationResponseData>(apiClient.endpoints.registration, body, {
-        headers: { 'Content-Type': 'application/json; charset=utf-8' },
         skipBearerAuth: true,
       })
 
@@ -307,7 +306,6 @@ const useRegistrationApi = (apiClient: BCSCApiClient | null, isClientReady: bool
             {
               skipBearerAuth: true,
               headers: {
-                'Content-Type': 'application/json; charset=utf-8',
                 Authorization: `Bearer ${registrationAccessToken}`,
               },
             }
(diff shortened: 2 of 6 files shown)


packages/bcsc-core/android/src/test/java/com/bcsccore/JSONSerializationTest.kt
```diff
@@ -0,0 +1,145 @@
...
+import org.junit.Assert.assertTrue
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+
+/**
+ * Verifies that org.json.JSONObject preserves Unicode characters
+ * in the same patterns used by getDynamicClientRegistrationBody.
+ */
+@RunWith(RobolectricTestRunner::class)
```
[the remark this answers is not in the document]
bryce-mcmath: Doesn't work without it

app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
```diff
@@ -187,7 +187,7 @@ const useRegistrationApi = (apiClient: BCSCApiClient | null, isClientReady: bool
       logger.info('Generated dynamic client registration body')
 
       const { data } = await apiClient.post<RegistrationResponseData>(apiClient.endpoints.registration, body, {
-        headers: { 'Content-Type': 'application/json' },
+        headers: { 'Content-Type': 'application/json; charset=utf-8' },
         skipBearerAuth: true,
```
MacDeluca, line 190: question: Can we remove this header if it's already defined in the client headers?

app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx
```diff
@@ -307,7 +307,7 @@ const useRegistrationApi = (apiClient: BCSCApiClient | null, isClientReady: bool
             {
               skipBearerAuth: true,
               headers: {
-                'Content-Type': 'application/json',
+                'Content-Type': 'application/json; charset=utf-8',
```
MacDeluca: question: same here?

app/src/bcsc-theme/api/client.ts
```diff
@@ -74,7 +74,7 @@ class BCSCApiClient {
     this.logger = logger
     this.client = axios.create({
       headers: {
-        'Content-Type': 'application/json',
+        'Content-Type': 'application/json; charset=utf-8',
```
MacDeluca, line 77: suggestion: I would add a test for this to ensure this header is set correctly

MacDeluca: You could probably go one step further and test to make sure the payload is correctly serialized, but I think the assumption that if this header exists axios will do the correct thing is fine.

MacDeluca (approved):
🚢

PR #3485, opened 2026-03-18
feat: report errors to loki

MacDeluca:
# Summary of Changes

 Error Reporting & Error Boundary

  - Added "Report" button to BCSCErrorModal — users can now report errors directly from the error modal
  - Refactored ErrorBoundary — updated getReportError to use the new toBifoldError utility, and added analytics tracking for boundary-caught errors
  - New toBifoldError utility in error-utils.ts — centralizes conversion from Error/AppError → BifoldError
  - Updated ErrorAlertContext to support the new reporting flow
  - Deprecated UI properties in ErrorRegistry — marking properties that shouldn't live on the registry

# Testing Instructions

App should work the same, errors reported using logger.report.

# Acceptance Criteria

# Related Issues

https://github.com/orgs/bcgov/projects/108/views/1?pane=issue&itemId=164895786&issue=bcgov%7Cbc-wallet-mobile%7C3412


app/src/errors/errorRegistry.ts
```diff
@@ -40,9 +40,9 @@ export interface ErrorDefinition {
   statusCode: number
   /** App event code */
   appEvent: AppEventCode
-  /** i18n key for title */
+  /** @depreacted TODO (MD): Remove UI properties from ErrorDefinition - i18n key for title */
```
jleach: Type-o

app/src/errors/components/ErrorBoundary.tsx
```diff
@@ -33,33 +37,53 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps, ErrorBoundarySta
...
+    logger.error('ErrorBoundary reported:', error)
+    logger.report(reportError)
   }
 
   render(): React.ReactNode {
     const { hasError, error } = this.state
-    const { t } = this.props
 
     if (hasError && error) {
+      const reportError = this.getReportError(error)
```
jleach, line 79: Is this a dupe call from like 70 above?

app/src/contexts/ErrorAlertContext.tsx
```diff
@@ -129,7 +129,7 @@ export const ErrorAlertProvider = ({ children, enableReport = true }: ErrorAlert
       {children}
       <BCSCErrorModal
         error={error}
-        visible={visible}
+        visible={Boolean(error)}
```
jleach: `visible` is derived from `error` (`Boolean(error)`) so the prop no longer carries independent information (was two vars before `visible` and `error`). Inside `BCSCErrorModal`, the guard `if (!visible || !error)` is redundant — both conditions are equivalent now.

Could simplify by removing the `visible` prop entirely and just checking `!error` inside the component. One fewer prop to maintain.

  Non-blocking — fine as a follow-up.

app/src/errors/components/ErrorModal.tsx
jleach: Here ☝🏻

jleach (changes requested):

app/src/contexts/ErrorAlertContext.tsx
MacDeluca: nice catch I will fix

app/src/errors/errorRegistry.ts
MacDeluca: I can't splel

app/src/errors/components/ErrorBoundary.tsx
MacDeluca: One is happening in handleReport() and one in render()

jleach (approved):
Plumbed into `dev` and I can see the logs when I trigger the modal:

Image

Issue #3435, opened 2026-03-13
Device Count Limit message is shown when only 2 devices are listed


RaulConstan:
### Description of problem
For v4, it was decided that 3 is the maximum limit for devices in SIT. (PROD is 5).  However, even at 3 the device limit warning on the Home screen is showing when there are only 2 registered devices.  See the video clip.

Additional note: could also be a delay in the actual 3rd device approved, and the Device Limit screen in the Account Tab simply not being updated correctly.  Means the message is correct, the count on the device limit is correct, but the list of items is incorrect as shown in the vid.

### Expected behavior
Regardless of environments, the limit should reflect actuals and the message should show when the limit is reached.

### Steps to reproduce
1. Have other 2 devices pending (limit should not change until anyone of these is approved)
2. Load 2 devices with the same Identity (limit should be 2)
3. On the second load (step 2 above), the limit should still be 2 and the message should not yet be triggered.

### Screenshots and/or log output

https://github.com/user-attachments/assets/44193373-9915-4908-815f-67b0283b690e

### Environment
SIT

Build #:
5183

Android Device Model:

iOS Device Model:
14 Plus (iOS 26.3.1)

al-rosenthal:
I tested this today on build 5290 and wasn't able to reproduce, I'll ping you after our morning @RaulConstan

RaulConstan:
Confirmed in 5508.

PR #3494, opened 2026-03-18
fix: pending device registration

al-rosenthal:
# Summary of Changes

- added a forced token refresh when account is verified

This forced token refresh kicks the device registration from pending to verified in the database

# Testing Instructions

- Verify an account
- Click ok on success screen
- Immediately navigate to account page
- Open My devices
- see your device in the list

# Acceptance Criteria

You can see your newly verified device in an active state in 'My Devices' webview

N/A

# Related Issues
[BC-Wallet: 3435](https://github.com/bcgov/bc-wallet-mobile/issues/3435)


app/src/bcsc-theme/api/hooks/useCredentialApi.tsx
al-rosenthal, line 1: This was added during investigation into the issue. It wasn't a part of the fix but this endpoint is ready to use when we need it

app/src/bcsc-theme/api/hooks/useApi.tsx
al-rosenthal, line 1: This was added during investigation into the issue. It wasn't a part of the fix but this endpoint is ready to use when we need it

app/src/bcsc-theme/features/verify/_models/useVerificationResponseViewModel.tsx
al-rosenthal, line 1: this actually fixes the issue. Once the api facilitates a new token transaction it kicks the device registration from pending to active. So if the user looks at their devices from the account screen everything is in order

app/src/bcsc-theme/features/verify/_models/useVerificationResponseViewModel.tsx
```diff
@@ -30,26 +32,27 @@ const useVerificationResponseViewModel = () => {
...
     setIsSettingUpAccount(true)
     try {
-      // this marks their account as verified, so we know to navigate them to the correct stack
-      await updateVerified(true)
       // this cleans up old metadata from the verification process (photos, address info)
       await updateUserMetadata(null)
       // this updates their registration status with their nickname and new access tokens
       await handleUpdateRegistration()
+      // force a token exchange so the backend activates the device registration before navigation
+      await token.getCachedIdTokenMetadata({ refreshCache: true })
```
jleach: I think this will fail silently (swallow the error) with just a log message - no? You could try `useTokenService` or build on `app/src/hooks/useAlerts.tsx` to find something that matches. Maybe there is a good match already.

app/src/bcsc-theme/api/hooks/useApi.tsx
```diff
@@ -25,6 +26,7 @@ const useApi = () => {
   const jwks = useJwksApi(apiClient)
   const video = useVideoCallApi(apiClient)
   const deviceAttestation = useDeviceAttestationApi(apiClient)
+  const credential = useCredentialApi(apiClient)
```
jleach: Seems like dead code - no? Should we add in a separate PR when it's needed?

jleach (changes requested):

app/src/bcsc-theme/api/hooks/useApi.tsx
al-rosenthal: I added it while I was investigating the issue and thought to keep it because it didn't exist already. I'll take it out of this PR and if we need it we can add it then

app/src/bcsc-theme/features/verify/_models/useVerificationResponseViewModel.tsx
```diff
@@ -8,48 +9,37 @@ const useVerificationResponseViewModel = () => {
...
+  const { getCachedIdTokenMetadata } = useTokenService()
   const { updateVerified, updateUserMetadata } = useSecureActions()
   const [isSettingUpAccount, setIsSettingUpAccount] = useState(false)
 
   const handleUpdateRegistration = useCallback(async () => {
     try {
-      const registrationAccessToken = store.bcscSecure.registrationAccessToken
-      const selectedNickname = store.bcsc.selectedNickname
-
-      if (!registrationAccessToken) {
```
al-rosenthal, line 19: both of these are handled by `updateRegistration`

jleach (approved):

PR #3506, opened 2026-03-19
feat: e2e testing

KjartanE:
# Summary of Changes

- Adds an **`e2e/` package** with **WebDriverIO + Appium** for BC Services Card flows. It has its own `package.json`; run **`cd e2e && yarn install`** (it is not listed in the root `workspaces.packages` entry, which only includes `app` and `packages/*`).
- **Screen automation model**: a shared **`BaseScreen`** (`e2e/src/screens/BaseScreen.ts`) plus a central **`e2e/src/testIDs.ts`** registry. Specs instantiate `new BaseScreen(TestIDs.SomeFlow)` for typed taps/waits/scrolls—there are no per-screen classes under `e2e/src/screens/bcsc/`.
- **Specs** under `e2e/test/bcsc/`: smoke; onboarding (`simple`, `advanced`); main app (`main`, settings, account, tab navigation); verify subsets (simple/advanced, cards/combo, nickname step, in-person verification); plus `e2e_simple` / `e2e_advanced` orchestrations.
- **Configs** under `e2e/configs/`: shared WDIO base, local Appium (iOS sim, Android emu, USB devices for both), and Sauce Labs RDC.
- **Helpers**: gestures, biometrics, camera, notifications, Sauce annotations, **`iosPermissions`** (e.g. local network / iOS system prompts). Timeouts live in **`e2e/src/constants.ts`**.
- **App / selector alignment**: BCSC UI uses **`testIdWithKey` from `@bifold/core`** (and related patterns) so `testID`s stay stable; the E2E registry uses the resolved **`com.ariesbifold:id/...`** strings Appium targets on both platforms where applicable.
- **`e2e/README.md`** covers setup and run commands; **`e2e/.env.saucelabs.example`** templates cloud credentials.
- **`e2e/apps/README.md`** documents placing **`BCWallet.app`** (simulator), **`BCWallet.ipa`** (physical iOS, with signing/export notes), and **`BCWallet.apk`** (Android), with **`IOS_APP`** / **`ANDROID_APP`** overrides.

# Testing Instructions

1. From repo root: `yarn install` (app and packages).
2. **`cd e2e && yarn install && yarn setup`** (Appium drivers: uiautomator2 + xcuitest).
3. Build and copy binaries into `e2e/apps/` per **`e2e/apps/README.md`** (expected default names: `BCWallet.app`, `BCWallet.ipa`, `BCWallet.apk`).
4. **Local:** iOS sim — `yarn test:ios:local`; Android — start an emulator (e.g. `yarn emulator:android`), then `yarn test:android:local`. Optional: `test:ios:device` / `test:android:device` for USB devices (see `e2e/README.md` for `IOS_UDID`, signing, etc.).
5. **Sauce Labs:** copy `e2e/.env.saucelabs.example` → `.env.saucelabs`, fill credentials, then `yarn test:ios:sauce` / `yarn test:android:sauce` (or `yarn test:sauce`).
6. Spot-check Metro/app build still runs; review **Podfile.lock / gradle lockfiles** if CI pins native deps.

# Acceptance Criteria

- E2E package installs and **`yarn setup`** completes without errors on a clean machine (**Node 20.19+** per root `engines`, documented prereqs).
- At least one **local** target (iOS sim or Android emu) and the **documented Sauce** path can run the suite against a provided build.
- **`testID`s** in app and entries in **`e2e/src/testIDs.ts`** stay aligned where specs depend on them; manual smoke on affected BCSC flows if possible.

N/A (test infrastructure and selector support; no user-facing UI redesign).

# Related Issues

N/A


MacDeluca:
So dope!

I would suggest we try and seperate this PR into a few smaller PR's if possible. At least the BCSC testID changes into it's own PR and any other logical groupings that make sense.

Maybe just the initial e2e screen class and it's dependencies with a single e2e test? Then we can follow this PR with the rest of the tests?

What do you think?

KjartanE:
> So dope!
> 
> I would suggest we try and seperate this PR into a few smaller PR's if possible. At least the BCSC testID changes into it's own PR and any other logical groupings that make sense.
> 
> Maybe just the initial e2e screen class and it's dependencies with a single e2e test? Then we can follow this PR with the rest of the tests?
> 
> What do you think?

I definitely agree, I will look into setting up a PR for updating the app testIDs

e2e/src/screens/bcsc/main/Account.e2e.ts
```diff
@@ -0,0 +1,38 @@
...
+
+class AccountE2EScreen extends BaseScreen {
+  async waitForDisplayed(timeout = 60_000) {
+    const el = await this.findByText('App expiry date')
+    await el.waitForDisplayed({ timeout: timeout })
+  }
+
+  async waitForMyDevices(timeout = 60_000) {
+    await super.waitForDisplayed(timeout, AccountTestIds.MyDevices)
+  }
```
MacDeluca: question: What do we get by adding these additional boilerplate class wrappers around the BaseScreen class / API?

Would it not be better to just call the BaseScreen methods directly inside the tests to prevent having this additional layer for all the e2e tests?

Your BaseScreen class is so awesome and already very descriptive. I dont know how much we get by calling these thin wrapper methods instead of the original methods.

```ts
it('should tap my devices', () => {
    AccountE2EScreen.tapMyDevices() // currently

    BaseScreen.tapByTestId(AccountTestIds.MyDevices) // proposed
})
```

Let me know what I'm missing.

KjartanE: > question: What do we get by adding these additional boilerplate class wrappers around the BaseScreen class / API?
> 
> Would it not be better to just call the BaseScreen methods directly inside the tests to prevent having this additional layer for all the e2e tests?
> 
> Your BaseScreen class is so awesome and already very descriptive. I dont know how much we get by calling these thin wrapper methods instead of the original methods.
> 
> ```ts
> it('should tap my devices', () => {
>     AccountE2EScreen.tapMyDevices() // currently
> 
>     BaseScreen.tapByTestId(AccountTestIds.MyDevices) // proposed
> })
> ```
> 
> Let me know what I'm missing.

The wrappers are thin on purpose: they’re a page object layer, not extra logic.

Mainly looking at Readability: AccountE2EScreen.tapMyDevices() tells you which screen you’re on; a wall of BaseScreen.tapByTestId(...) doesn’t.

One place for IDs — testIDs live with that screen so updates don’t scatter through every spec.

BaseScreen stays the shared primitives; the screen class is just a named façade so tests stay scannable.

MacDeluca: I guess I just feel like it's quite a lot of extra code to maintain.

I feel we get a similar experience in the test suites by just using:
```ts
const AccountE2EScreen = new BaseScreen()
const AccountTestIds = {
    MyDevices = 'testid:blah'
}

it("blah", () => {
    AccountE2EScreen.tapByTestId(AccountTestIds.MyDevices)
})
```

@al-rosenthal @bryce-mcmath @jleach Other devs feel free to chime in on what you think.

bryce-mcmath: @KjartanE with this most recent suggestion from Mac we get very similar readability with a lot less code

KjartanE: I think code readability is important, for clarity and simplicity having direct function names explain the actions in which the test is taking. 

Example:
```ts
it('should navigate through the Setup Types screen', async () => {
    await SetupTypesE2EScreen.waitForDisplayed()
    await SetupTypesE2EScreen.selectSomeoneElseId()
    await SetupTypesE2EScreen.selectOtherPersonPresentYes()
    await SetupTypesE2EScreen.tapContinue()
  })
```

vs:

```ts
const SetupTypesE2EScreen = new BaseScreen()
const SetupTypesTestIds = {
  Continue: 'com.ariesbifold:id/Continue',
  SomeoneElseIdRadioGroup: "com.ariesbifold:id/MyOwnIdRadioGroup-option-SomeoneElse'sID",
  OtherPersonPresentRadioGroupYesOption: 'com.ariesbifold:id/OtherPersonPresentRadioGroup-option-Yes',
}

it("should navigate through the Setup Types screen", () => {
   await SetupTypesE2EScreen.waitForDisplayed(20_000, SetupTypesTestIds.Continue)
   await SetupTypesE2EScreen.tapByTestId(SetupTypesTestIds.SomeoneElseIdRadioGroup)
   await SetupTypesE2EScreen.tapByTestId(SetupTypesTestIds.OtherPersonPresentRadioGroupYesOption)
   await SetupTypesE2EScreen.tapByTestId(SetupTypesTestIds.Continue)
})
```

In addition we are not tracking all testIDs in example two. If we did then we are approaching example one.

bryce-mcmath: I think I might be less invested than you two but unless someone else wants to be the tie breaker I think it's fine to go with the current approach - to me there is slightly more appeal to increasing readability than reducing boilerplate, since code is so easy to write lately

bryce-mcmath: easy to change later on. one thing we could also do is add `press`, `tap`, `check`, `select` methods to BaseScreen that are all under the hood tapByTestId, but which allows for distinguishing from the "feel" of the action contextually in a similar way. can also set a default timeout in BaseScreen, make it an optional second argument to waitForDisplayed. that again would increase the readability but would remove the need for it. we can then pass a set variable with a clear name rather than the number directly when needed

KjartanE: Yeah its a kinda a lot cleaner :D

MacDeluca: > easy to change later on. one thing we could also do is add `press`, `tap`, `check`, `select` methods to BaseScreen that are all under the hood tapByTestId, but which allows for distinguishing from the "feel" of the action contextually in a similar way. can also set a default timeout in BaseScreen, make it an optional second argument to waitForDisplayed. that again would increase the readability but would remove the need for it. we can then pass a set variable with a clear name rather than the number directly when needed

Great suggestion, love the idea of adding some additional "renamed" methods. Reminds me of Jest ie: `it` / `test`

e2e/src/testIDs.ts
MacDeluca, line 1: praise: ❤️

e2e/src/screens/bcsc/main/Account.e2e.ts
al-rosenthal: I like what Kjartan has going on here, I find it easier to read. It adds overhead for writing a test but it's also nice that it's self contained. I think the eventual goal is to get some less technical people helping out with the test writing and `.waitForDisplay()` is a lot easier to understand than `.waitForDisplay(20_000, SetupTypeTestIds.Continue)`

jleach: I think @MacDeluca is right in that we don't gain much but I do prefer the readability of @KjartanE example above. I know that's not helpful, but maybe this is:

I think tests are an area we can lean more into AI to write to avoid "test rot". Tests are not user facing and we don't ship them into production so it's a great fit for AI. I'd probably just tell it to write or fix a test using established patterns and best practices. What would be helpful to me is being able to look over it's work with the maximum readability so I can read them like a story. You know, so it's not gas-lighting me with stuff like `expect(true).toBe(true)`. So my vote is for @KjartanE's notation and @bryce-mcmath additional methods.

**TL;DR**: If AI is going to write tests then favour reviewability over writability.

Just my 2-tokens worth.

KjartanE: After looking at the massive file number for "Files Changed", I agreed with @MacDeluca that it was to much more code to deal with long term. Where any page updated later would have to update the e2e page as well. In addition to the search fields filling up with e2e screens, I think removing the bulk is a great idea. 

Having one file to update any new `testIDs` we may want to interact with is alot cleaner and simler. I think a hybrid approach to what @bryce-mcmath mentioned, A simple naming convention but abstracting the BaseScreen to take the testIDs as an input param.

```ts
const AccountSetup = new BaseScreen(TestIDs.AccountSetup)

  it('should display the Account Setup screen and tap Add Account', async () => {
    await acceptLocalNetworkPermissionIfPresent()
    await AccountSetup.waitFor('AddAccount')
    await AccountSetup.tap('AddAccount')
  })

export const TestIDs = {
  AccountSetup: {
    AddAccount: 'com.ariesbifold:id/AddAccount',
    TransferAccount: 'com.ariesbifold:id/TransferAccount',
    DeveloperMode: 'com.ariesbifold:id/DeveloperMode',
  },
...
```

The abstract functions are context aware to the name and location of the testID.

jleach (approved):

Issue #3492, opened 2026-03-18
Form Labels Do Not Render Correct Font Family 'BC Sans'


jeznorth:
The form labels are not rendering 'BC Sans'

## Expected Behaviour
- Font rendering should be consistent with standards unless otherwise specified.

## Steps to reproduce
- Start setting up the app using 'Setup with other ID'
- Select 'Canadian Citizenship Card'
- Continue through the steps until you see the form.

## Screenshots
<img width="140" height="400" alt="Image" src="https://github.com/user-attachments/assets/39cc365f-2244-4bca-bc37-9a4de703fbcb" />

## Environment
- BC Services Card Dev
- Build #5394

Android Device Model:
- Pixel 6

iOS Device Model:
- ?

Issue #3490, opened 2026-03-18
Enter Card Serial Number - Validation Message Styling / Formatting Issues


jeznorth:
The validation message is incorrectly spaced and does not meet accessibility standards.

## Expected Behaviour
- Spacing / colour should be consistent with design specification.
- Input label should appear as 'bold' (e.g. 'Enter serial number'). 
- Page title is incorrect size.

### Form Design Specifications
https://www.figma.com/design/LaNyNJGxNMsxlqUCNTWy1X/BCSC-V4---Phase-1?node-id=2911-210&t=QG4V5DCR1i4T2g0Q-1

### Page Formatting (Page title)
https://www.figma.com/design/LaNyNJGxNMsxlqUCNTWy1X/BCSC-V4---Phase-1?node-id=1638-10502&t=uosDz4ybVookzD4b-1

## Screenshot
<img width="140" height="400" alt="Image" src="https://github.com/user-attachments/assets/7a120cb1-7970-494a-a5fd-b53d48549aa8" />

## Additional Notes
It might be worth looking at the 'Account Nickname' screen to ensure it's using a standardized pattern.

PR #3520, opened 2026-03-21
fix: replace custom textinputs and limitedtextinput for inputwithvalidation

bryce-mcmath:
# Summary of Changes

This PR fixes a couple minor issues with forms by making all  standard form inputs in bcsc all use the same InputWithValidation component rather than a mix of that, LimitedTextInput, and custom TextInputs. The only remaining custom inputs are the pairing code and services search input, which both seemed different enough to be left as is.

One thing we don't have a ticket for but should probably be up for discussion is clearing errors on user input. We're probably at the point where adding zod or similar makes sense

# Testing Instructions

Check out the inputs: nickname, serial, email

# Acceptance Criteria

Inputs still work and appear as they should as per Figma

input_error
input_highlighted
input_email

# Related Issues

#3490
#3492


al-rosenthal (approved):
tested locally and looks good on iOS and android

Issue #3500, opened 2026-03-19
BCSC Variants (QA, TEST, PROD)


cvarjao:
- Icons Assets
- Create variant bundle

bryce-mcmath:
Assets and variant bundles are done, last thing that needs to be updated is the google services (firebase) file secrets. @jleach is doing this, it will require no coding changes just updated secrets

jleach:
Confirmed publishing.

PR #3522, opened 2026-03-22
cicd: qa, test, prod variants

bryce-mcmath:
# Summary of Changes
Adds all assets and variant config for qa, test, and prod. Adds qa, test, and prod to CICD, only building on merge to, or workflow dispatch from, main; to save time, PRs only build bcsc-dev and bcwallet-prod

# Testing Instructions

N/A

# Acceptance Criteria

N/A

N/A

# Related Issues

#3500


.github/workflows/main.yaml
```diff
@@ -563,8 +613,23 @@ jobs:
           plutil -replace aps-environment -string production app/ios/BCWallet/BCWallet.entitlements
           plutil -p app/ios/BCWallet/BCWallet.entitlements
 
+      - name: Create prod signing credentials
```
jleach, line 626: Was the fork guard (`if: github.event.pull_request.head.repo.fork != true`) intentionally removed here? It was on this step before — just want to make sure it's not an accident from the reorder. Without this guard PRs on forks will fails because they don't have access to secrets.

.github/workflows/main.yaml
```diff
@@ -563,8 +613,23 @@ jobs:
...
+        uses: ./.github/workflows/actions/create-ios-sig-creds
+        with:
+          certificate: ${{ secrets.APPLE_APP_STORE_BUILD_CERTIFICATE_BASE64 }}
+          certificate_password: ${{ secrets.APPLE_APP_STORE_BUILD_CERTIFICATE_PASSWD }}
+          app_store_connect_private_key: ${{ secrets.APP_STORE_CONNECT_PRIVATE_KEY_95 }}
+          provisioning_profile: ${{ secrets[env.IOS_PROVISIONING_PROFILE_SECRET] }}
+
       - name: Generate iOS signing config
```
jleach, line 637: If we're restoring the fork guard on the step above, we'll probably need it here too — the `PP_UUID` extraction will fail if the provisioning profile secret is empty on fork PRs.

jleach (approved):

PR #3548, opened 2026-03-24
feat: recover mediation when the connection activity has expired

jamshale:
# Summary of Changes

This PR will tag mediation connections with a last seen value. Then if an initialization/pickup fails when a connection has expired due to lack of activity the agent will delete the mediation records and create new ones, and register the previous recipient keys with the mediator. 

This allows the mediator to delete records when they have become stale and prevent the DB from growing indefinitely. If an agent is started up after activity check has expired it can recover and existing connections will continue to work with a new mediation record.

An edge case where the mediation is created multiple times without it being deleted in the mediator DB will be handled on the mediator side.

Only will try to recover if the expiry time has be reached. Meaning the wallet hasn't been active or opened for that period of time. This expiry time should be coordinated between the cron job clean up process, but has some fail proof guards if they weren't coordinated.

# Testing Instructions
1. Initialize a new agent and establish a mediation connection.
2. Create connections with issuers/verifiers
3. Close the agent.
4. Delete the mediation and related records from the mediator DB.
5. Reopen the existing agent with expiry set to 0 days.
6. The initial connection pickup request will timeout and a new mediation connection will be established. There will be an error displayed in dev mode saying pickup failed. This is expected
7. Existing issuer/verifier connections continue to work as if nothing happened.

Edge cases. 
 - With the mediator update to enable this feature we have planned the mediation will recover even if it's not deleted from the mediation DB. It will remove the recipient keys from the old mediation record and allow them to be created on a new mediation record.
 - If the recover process fails for some reason like connection issues it will attempt to restore to previous state.

Other services:
 - Mediator --> https://github.com/openwallet-foundation/didcomm-mediator-credo/tree/live-mode-postgres commit https://github.com/openwallet-foundation/didcomm-mediator-credo/commit/4101fa818206d14d4a341a72f7d54da45bc13d0c . Contains tagging and allowing a new mediation record with same recipient keys features
 - https://github.com/openwallet-foundation/acapy-tools/pull/86 This is the current cron job for cleaning mediation records. example usage 
 ```
askar-tools --strategy mediator-cleanup --uri postgres://myuser:mypassword@localhost:5432/mediator_wallet_test --wallet-name mediator_wallet_test --wallet-key mediator_wallet_key --inactive-days-threshold 0 --pickup-repository-uri postgres://myuser:mypassword@localhost:5432/messagepickuprepository
```

### DIFF
--- a/app/src/hooks/useBCAgentSetup.ts
+++ b/app/src/hooks/useBCAgentSetup.ts
@@ -35,6 +35,7 @@ import { Config } from 'react-native-config'
 import { CachesDirectoryPath } from 'react-native-fs'
 
 const DEFAULT_MEDIATION_EXPIRED_THRESHOLD_DAYS = '90'
+const CONNECTION_COMPLETION_TIMEOUT_MS = 10000
 
 const loadCachedLedgers = async (): Promise<IndyVdrPoolConfig[] | undefined> => {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -195,7 +196,8 @@ const useBCAgentSetup = () => {
     if (connection.state === DidExchangeState.Completed) {
       return connection
     }
-    return new Promise((resolve) => {
+
+    return new Promise<ConnectionRecord>((resolve, reject) => {
       const listener = (event: ConnectionStateChangedEvent) => {
         const { connectionRecord } = event.payload
 
@@ -204,12 +206,22 @@ const useBCAgentSetup = () => {
           connectionRecord.state === DidExchangeState.Completed &&
           connectionRecord.isReady
         ) {
-          agent.events.off(ConnectionEventTypes.ConnectionStateChanged, listener)
+          cleanup()
           resolve(connectionRecord)
         }
       }
 
+      const cleanup = () => {
+        agent.events.off(ConnectionEventTypes.ConnectionStateChanged, listener)
+        clearTimeout(timeoutId)
+      }
+
       agent.events.on(ConnectionEventTypes.ConnectionStateChanged, listener)
+
+      const timeoutId = setTimeout(() => {
+        cleanup()
+        reject(new Error(`Timed out waiting for connection ${connection.id} to complete`))
+      }, CONNECTION_COMPLETION_TIMEOUT_MS)
     })
   }
 
@@ -262,14 +274,20 @@ const useBCAgentSetup = () => {
       }
 
       const daysSinceLastSeen = moment().diff(moment(lastSeen), 'days')
-      const mediationExpiredThresholdDays = Number.parseInt(
-        Config.MEDIATION_EXPIRED_THRESHOLD_DAYS || DEFAULT_MEDIATION_EXPIRED_THRESHOLD_DAYS
-      )
+      const mediationExpiredThresholdConfig = Config.MEDIATION_EXPIRED_THRESHOLD_DAYS || DEFAULT_MEDIATION_EXPIRED_THRESHOLD_DAYS
+      let mediationExpiredThresholdDays = Number.parseInt(mediationExpiredThresholdConfig, 10)
+      if (Number.isNaN(mediationExpiredThresholdDays)) {
+        logger.warn(
+          `Invalid mediation expired threshold config value: ${mediationExpiredThresholdConfig}. Falling back to default of ${DEFAULT_MEDIATION_EXPIRED_THRESHOLD_DAYS} days.`
+        )
+        mediationExpiredThresholdDays = Number.parseInt(DEFAULT_MEDIATION_EXPIRED_THRESHOLD_DAYS, 10)
+      }
+
       if (daysSinceLastSeen < mediationExpiredThresholdDays) {
         logger.info(
           `Mediation connection last seen ${daysSinceLastSeen} days ago, which is below the expiration threshold. No need to reset mediation.`
         )
-        throw new Error('Mediation connection is not passed the expiration threshold, no need to reset mediation')
+        return
       }
       logger.info(
         `Mediation connection last seen ${daysSinceLastSeen} days ago, which is above the expiration threshold. Proceeding with mediation reset.`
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jleach** [comment]
@jamshale Thoughts on this going into the current production release or is it meant to go much later with the Single App phase 2 a.k.a v4.1 release?

**Comment by jamshale** [comment]
We should get it into both apps. It's important we have this feature for apps we control before we start deleting records from the mediator database. Having this and being connected to the mediator we control is critical to allow a recover process.

We do have to decide an expiry time to coordinate between the services, although with a change to the mediator we have planned, it will recover whether the mediation record gets deleted or not. This expiry time is just when it decides to try and recover on fail, on the wallet side.

File: app/src/hooks/useBCAgentSetup.ts [hunk 1]
```diff
@@ -175,6 +191,156 @@ const useBCAgentSetup = () => {
...
+        await agent.mediationRecipient.setDefaultMediator(newMediationRecord)
+        await agent.mediationRecipient.initiateMessagePickup(
+          newMediationRecord,
+          MediatorPickupStrategy.PickUpV2LiveMode
+        )
+      } catch (error) {
+        logger.error(`Error during mediation recovery. Attempting recovery of the deleted mediation records: ${error}`)
+        await mediationRepository.save(agent.context, oldMediationRecord)
+        await connectionRepository.save(agent.context, oldMediatorConnectionRecord)
+        if (oldOobRecord && !newConnectionEstablished) {
```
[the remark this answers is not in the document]
jamshale: This was done on purpose because the agent can not have 2 oob records with the same invitation. So if a new connection was established we need to leave the oob record deleted, but we restore it if the connection was never established.

app/src/hooks/useBCAgentSetup.ts
```diff
@@ -20,6 +34,9 @@ import { useCallback, useRef, useState } from 'react'
 import { Config } from 'react-native-config'
 import { CachesDirectoryPath } from 'react-native-fs'
 
+const DEFAULT_MEDIATION_EXPIRED_THRESHOLD_DAYS = '90'
```
jleach, line 37: Meant to be a a string?

app/src/hooks/useBCAgentSetup.ts
```diff
@@ -204,8 +389,12 @@ const useBCAgentSetup = () => {
       logger.info('Initializing agent...')
       await newAgent.initialize()
 
-      logger.info(`checking mediator type for ${mediatorUrl}`)
-      await newAgent.mediationRecipient.initiateMessagePickup(undefined, MediatorPickupStrategy.PickUpV2LiveMode)
+      try {
+        await newAgent.mediationRecipient.initiateMessagePickup(undefined, MediatorPickupStrategy.PickUpV2LiveMode)
+      } catch (error) {
+        logger.error(`Error initiating message pickup: ${error}`)
+        await recoverMediationIfExpired(newAgent, mediatorUrl)
```
jleach: When `recoverMediationIfExpired` returns early (not expired), the original pickup error gets swallowed and the agent continues without active message pickup. Worth re-throwing the original error in that case so the caller knows pickup isn't running.

Thoughts?

app/src/hooks/useBCAgentSetup.ts
```diff
@@ -175,6 +192,174 @@ const useBCAgentSetup = () => {
...
+        return
+      }
+
+      const mediationConnectionRecord = await connectionRepository.getById(
+        agent.context,
+        oldMediationRecord.connectionId
+      )
+      if (!mediationConnectionRecord) {
+        logger.warn('No connection record found for mediation record')
+        return
```
jleach: Nit: `mediationConnectionRecord` (line 228) and `oldMediatorConnectionRecord` (line 267) are the same fetch with the same ID — could reuse the first one instead of hitting the repo twice.

jleach:
> It's important we have this feature for apps we control before we start deleting records from the mediator database.

When would we start doing this? Askign because it'll matter if we do it before or after Phase 2 when we plan to retire BC Wallet 1.0.xx.

jamshale:
> > It's important we have this feature for apps we control before we start deleting records from the mediator database.
> 
> When would we start doing this? Askign because it'll matter if we do it before or after Phase 2 when we plan to retire BC Wallet 1.0.xx.

We were hoping to start the process of tagging records with last seen time as soon as possible but records won't actually be deleted from the database until they've passed the active expiry time. We haven't decided on an active expiry time yet. I think it should be between 3 and 6 months, meaning the old stale records won't be deleted for 3 to 6 months. 

If phase 2 is going to be complete before the decided active expire time then it doesn't need to go into bc wallet but only the new app.

app/src/hooks/useBCAgentSetup.ts
jamshale: correct. I passed in the original error and throw it for these paths now.

app/src/hooks/useBCAgentSetup.ts
jamshale: Yes, It's to match the corresponding Config.MEDIATION_EXPIRED_THRESHOLD_DAYS type as a string and handle the default the same way.

jleach:
> If phase 2 is going to be complete before the decided active expire time then it doesn't need to go into bc wallet but only the new app.

Ok. We're going to start a few tasks for Phase 2 a.k.a BCSC v4.1. When it's created we'll merge this change in there. It'll go into 4.1, probably in July(ish).

jleach (dismissed):

jleach:
@jamshale The checks look stuck on this PR. Able to push a no-op to make them trigger?
`git commit --allow-empty -s -m "chore: trigger ci"`

jamshale:
Ugh. Messed this up updating with main. Will fix...

jamshale:
Not sure why the actions aren't running. Maybe has something to do with the target branch. I can try and have a look.

jleach:
@jamshale I figured it out. The fix will be in #3708.

jleach (approved):

PR #3566, opened 2026-03-26
feat: e2e testing suites

KjartanE:
# Summary of Changes

Restructure E2E specs into WebdriverIO named suites, replacing runtime flow-mode conditionals with dedicated per-suite spec files.

- **`e2eConfig.ts` simplified** — removed `FlowMode`, `E2E_FLOW`, `FLOW_PRESETS`, and all preset-driven config fields. Only `VARIANT` selection remains.
- **`.env.e2e.example` cleaned up** — removed `E2E_FLOW` variable (no longer used).
- **Named suites added to `wdio.shared.conf.ts`** — `smoke`, `happy-path`, and `full-regression`, selectable via `--suite <name>`.
- **`happy-path` suite** — straight-through onboarding (PIN auth, no detours), combined-card verification with `TestUsers.basic` (nickname, card serial/birthdate entry), in-person method, tab/settings/account navigation.
- **`full-regression` suite** — onboarding with transfer detour, setup type interaction, and help detours; non-photo card verification with additional ID evidence (passport photo); in-person method; tab/settings/account navigation.
- **Card-type config pattern** — `verify/card-type/card-context.ts` holds shared mutable context (`testUser`, `cardTypeButton`); card-type config modules (`config-combined-card.ts`, `config-non-photo-card.ts`) set it at import time so downstream specs (`card-csn.spec.ts`, `in-person-verification.spec.ts`) read values lazily.
- **Old conditional specs removed** — `e2e.spec.ts` orchestrator and the shared `onboarding/`, `verify/`, `main/` specs with `if (config.xxx)` branching are deleted.
- **README updated** — documents the new suite structure, directory tree, migration from `E2E_FLOW` to `--suite`, and updated run commands.
- **CI workflow updated** (`main.yaml`) — "Determine test mode" step replaced with "Determine test suite" using `--suite smoke` for PRs; E2E runs on `main` are temporarily disabled. Old `--spec` / `e2e.spec.ts` references removed.

# Testing Instructions

1. Run the smoke suite (default): `yarn wdio configs/local/wdio.ios.local.sim.conf.ts --suite smoke`
2. Run the happy-path suite: `yarn wdio configs/local/wdio.ios.local.sim.conf.ts --suite happy-path`
3. Run the full-regression suite: `yarn wdio configs/local/wdio.ios.local.sim.conf.ts --suite full-regression`

# Acceptance Criteria

- [x] `wdio.shared.conf.ts` defines named `suites` for `smoke`, `happy-path`, and `full-regression`
- [x] Each suite has its own dedicated spec files with explicit test steps (no runtime conditionals)
- [x] `FlowMode`, `E2E_FLOW`, and `FLOW_PRESETS` are removed from `e2eConfig.ts` (and `.env.e2e.example`)
- [x] All `if (config.xxx)` conditionals are removed from spec files
- [x] The existing `test/bcsc/e2e.spec.ts` orchestrator is replaced by per-suite orchestrators
- [x] Suites can be run via `--suite smoke`, `--suite happy-path`, `--suite full-regression`
- [x] `VARIANT` env var continues to work for app-variant selection
- [x] Documentation is updated with the new suite structure and how to run each suite
- [x] CI workflow uses `--suite` flag and disables E2E on `main`

N/A

# Related Issues

N/A


e2e/test/bcsc/full-regression/verify/verify.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+    await EvidenceIDCollection.tap('Continue')
+  })
+})
+
+describe('In-Person Verification', () => {
+  it('should navigate through the Setup Steps screen and tap Step 5', async () => {
+    await SetupSteps.waitFor('Step5', 10_000)
+    await SetupSteps.tap('Step5')
+  })
+
```
MacDeluca: thought: Would this make sense as it's own "spec" file?

thought(2): If we keep each spec small and focused on a single flow or feature, full scenarios can be composed by importing the relevant pieces. This makes it easier to cover all permutations and mix and match specs to represent each valid path without duplicating test logic.

```ts
import './onboarding/onboarding.spec.js'
import './onboarding/biometrics-authentication.spec.js'
import './verify/verify.spec.js'
import './verify/combo-card-csn.spec.js'
import './verify/in-person-verification.spec.js'
import './main/main.spec.js'
...
```

KjartanE: This was my initial direction, the granularity of this setup is nice. But the commonality between the happy path and full regression is very small. So there isnt much reuse of these file. For clarity and readability it does help. Will spend some time on this to see if theres a happy medium.

MacDeluca: I think even the happy path has its own permutations and branches. So there's still value in the modular approach even at that level. 

I'm thinking combo, non-photo, non-bcsc etc but without stopping the app, backing out of flows, backgrounding etc that we would do in the "full-regression".

No action needed, just blorfing some thoughts out.

e2e/test/bcsc/verify/card-type/card-context.ts
```diff
@@ -0,0 +1,18 @@
...
+ * Config modules (e.g. config-combined-card.ts) set these at module evaluation time;
+ * downstream specs read them lazily inside `it` blocks.
+ */
+export const verifyContext: {
+  testUser: TestUser
+  cardTypeButton: CardTypeButton
+  cardTypeLabel: string
+} = {
+  testUser: null as unknown as TestUser,
+  cardTypeButton: null as unknown as CardTypeButton,
```
jleach: Nit: these `null as unknown` defaults will produce cryptic errors deep in a spec if a suite orchestrator forgets to import a config module. A lazy guard would make that failure obvious:

```ts
export function getVerifyContext() {
  if (!verifyContext.testUser) throw new Error('verifyContext not configured — import a config-*.ts module before this spec')
  return verifyContext
}
```

Or even just an `assert` at the top of specs that read it.

e2e/test/bcsc/verify/card-csn.spec.ts
```diff
@@ -0,0 +1,43 @@
+import { BaseScreen } from '../../../src/screens/BaseScreen.js'
+import { BCSC_TestIDs } from '../../../src/testIDs.js'
+import { verifyContext } from './card-type/card-context.js'
+
+const SetupSteps = new BaseScreen(BCSC_TestIDs.SetupSteps)
+const IdentitySelection = new BaseScreen(BCSC_TestIDs.IdentitySelection)
+const SerialInstructions = new BaseScreen(BCSC_TestIDs.SerialInstructions)
+const ManualSerial = new BaseScreen(BCSC_TestIDs.ManualSerial)
+const EnterBirthdate = new BaseScreen(BCSC_TestIDs.EnterBirthdate)
+
```
jleach, line 10: This `describe` label reads `verifyContext.cardTypeLabel` at module evaluation time — works today because the config module is imported first in the orchestrator, but if someone reorders imports it silently becomes `"BCSC  Card"`. Might be worth a comment in the orchestrator files noting that config imports must come before specs that reference `verifyContext` at the top level.

e2e/test/bcsc/verify/card-type/card-context.ts
KjartanE: Fair point, updated to utilize get function!

e2e/test/bcsc/verify/card-csn.spec.ts
KjartanE: true its not very clear, added comment.

jleach (approved):

PR #3571, opened 2026-03-27
fix: detect keyboard activity as user activity

### LINKED ISSUE #3554: Keyboard activity is not detected as activity
### Description of problem
Users can be locked out if they spend too much time editing text in a text input without any gestures outside of it. 

### Expected behavior
TextInput activity is valid activity and users should not be locked out in this case. 

### Steps to reproduce
Shorten lockout time to 1 minute, get to a text input like the birthdate field for example, don't tap outside of it, but use the keyboard semi-frequently *only* in the input, notice you get locked out

N/A

Build #:

Android Device Model:

iOS Device Model:

### Workaround
Tap outside the input

jleach:
## Summary

- Expose `reportActivity` on `BCSCActivityContext` and subscribe to `keyboardDidShow`/`keyboardDidHide` events to reset the inactivity timer when the keyboard opens or closes
- Call `reportActivity()` from `InputWithValidation` on every keystroke so typing resets the inactivity timer and prevents auto-lockout

Closes #3554

## Test plan

- set the apps minute to 1 minutes
- begin setup, get to a text input field (i.e passport number)
- enter 1 number every 20 seconds for 5 numbers.
- the app should not lock.

### DIFF
--- a/app/src/bcsc-theme/components/CodeInput.tsx
+++ b/app/src/bcsc-theme/components/CodeInput.tsx
@@ -1,3 +1,4 @@
+import { useBCSCActivity } from '@/bcsc-theme/contexts/BCSCActivityContext'
 import { PAIRING_CODE_LENGTH as CELL_COUNT } from '@/constants'
 import { ThemedText, useTheme } from '@bifold/core'
 import React, { Fragment, useCallback } from 'react'
@@ -19,13 +20,15 @@ interface CodeInputProps {
 
 const CodeInput = ({ value, onChange, error, onErrorClear, separator, textInputProps }: CodeInputProps) => {
   const { ColorPalette, Spacing, Inputs } = useTheme()
+  const { reportActivity } = useBCSCActivity() ?? {}
 
   const onChangeText = useCallback(
     (text: string) => {
+      reportActivity?.()
       onErrorClear?.()
       onChange(text)
     },
-    [onChange, onErrorClear]
+    [onChange, onErrorClear, reportActivity]
   )
 
   const [props, getCellOnLayoutHandler] = useClearByFocusCell({
--- a/app/src/bcsc-theme/components/InputWithValidation.tsx
+++ b/app/src/bcsc-theme/components/InputWithValidation.tsx
@@ -1,3 +1,4 @@
+import { useBCSCActivity } from '@/bcsc-theme/contexts/BCSCActivityContext'
 import { hitSlop } from '@/constants'
 import { a11yLabel } from '@/utils/accessibility'
 import { testIdWithKey, ThemedText, useTheme } from '@bifold/core'
@@ -46,6 +47,7 @@ type InputWithValidationProps = {
  */
 export const InputWithValidation: React.FC<InputWithValidationProps> = (props: InputWithValidationProps) => {
   const { Inputs, ColorPalette, Spacing } = useTheme()
+  const { reportActivity } = useBCSCActivity() ?? {}
   const inputRef = useRef<TextInput>(null)
   const [isFocused, setIsFocused] = useState(false)
 
@@ -119,6 +121,7 @@ export const InputWithValidation: React.FC<InputWithValidationProps> = (props: I
             style={[styles.input, props.inputProps]}
             value={props.value}
             onChangeText={(text) => {
+              reportActivity?.()
               if (props.error) {
                 props.onErrorClear?.()
               }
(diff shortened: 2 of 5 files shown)


MacDeluca:
Should we report activity on the other inputs? (Only a few others)

ie: 
CodeField: 82
Services.tsx: 105
PinInput.tsx: 81

app/src/bcsc-theme/contexts/BCSCActivityContext.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+    })
+
+    // Advance another 4 minutes — would have exceeded 5 min total without reset
+    act(() => {
+      jest.advanceTimersByTime(4 * 60 * 1000)
+    })
+
+    // Should still have a valid context (not logged out)
+    expect(result.current.reportActivity).toBeDefined()
+  })
```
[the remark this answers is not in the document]
jleach: Good call — updated the tests to assert on mockLogout. 👍

app/src/bcsc-theme/contexts/BCSCActivityContext.test.tsx
```diff
@@ -0,0 +1,103 @@
...
+      result.current.reportActivity()
+    })
+
+    // Resuming should work normally after
+    act(() => {
+      result.current.resumeActivityTracking()
+    })
+
+    expect(result.current.reportActivity).toBeDefined()
+  })
```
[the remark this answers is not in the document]
jleach: Same fix — tests now assert logout isn't called while paused. Thanks\!

jleach:
Good question.

- **PINInput** — lives in AuthStack, outside BCSCActivityProvider so the context isn't available there. No-op.
- **CodeField** — short pairing codes, pretty unlikely someone types long enough to hit the lockout. Also the keyboardDidShow/keyboardDidHide listeners we added will reset the timer when the keyboard opens, so there's a safety net.
- **Services.tsx** — same deal with the keyboard listeners covering it.

That said if you think CodeField or Services are worth wiring up explicitly, happy to add it — it's a small change. LMK.

MacDeluca:
> Good question.
> 
> * **PINInput** — lives in AuthStack, outside BCSCActivityProvider so the context isn't available there. No-op.
> * **CodeField** — short pairing codes, pretty unlikely someone types long enough to hit the lockout. Also the keyboardDidShow/keyboardDidHide listeners we added will reset the timer when the keyboard opens, so there's a safety net.
> * **Services.tsx** — same deal with the keyboard listeners covering it.
> 
> That said if you think CodeField or Services are worth wiring up explicitly, happy to add it — it's a small change. LMK.

The edge case is a user with a short auto-lock timeout who has the keyboard open while retrieving a pairing code or browsing the service list, so maybe it's worth adding direct activity tracking to CodeField and Services?


edit: The `keyboardDidShow` and `keyboardDidHide` benefits are probably pretty marginal (assuming user interacted right before and right after keyboard event). I think the `onChangeText` triggering the activity listener is the real benefit in this PR.

app/src/bcsc-theme/components/CodeInput.tsx
```diff
@@ -19,13 +20,15 @@ interface CodeInputProps {
 
 const CodeInput = ({ value, onChange, error, onErrorClear, separator, textInputProps }: CodeInputProps) => {
   const { ColorPalette, Spacing, Inputs } = useTheme()
+  const { reportActivity } = useBCSCActivity() ?? {}
```
MacDeluca, line 23: nit: We can remove the fallback value for the hook?

```ts
const { reportActivity } = useBCSCActivity()
```

app/src/bcsc-theme/components/InputWithValidation.tsx
```diff
@@ -46,6 +47,7 @@ type InputWithValidationProps = {
  */
 export const InputWithValidation: React.FC<InputWithValidationProps> = (props: InputWithValidationProps) => {
   const { Inputs, ColorPalette, Spacing } = useTheme()
+  const { reportActivity } = useBCSCActivity() ?? {}
```
MacDeluca, line 50: nit: same

app/src/bcsc-theme/features/services/Services.tsx
```diff
@@ -27,6 +28,7 @@ type ServicesNavigationProp = StackNavigationProp<BCSCMainStackParams, BCSCScree
  * @return {*} {React.ReactElement} The Services screen component.
  */
 const Services: React.FC = () => {
+  const { reportActivity } = useBCSCActivity() ?? {}
```
MacDeluca, line 31: nit: same

MacDeluca (approved):
Works on iOS and Android! 

Nice work 🚢

Issue #3502, opened 2026-03-19
Logging into a service from the service listing (in app) leaves the user stranded on the "You're done in this app" screen instead of return them to the home tab


gvodden:
### Description of problem
Testing logging into a service listing from within the app:
Logging into a service from the service listing (in app) leaves the user stranded on the "You're done in this app" screen instead of return them to the home tab.

The only option available to the user is a [Close] button which simply backgrounds the app, and when the app is foregrounded again, they are still on the same screen instead of returned to the Home tab

### Expected behavior
v3 behaviour is to return the user to the home tab after accessing the service.
(this process started in the app, and should leave the app functional)

### Steps to reproduce
open verified app
tap "Where to use"
Select a service (eg: BC Parks)
proceed through the login flow to the "you're done in this app" screen
tap Close
bring app to foreground and note which functions are available.

### Screenshots and/or log output
https://github.com/user-attachments/assets/86f46022-044b-48c4-86a7-7a883fa44f7b

### Environment: SIT 

Build #: 5447

Android Device Model: s24
iOS Device Model:

### Workaround - completely close app and re-open

cvarjao:
The "close" button after login needs to do some navigation or state management before switching to the original app

gvodden:
looking good in build 5761

PR #3574, opened 2026-03-30
fix: from app switch navigation on android

bryce-mcmath:
# Summary of Changes

This PR fixes an issue on Android where users would get stuck in a loop unable to leave the "You're done in this app" screen, simply by using the same automatic navigation that iOS uses.

# Testing Instructions

On Android, go through the deep link login process, ensure you don't get stuck on the aforementioned screen.

# Acceptance Criteria

Don't get stuck

N/A

# Related Issues

#3502

### DIFF
--- a/app/src/bcsc-theme/features/pairing/PairingConfirmation.test.tsx
+++ b/app/src/bcsc-theme/features/pairing/PairingConfirmation.test.tsx
@@ -153,6 +153,28 @@ describe('PairingConfirmation', () => {
       expect(queryByTestId('com.ariesbifold:id/Close')).toBeTruthy()
     })
 
+    it('navigates home when the app goes to background', () => {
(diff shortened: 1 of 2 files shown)


jleach:
On Android, tapping Close calls `exitApp()`, which should also fire a `background` state change — so both the exit and the nav reset would run. Any chance of a race there? Probably harmless since the app is closing anyway.

jleach (approved):

bryce-mcmath:
> On Android, tapping Close calls `exitApp()`, which should also fire a `background` state change — so both the exit and the nav reset would run. Any chance of a race there? Probably harmless since the app is closing anyway.

The app exits first and nothing else happens, it should be fine. App reopens again and works fine aftwards

PR #3595, opened 2026-03-30
feat: accessibility linter

al-rosenthal:
# Summary of Changes

- added new accessibility linter to project 
- added ruleset to linter config, it should flag any interactive component that's missing either a label or a flag to ignore as an error
- added accessibility hints to buttons that take the user out of the app
- added translations for hints that take the user out of the app
- updated snapshots

# Testing Instructions

# Acceptance Criteria

Lint checks on this PR pass

# Related Issues


MacDeluca:
thought: I wonder if this is better handled by linting?

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y

al-rosenthal:
> thought: I wonder if this is better handled by linting?
> 
> https://github.com/jsx-eslint/eslint-plugin-jsx-a11y

ooh great idea! I was also thinking we could add them to our jest tests as well but, linting seems like it could catch everything in a single shot instead of writing a bunch of tests

app/src/bcsc-theme/features/onboarding/AccountSetupScreen.tsx
```diff
@@ -71,6 +71,7 @@ const AccountSetupScreen = ({ navigation }: AccountSetupScreenProps) => {
       <Pressable
         onPress={incrementDeveloperMenuCounter}
         style={styles.pressableArea}
+        accessible={false}
```
al-rosenthal, line 74: we want this component to be ignored by screen readers, it's a secret

app/eslint.config.mjs
```diff
@@ -38,9 +39,21 @@ export default [
...
       },
     },
+    plugins: {
+      ...reactPlugin.configs.flat.recommended.plugins,
+      'react-native-a11y': reactNativeA11y,
+    },
     rules: {
       ...reactHooks.configs.recommended.rules,
       ...reactPlugin.configs.flat.recommended.rules,
+      'react-native-a11y/has-accessibility-props': 'error',
```
al-rosenthal, line 49: these rules check and throw errors if an interactive component is missing an accessibilityLabel, accessibilityHint or accessible props, not all of them together.

e2e/src/helpers/a11y.ts
```diff
@@ -0,0 +1,39 @@
+import assert from 'node:assert'
```
al-rosenthal: This can be removed, it's dead code now, but these helper functions can search for accessibility labels during an e2e test

e2e/eslint.config.mjs
```diff
@@ -0,0 +1,24 @@
+import pluginJs from '@eslint/js'
```
al-rosenthal: added this so the inline linter wasn't complaining about things like `driver` or the`$` seen in `BaseScreen`

package.json
```diff
@@ -74,6 +74,7 @@
     "react-native-bcsc-core": "workspace:*"
   },
   "devDependencies": {
+    "eslint-plugin-react-native-a11y": "^3.5.1",
```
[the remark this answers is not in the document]
al-rosenthal: moved

e2e/src/helpers/a11y.ts
MacDeluca: I would probably stash them and remove from this PR until needed

e2e/eslint.config.mjs
```diff
@@ -0,0 +1,24 @@
...
+import wdioPlugin from 'eslint-plugin-wdio'
+import tseslint from 'typescript-eslint'
```
KjartanE: imports `typescript-eslint`, which is not in `e2e/package.json`. Also `@eslint/js` is not installed within e2e dependencies. It does have `@typescript-eslint/eslint-plugin / parser` already

app/src/bcsc-theme/features/settings/SettingsContent.tsx
```diff
@@ -294,21 +294,25 @@ export const SettingsContent: React.FC<SettingsContentProps> = ({
             title={t('BCSC.Settings.Feedback')}
             onPress={onPressFeedback}
             testID={testIdWithKey('Feedback')}
+            accessibilityHint={t('Global.A11y.OpensInBrowser')}
```
KjartanE, line 297: There are 4 identical hints back to back. Maybe we make distinct hints for each unique button?

app/src/bcsc-theme/features/settings/components/SettingsActionCard.tsx
```diff
@@ -46,6 +47,7 @@ export const SettingsActionCard = (props: SettingsActionCardProps) => {
       onPress={props.onPress}
       accessibilityRole="button"
       accessibilityLabel={a11yLabel(props.title)}
+      accessibilityHint={props.accessibilityHint}
```
KjartanE, line 50: This `props.accessibilityHint` could be undefined, should we require so no button is left unhinted?

app/src/bcsc-theme/features/settings/SettingsContent.tsx
al-rosenthal: We could add the names? There are more accessibility props buried in the action card component so, maybe those should be surfaced as well. The hints themselves are just extra context, so with the screen reader enabled, when the user clicks the button it should read out "Feedback, this will open in the browser" so I'm not sure what else we could add, open to suggestions

app/src/bcsc-theme/features/settings/components/SettingsActionCard.tsx
al-rosenthal: Hints are mostly just for extra context, if a button is doing something that might be a little confusing to a user. The rest of these action cards just open a screen with more information so I don't think we'd gain much by forcing it on all of them.

e2e/eslint.config.mjs
al-rosenthal: updated the imports to use those already installed in e2e. This worked previously because of the workspaces are smart enough to look at app packages to resolve things

e2e/src/helpers/a11y.ts
al-rosenthal: removed

KjartanE (approved):
Tested locally, works as expected. good to go!

Issue #3603, opened 2026-03-31
303 error is triggered double-tapping buttons


RaulConstan:
### Description of problem
See the video clip for how to trigger it.

### Expected behavior
Double-clicking any buttons in the v4 app does not trigger an error

### Steps to reproduce
1. Start the login process
2. On confirmation screen tap the continue button 2 or more times in quick succession.

### Screenshots and/or log output

https://github.com/user-attachments/assets/04f2b0d4-b826-4343-a009-4b81d9767aed

### Environment
SIT

Build #:
5775

Android Device Model:
Samsung S24

iOS Device Model:
14 Plus

MacDeluca:
Options

- Order of operations (navigation first)
- Delay inbetween button presses (debounce?)
- Disable button when loading
- Loading state added to button

al-rosenthal:
available in 5937

RaulConstan:
Confirmed in 5937. (Note: there is also the unsupported version issue that triggers the 303, which is different that this one).

PR #3627, opened 2026-04-02
fix: button double tap

al-rosenthal:
# Summary of Changes

- added logic to Service login to ensure button doesn't allow double tap
- added logic around all account details
- added state logic in start call and before you call screens
- updated tests and snapshots

# Testing Instructions

- Convert project
- Verify an account
- Select a service to login into
- Select app as login option
- Hit the continue button as many times as you can before the screen navigates back
- Click the `All account details` button as many times as you can before it navigates

# Acceptance Criteria

No api errors when double tapping buttons

N/A

# Related Issues

[BC-Wallet-Mobile: 3603](https://github.com/bcgov/bc-wallet-mobile/issues/3603)


app/src/bcsc-theme/features/services/ServiceLoginScreen.tsx
```diff
@@ -183,93 +182,103 @@ const ServiceLoginDefaultView = ({
...
-        <View style={styles.continueButtonContainer}>
+        <View style={styles.buttonsContainer}>
+          <View style={styles.continueButtonContainer}>
+            <Button
+              title="Continue"
+              accessibilityLabel={a11yLabel('Continue')}
+              testID={testIdWithKey('ServiceLoginContinue')}
+              buttonType={ButtonType.Primary}
+              onPress={async () => {
+                const now = Date.now()
```
al-rosenthal: I tried this with just a useState control but when logging into a service, the underlying navigation returns before the screen transitions so it is still possible to hit the button again.

bryce-mcmath: Did we just need to await the function? I notice it wasn't awaited before. Sure we can't just use a boolean, even if it's a ref?

bryce-mcmath:
I think isLoading would be more appropriate than isDisabled for a lot of these. We can also add the ButtonLoading animated component in the button while it's true, not just disabling. That way the user knows to wait

app/src/bcsc-theme/features/services/ServiceLoginScreen.tsx
```diff
@@ -183,93 +183,104 @@ const ServiceLoginDefaultView = ({
...
+                // debounce added to give time for navigation before allowing another button press
+                if (now - lastPressRef.current < 1000) {
+                  return
+                }
+                lastPressRef.current = now
+                await onContinue()
```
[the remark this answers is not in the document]
al-rosenthal: this runs into the same issue as just the useState, the function completes before navigation occurs so the button becomes active before it should

app/src/bcsc-theme/features/services/ServiceLoginScreen.tsx
al-rosenthal: yeah that was the first thing I tested. `onContinueWithQuickLoginUrl` completes once the URL is passed to the OS, so with a useRef or useState, the function returns and marks is as active before we expect it to.

al-rosenthal:
> I think isLoading would be more appropriate than isDisabled for a lot of these. We can also add the ButtonLoading animated component in the button while it's true, not just disabling. That way the user knows to wait

app/src/bcsc-theme/features/account/Account.tsx
```diff
@@ -37,6 +37,7 @@ const Account: React.FC = () => {
   const getQuickLoginURL = useQuickLoginURL()
   const { account, refreshAccount } = useAccount()
   const { idToken, refreshIdToken } = useIdToken()
+  const [isDisabled, setIsDisabled] = useState(false)
```
bryce-mcmath, line 40: `accountDetailsIsLoading`?

bryce-mcmath: or something like that? Loading vs disabled is maybe a bit too nitpicky but there are other buttons on the page so a bit of specificity might be good

bryce-mcmath (approved):
Works, just one nit about the naming of that state var in Account.tsx

I'm good with merging this though

app/src/bcsc-theme/features/account/Account.tsx
al-rosenthal: I left this as disabled because adding the loading component looked very strange. These "buttons" are closer to tiles so our loading animation doesn't look right. Left it as disabled as there is no other references to loading

PR #3650, opened 2026-04-07
fix(registration): track build number and force update on v3→v4 upgrade

jleach:
## Summary

Device registration was not being updated on app upgrades (v3→v4) or build-only changes.

The v3 native app stored both version and build number in a `DeviceInfo` object and compared all fields on launch — any change triggered a registration update PUT. The v4 rewrite only compared `getVersion()` (marketing version) and initialized it to the current value on store creation, so build-only bumps were invisible and v3→v4 upgrades never saw a mismatch.

Fix: track both version and build number in state, initialize as empty so the first run always triggers an update, and compare both on subsequent launches.

- Added `appBuildNumber` to `BCSCState`, initialized both version fields to empty strings
- Updated `UpdateDeviceRegistrationSystemCheck` to compare both and treat blanks as forced update
- Tests cover build mismatch, blank values (v3→v4 upgrade path)

## Test plan

You'll see the PUT (below) in the logs now when you update from 3 -> 4 or when the build updates. You can find missing build numbers, or know ones from the support tools.

```console
[PUT] https://idsit.gov.bc.ca/device/register/d3dc4721-017d-4149-b720-c31f3c7cfee32 
```
- [x] Run repeated times, no change to build number, no update (PUT missing from logs).
- [x] Upgrade from v3 -> v4, registration update triggered (PUT in logs)
- [x] Run, increment build number, run again update triggered (PUT in logs).

### DIFF
--- a/app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.test.ts
+++ b/app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.test.ts
@@ -12,8 +12,7 @@ describe('UpdateDeviceRegistrationSystemCheck', () => {
     jest.useRealTimers()
   })
   describe('runCheck', () => {
-    it('should return true when app version is the same', async () => {
-      const mockAppVersion = '1.0.0'
+    it('should return true when app version and build number are the same', async () => {
       const mockUtils = {
         dispatch: jest.fn(),
         translation: jest.fn() as any,
@@ -21,18 +20,17 @@ describe('UpdateDeviceRegistrationSystemCheck', () => {
       }
       const updateRegistrationMock = jest.fn()
 
-      const mockGetVersion = jest.spyOn(deviceInfo, 'getVersion').mockReturnValue('1.0.0')
+      jest.spyOn(deviceInfo, 'getVersion').mockReturnValue('1.0.0')
+      jest.spyOn(deviceInfo, 'getBuildNumber').mockReturnValue('100')
 
-      const systemCheck = new UpdateDeviceRegistrationSystemCheck(mockAppVersion, updateRegistrationMock, mockUtils)
+      const systemCheck = new UpdateDeviceRegistrationSystemCheck('1.0.0', '100', updateRegistrationMock, mockUtils)
 
       const result = systemCheck.runCheck()
 
-      expect(mockGetVersion).toHaveBeenCalled()
       expect(result).toBe(true)
     })
 
     it('should return false when app version is different', async () => {
-      const mockAppVersion = '1.0.0'
       const mockUtils = {
         dispatch: jest.fn(),
         translation: jest.fn() as any,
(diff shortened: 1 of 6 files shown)


app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.ts
```diff
@@ -9,38 +9,41 @@ type UpdateRegistrationFunction = () => Promise<RegistrationResponseData>
...
-   * @returns {*} {boolean} - A boolean indicating if the device registration is up to date.
-   */
   runCheck() {
-    return this.lastAppVersion === getVersion()
+    // Force update if stored values are blank (v3→v4 upgrade or first run)
+    if (!this.lastAppVersion || !this.lastAppBuildNumber) {
+      return false
+    }
+
+    return this.lastAppVersion === getVersion() && this.lastAppBuildNumber === getBuildNumber()
```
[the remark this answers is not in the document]
jleach: Debug logging was removed in a follow-up commit. This line no longer exists.

app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.test.ts
```diff
@@ -12,63 +12,113 @@ describe('UpdateDeviceRegistrationSystemCheck', () => {
...
         logger: new MockLogger(),
       }
       const updateRegistrationMock = jest.fn()
 
-      const mockGetVersion = jest.spyOn(deviceInfo, 'getVersion').mockReturnValue('1.0.0')
+      jest.spyOn(deviceInfo, 'getVersion').mockReturnValue('1.0.0')
+      jest.spyOn(deviceInfo, 'getBuildNumber').mockReturnValue('100')
 
-      const systemCheck = new UpdateDeviceRegistrationSystemCheck(mockAppVersion, updateRegistrationMock, mockUtils)
+      const systemCheck = new UpdateDeviceRegistrationSystemCheck('1.0.0', '100', updateRegistrationMock, mockUtils)
```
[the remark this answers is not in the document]
jleach: Good catch — added jest.restoreAllMocks() in afterEach.

app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.ts
```diff
@@ -9,38 +9,43 @@ type UpdateRegistrationFunction = () => Promise<RegistrationResponseData>
...
-  /**
-   * Handles the failure case where the device registration needs to be updated.
-   *
-   * @returns {*} {Promise<void>} - A promise that resolves when the device registration update process is complete.
-   */
   async onFail() {
-    this.utils.logger.info('UpdateDeviceRegistrationSystemCheck: Updating device registration due to version change')
+    this.utils.logger.info(
+      'UpdateDeviceRegistrationSystemCheck: Updating device registration due to version/build change'
+    )
```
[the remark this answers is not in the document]
jleach: Fixed — onFail() now distinguishes between blank stored values (first run/upgrade) and an actual version/build mismatch.

app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.ts
```diff
@@ -1,6 +1,6 @@
 import { RegistrationResponseData } from '@/bcsc-theme/api/hooks/useRegistrationApi'
 import { BCDispatchAction } from '@/store'
-import { getVersion } from 'react-native-device-info'
+import { getBuildNumber, getVersion } from 'react-native-device-info'
 import { SystemCheckStrategy, SystemCheckUtils } from './system-checks'
```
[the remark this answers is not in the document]
jleach: The debug logging was added in an earlier commit and removed before this review. The PR description has been updated and no longer mentions it.

app/src/store.tsx
```diff
@@ -489,7 +491,7 @@ const bcReducer = (state: BCState, action: ReducerAction<BCDispatchAction>): BCS
       return newState
     }
     case BCSCDispatchAction.UPDATE_APP_VERSION: {
```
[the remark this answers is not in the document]
jleach: Fair point. Keeping the name as-is to avoid churn — not worth renaming across all dispatch sites for this. The reducer line is self-documenting.

app/src/store.test.ts
```diff
@@ -85,3 +87,13 @@ describe('getInitialEnvironment', () => {
...
+
+describe('reducer', () => {
+  it('UPDATE_APP_VERSION sets appVersion and appBuildNumber', () => {
+    const state = { ...initialState, bcsc: { ...initialState.bcsc, appVersion: '', appBuildNumber: '' } }
+    const result = reducer(state, { type: BCDispatchAction.UPDATE_APP_VERSION })
+
+    expect(result.bcsc.appVersion).toBe('4.0.0')
+    expect(result.bcsc.appBuildNumber).toBe('100')
+  })
+})
```
MacDeluca, line 99: praise: ❤️

app/src/store.tsx
```diff
@@ -381,7 +382,8 @@ const dismissPersonCredentialOfferState: DismissPersonCredentialOffer = {
 }
 
 export const initialBCSCState: BCSCState = {
-  appVersion: getVersion(),
+  appVersion: '',
+  appBuildNumber: '',
```
MacDeluca, line 386: suggestion: Null instead of empty strings?

app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.ts
```diff
@@ -9,38 +9,45 @@ type UpdateRegistrationFunction = () => Promise<RegistrationResponseData>
...
-   *
-   * @returns {*} {boolean} - A boolean indicating if the device registration is up to date.
-   */
   runCheck() {
-    return this.lastAppVersion === getVersion()
+    // Force update if stored values are blank (v3→v4 upgrade or first run)
+    if (!this.lastAppVersion || !this.lastAppBuildNumber) {
+      return false
+    }
+
```
MacDeluca, line 41: question: This block isn't needed, but I assume for clarity?

app/src/store.tsx
jleach: I do think null is a more better in typescript / js but maybe not in this case. I went with empty strings intentionally here. The falsy check in `runCheck()` (`!this.lastAppVersion || !this.lastAppBuildNumber`) catches both `''` (fresh init) and `undefined` (existing users hydrating without `appBuildNumber` in their persisted state). Using `null` would mean changing the type to `string | null` and adding null guards, but the behavior would be the same. Keeping it as plain `string` stays consistent with what `getVersion()` / `getBuildNumber()` return.

app/src/services/system-checks/UpdateDeviceRegistrationSystemCheck.ts
jleach: I think you're right — not needed. It does make the intent a little more clear though. I'm fine either way.

MacDeluca: Yea lets keep it, the comment is also helpful in understanding the flow.

app/src/store.tsx
MacDeluca: I like this reasoning ✅ 

Keeping the type a string makes sense, the value is populated immediately after init anyways.

MacDeluca (approved):

Issue #3605, opened 2026-03-31
Accessibility - Error Screen Color scheme not WCAG compatible


RaulConstan:
### Description of problem
The "Show details" text in yellow/orange on white background does not adhere to contrast standards...
See the screen shot.

### Steps to reproduce
1. Trigger an error in v4

### Screenshots and/or log output

<img width="461" height="1128" alt="Image" src="https://github.com/user-attachments/assets/0ff475d7-d177-4871-84a5-1fd44d244f34" />

### Environment
SIT

Build #:
5775

Android Device Model:
Samsung 24

iOS Device Model:
14 Plus

RaulConstan:
The following video clip is only for reference to show how the Alert/Error popup screen scales on Android with font size.

https://github.com/user-attachments/assets/32581a2f-52ee-46ed-93cb-68d1785a39e5

bryce-mcmath:
@jeznorth can we get a rough wireframe, mainly just for the colouring?

bryce-mcmath:
@RaulConstan here is a screenshot of the new error modal, you can see along with the colours being updated, there is also the option to hide the details once you have shown them — an extra lil improvement
<img width="585" height="1266" alt="Image" src="https://github.com/user-attachments/assets/f47c54cb-3061-4b73-a78c-0ea722586b5e" />

RaulConstan:
Thank you. Confirmed based on the screen shot above.

PR #3651, opened 2026-04-07
fix: error info card styles and accordian behaviour

bryce-mcmath:
# Summary of Changes

Changes colouring and accordian behaviour of ErrorInfoCard component, simplifies styling

# Testing Instructions

Throw a random error, see what it looks like. No functional changes

# Acceptance Criteria

No functional changes, colours are correct

This is what it looks like now:

https://github.com/user-attachments/assets/d9390b2e-e5d6-4304-a484-9b6149e2fb1f

Here are some quick wireframes Jer put together for this (it used to be yellow Show details text and the details could not be hidden once revealed):
Screenshot 2026-04-07 at 2 16 06 PM

# Related Issues

#3605

### DIFF
--- a/app/src/errors/components/ErrorInfoCard.tsx
+++ b/app/src/errors/components/ErrorInfoCard.tsx
@@ -102,6 +102,7 @@ export const ErrorInfoCard: React.FC<ErrorInfoCardProps> = ({
             <TouchableOpacity
               accessibilityLabel={showDetails ? t('Global.HideDetails') : t('Global.ShowDetails')}
               accessibilityRole="button"
+              accessibilityState={{ expanded: showDetails }}
               testID={testIdWithKey('ShowDetails')}
               style={styles.showDetailsTouchable}
               onPress={() => setShowDetails((prev) => !prev)}
--- a/app/src/errors/components/ErrorInfoCard.tsx
+++ b/app/src/errors/components/ErrorInfoCard.tsx
@@ -5,30 +5,31 @@ import React, { useMemo, useState } from 'react'
 import { useTranslation } from 'react-i18next'
 import { ScrollView, StyleSheet, Text, TouchableOpacity, useWindowDimensions, View } from 'react-native'
 import { getBuildNumber, getVersion } from 'react-native-device-info'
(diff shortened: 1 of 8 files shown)


app/src/errors/components/ErrorInfoCard.tsx
```diff
@@ -5,31 +5,32 @@ import React, { useMemo, useState } from 'react'
...
 import { ErrorModalAction } from './ErrorModal'
 
-export const fallbackColors = {
+// This component is used outside the theme provider in many cases and has style requirements
+// that the current theme doesn't support well, so we define the colours here
+export const colors = {
   cardBackground: '#FFFFFF',
   cardBorder: '#D3D3D3',
   shadow: '#000000',
   icon: '#313132',
```
MacDeluca, line 18: question: Should we reference the colour palette and theme directly here?

app/src/errors/components/ErrorInfoCard.tsx
```diff
@@ -166,139 +175,133 @@ export const ErrorInfoCard: React.FC<ErrorInfoCardProps> = ({
...
-      paddingTop: 8,
-      alignSelf: 'center',
-      fontSize: 12,
-      color: colors.textSecondary,
-    },
-  })
+const styles = StyleSheet.create({
+  card: {
+    backgroundColor: colors.cardBackground,
+    borderRadius: 8,
```
MacDeluca, line 181: And here?

app/src/errors/components/ErrorInfoCard.tsx
bryce-mcmath: I feel you. Because it's kind of a weird component that the Bifold theme structure doesn't match to, the colours don't match up exactly with either theme, BC Wallet or BCSC. It's a bit of a mix and match. I can use both direct imports of both theme's ColorPalettes but that seems counterproductive and not the best use of time

MacDeluca:
@bryce-mcmath Can you attach screenshots of errors with actions? (default & destructive styling)

bryce-mcmath:
@MacDeluca the video has the default button, here's destructive:
kablow

MacDeluca (approved):
Works on both iOS and Android!

PR #3665, opened 2026-04-09
feat: e2e cidc upgrade

KjartanE:
# Summary of Changes

Configure SauceLabs infrastructure so E2E tests run reliably in CI/CD pipelines.

### PR smoke tests (active)

- Every PR triggers a `smoke` suite on 1 iOS + 1 Android device via the reusable `e2e.yml` workflow (called from `main.yaml`).

### Nightly full regression + biometrics (blocked on IP whitelisting)

- Full regression is **not** run inline on merge to main — this keeps SauceLabs devices free during the day when multiple PRs merge.
- Instead, `e2e-nightly.yml` runs at midnight PT and checks whether any commits landed on main in the last 24 hours. If merges happened, it runs `full-regression` on 3 iOS (18/17/16) + 3 Android (15/14/13) OS versions, plus non-blocking biometric tests. If nothing merged, it skips entirely.
- Manual dispatches via the Actions UI always run regardless of the merge check.
- The nightly schedule is currently disabled pending IP whitelisting.

### Verified-state suite (scaffolded)

- A new `verified` suite runs specs sequentially with app state **persisted across sessions** (`noReset=true` / `fullReset=false`). The first spec runs the full onboarding + verification flow to reach the verified home screen; subsequent specs skip re-verification entirely.
- Dedicated config files (`configs/sauce/verified/wdio.{ios,android}.verified.sauce.rdc.conf.ts`) mirror the biometrics pattern — capability overrides are isolated, keeping the shared configs clean. `maxInstances: 1` is enforced at the config level to guarantee sequential execution.
- Wired into `e2e.yml` as a non-blocking `e2e-verified` job (like biometrics), and available as a `workflow_dispatch` option. The suite entry in `wdio.shared.conf.ts` is registered and ready; post-verified specs are added by importing under the Phase 2 comment in `verified.spec.ts`.

### Reusable E2E workflow

- `e2e.yml` accepts a `device_matrix` JSON array, `suite` name, and `build_number`. Each matrix entry spawns a separate SauceLabs RDC session (max-parallel: 2). Supports `smoke`, `happy-path`, `full-regression`, `biometrics`, and `verified` suites.

### Still in progress

- GitHub Actions runner IP whitelisting with BC Gov ID Check. Once complete, the nightly schedule can be enabled.

# Testing Instructions

1. Open a PR against `main` — verify the `e2e` / smoke job triggers after `ship-to-saucelabs` completes and runs on 1 iOS + 1 Android device.
2. Manually dispatch `e2e.yml` from the Actions UI with a known `build_number` and confirm tests execute on SauceLabs.
3. Manually dispatch `e2e-nightly.yml` — verify it resolves the latest main build number and runs the full device matrix (manual dispatches bypass the merge check).
4. Once IPs are whitelisted: uncomment the nightly schedule in `e2e-nightly.yml` and confirm it fires at midnight PT, only running when commits landed on main that day.
5. Manually dispatch `e2e.yml` with `suite: verified` — confirm the `e2e-verified` job runs on both iOS and Android, that phase 1 completes verification, and that any phase 2 specs start from the verified home screen without re-onboarding.

# Acceptance Criteria

- [x] CI/CD pipeline runs E2E smoke tests on at least 1 iOS and 1 Android device per PR
- [ ] GitHub Actions runner IPs are whitelisted for the verification process
- [ ] Nightly runs E2E full-regression on multiple devices and OS versions when merges land on main (uncomment schedule after whitelisting)
- [x] Pipeline configuration is documented (`e2e/README.md`, workflow comments, `.env.saucelabs.example`)
- [x] `verified` suite scaffolded — config files, suite aggregator, and CI job ready; post-verified specs added as they are written

N/A

# Related Issues

N/A

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -31,7 +31,7 @@ on:
   workflow_call:
     inputs:
       suite:
-        description: 'Test suite (smoke, happy-path, full-regression, biometrics)'
+        description: 'Test suite (smoke, happy-path, full-regression, verified, biometrics)'
         type: string
         required: true
       build_number:
@@ -60,6 +60,7 @@ on:
           - smoke
           - happy-path
           - full-regression
+          - verified
           - biometrics
         default: smoke
       build_number:
@@ -74,7 +75,7 @@ on:
 jobs:
   e2e-tests:
     name: '${{ inputs.suite }} - ${{ matrix.device.platform }} ${{ matrix.device.os_version }}'
-    if: inputs.suite != 'biometrics'
+    if: inputs.suite != 'biometrics' && inputs.suite != 'verified'
     runs-on: ubuntu-22.04
     permissions:
       contents: read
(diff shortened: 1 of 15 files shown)


KjartanE:
Happy to merge and enable nightly job and full-regression tests on another PR when whitelist IPs is completed

.github/workflows/e2e.yml
```diff
@@ -169,24 +117,43 @@ jobs:
...
         run: yarn install --immutable
 
-      - name: Run biometric tests
+      - name: Run WDIO tests
         working-directory: e2e
         env:
           SAUCE_USERNAME: ${{ secrets.SAUCE_USERNAME }}
           SAUCE_ACCESS_KEY: ${{ secrets.SAUCE_ACCESS_KEY }}
           SAUCE_REGION: us
           VARIANT: bcsc-dev
```
jleach, line 127: Noticed this is now hardcoded to `bcsc-dev` — the old workflow pulled it from the matrix. The README CI table still mentions "PR variants (2)" and "All 5 variants" which looks stale now. No biggie if variant support was intentionally dropped, just worth syncing the docs.

KjartanE: removed variants as each are the pretty much identical. All e2e tests should run the exact same no matter the variant so I figured to reduce complexity we will just run tests against `dev`. Will update the readme accordingly

jleach (approved):

Issue #3648, opened 2026-04-07
"Invalid QR Code" error when transferring from v3 to v4


RaulConstan:
### Description of problem
Transferring from iOS v3.15.1 to Android v4 (5891) seems to be completely blocked.

### Expected behavior
As per v3

### Steps to reproduce
1. Get QR code up on iOS
2. Scan QR code on Android

### Screenshots and/or log output

https://github.com/user-attachments/assets/2eec6ceb-d700-4fd9-ab4d-7e1d230581e9

### Environment
SIT

Build #:
5891

Android Device Model:
S21

iOS Device Model:
14 Plus

bryce-mcmath:
6100

RaulConstan:
On 6122, something calls the api to END the v3.x instead of giving the option to keep or remove...:
Some additional background and steps.
Background:
I was testing the QR transfer from v4 to v4 (Samsung SM-G990W source to iPhone target).
That worked
I was then testing the QR transfer from v3 to v4 (Pixel 6 source to iPhone target)
That worked - sort of.
Steps:
1. Had RC0000080 loaded onto the Pixel 6
2. Transferred via QR code to iPhone (see video)
3. BUT, I kept going and removed the account from the iPhone to repeat the QR transfer from Pixel 6 again to iPhone.
4. After I removed the account on iPhone I went back to the Pixel to get the QR code screen and found the account had been removed (shown in the video following the black screen & authentication)
The big question:
What triggered the removal of the account on the Pixel 6?  Was it step 3 when I removed the account from the iPhone?  Did the API to remove the iPhone mobile card get sent for the rest of the mobile cards?  In a video call with Jason, we saw live that the Samsung phone ALSO got removed - see the screen shot after the video.

https://github.com/user-attachments/assets/910e9465-2248-461b-ba56-0dca36ecf0a5

<img width="1505" height="1086" alt="Image" src="https://github.com/user-attachments/assets/bd253d06-8c41-4630-b01a-e5598f9bd02f" />

jleach:
@RaulConstan I traced the API calls. A device can only remove its own account. Our API call is:

  **Example**:
  DELETE https://idsit.gov.bc.ca/device/register/{clientId}
  Authorization: Bearer {registrationAccessToken}

  **Concrete example**:
  DELETE https://idsit.gov.bc.ca/device/register/5f3c2a1b-9e4d-4a7c-8b12-abcdef012345
[...]

Issue #3647, opened 2026-04-07
"Invalid QR Code" error when transferring from Android to iOS


RaulConstan:
### Description of problem
Transferring account using QR code from Android to iOS mobiles in v4 results in error.  (but the app is usable???)

Not sure if it's the identity: RC0000001 Jan 1, 2000 (expired, but loaded via emergency mode), or something else, but I am able to reproduce it almost always, see the video clip.

### Expected behavior
per v3

### Steps to reproduce
1. On Android, get the QR code to display
2. On iOS, start the Transfer process (from Create new/Transfer screen)
3. Follow the process until you scan the QR code
4. Scan the QR code

### Screenshots and/or log output
Took a couple of tries (it's semi intermittent):

https://github.com/user-attachments/assets/fbda22ca-57b9-4da3-b8b5-f49da8e63d01

Loki logs:
[Explore-logs-2026-04-07 11_21_17.txt](https://github.com/user-attachments/files/26546512/Explore-logs-2026-04-07.11_21_17.txt)

### Environment
SIT

Build #:
5937

Android Device Model:
S21 (Source) with v4 and RC000000 Jan 1, 2000 (Expired card that was loaded via emergency mode in Support Tools)

iOS Device Model:
14 Plus

bryce-mcmath:
6100

RaulConstan:
Fails on 6122 (both Android and iOS BCSC v4 are fresh installs):

https://github.com/user-attachments/assets/56b02bbf-fd4f-4f77-be18-85e0b5145df6

RaulConstan:
Note, still getting the error, but may be related to the IAS issue seen in the SIT R4.0 release.  QA and TEST work as expected.

PR #3681, opened 2026-04-14
fix: account transfer bugs

bryce-mcmath:
# Summary of Changes

This PR:
- expands attestation check success handling range to any 2xx (sometimes gets 201's)
- logs attestation check 400's but doesn't bubble them into QR scanner errors
- adds content gradient on lengthy instructions screen to indicate more content is available and scrollable
- uses proper ButtonLoading rather than ActivityIndicator so there is space between indicator and text content

# Testing Instructions

Do account transfer

# Acceptance Criteria

Works, no unexpected errors

N/A

# Related Issues

#3647 
#3648


KjartanE (approved):
Tested locally on both devices, works as intended

PR #3710, opened 2026-04-16
feat: login by pairing code

KjartanE:
# Summary of Changes

Add e2e coverage for the Log-In-From-Computer flow.

- `e2e/src/helpers/pairing-code.ts` — Node-side helper that mints a real pairing code by replaying the BC Parks demo web flow at `idsit.gov.bc.ca` (5 HTTP calls, no external deps)
- `e2e/test/bcsc/main/login-from-computer.spec.ts` — spec that drives Home → ManualPairingCode, types the minted code, and verifies PairingConfirmation renders and closes back to Home

Skips driving an on-device browser entirely, so the same path runs on iOS and Android without context-switching or parallel sessions.

# Testing Instructions

Preconditions: app is onboarded and resting on Home (e.g. run with `NO_RESET=true` after a full-regression pass).

```sh
cd e2e && yarn test:android:local --spec test/bcsc/main/login-from-computer.spec.ts
cd e2e && yarn test:ios:local     --spec test/bcsc/main/login-from-computer.spec.ts
```

# Acceptance Criteria

- Home → LogInFromComputer navigates to ManualPairingCode
- Pairing code minted from `idsit.gov.bc.ca` submits successfully, landing on PairingConfirmation
- Close returns the user to Home

N/A

# Related Issues

N/A

### DIFF
--- a/e2e/src/helpers/pairing-code.ts
+++ b/e2e/src/helpers/pairing-code.ts
@@ -174,7 +174,8 @@ async function readBodyOrThrow(step: string, response: Response): Promise<string
   const body = await response.text()
   if (!response.ok) {
     const snippet = body.slice(0, 200).replaceAll(/\s+/g, ' ').trim()
-    throw new Error(`${step} failed: HTTP ${response.status}${snippet ? ` — ${snippet}` : ''}`)
+    const detail = snippet ? ` — ${snippet}` : ''
+    throw new Error(`${step} failed: HTTP ${response.status}${detail}`)
   }
   return body
 }
--- a/e2e/src/helpers/pairing-code.ts
+++ b/e2e/src/helpers/pairing-code.ts
@@ -3,30 +3,39 @@
  * Node. Used by the login-from-computer spec so tests don't have to drive
  * an on-device browser.
  *
- * The flow targets the SIT (System Integration Testing) environment —
- * idsit.gov.bc.ca — and impersonates the "BC Parks Discover Camping"
- * demo relying party (clientId `urn:ca:bc:gov:demo:rp1`).
+ * Pinned to SIT (idsit.gov.bc.ca), same as every other automated BCSC
+ * helper in this repo — the test users in e2e/assets/USERS.md only exist
+ * there. `assertSitEnv` reads app/.env and throws early if the app under
+ * test is built for any other environment, so the failure shows up before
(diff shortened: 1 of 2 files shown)


e2e/src/helpers/pairing-code.ts
```diff
@@ -0,0 +1,193 @@
...
+ * After step 5 the backend holds the code in `state: "pairing"` for a few
+ * minutes, waiting for the mobile app to submit it via the normal pairing
+ * API. The caller is expected to immediately type the returned code into
+ * the app.
+ */
+
+const SIT_BASE = 'https://idsit.gov.bc.ca'
+const DEMO_RP_LOGIN = `${SIT_BASE}/demo/rp1/login`
+const SAML2_POST = `${SIT_BASE}/login/saml2`
+const LOGIN_ENTRY = `${SIT_BASE}/login/entry`
```
jleach: nit: this pins to SIT but the app supports PROD/SIT/QA/TEST/DEV/DEV2/PREPROD via `DEFAULT_ENVIRONMENT`. if the device under test is pointed anywhere other than SIT the minted code won't be valid. worth keying off the same env var, or at least asserting app env == SIT before running so it fails loudly instead of mysteriously.

e2e/src/helpers/pairing-code.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  // 3. Follow redirect to /login/entry. Server embeds the cardtap transaction
+  //    UUID as a JS literal — scrape it.
+  const entryRes = await fetch(LOGIN_ENTRY, {
+    redirect: 'manual',
+    headers: {
+      ...baseHeaders(jar),
+      Referer: DEMO_RP_LOGIN,
+    },
```
jleach: nit: regex depends on exact formatting (`var transactionID = "..."`). if the server ever swaps to single quotes or `let`/`const` the test breaks with a vague error. `cheerio` is already a devDependency in e2e/package.json — could use it for the input scrapes above, and loosen this one to tolerate quote/whitespace variation cheaply.

e2e/src/helpers/pairing-code.ts
KjartanE: added an assertion for SIT env

jleach (approved):

Issue #3720, opened 2026-04-20
Initialize wallet agent


MacDeluca:
### Description of feature / user story
- The BC Wallet agent needs to be initialized as a pre-req for BCWallet functionality.

### Thoughts
- Should this be non-blocking? If agent initialization fails what should the user see and have access to?
- Possibly initialized in AuthStack?

PR #3724, opened 2026-04-21
feat: scaffold BCSC wallet agent service

jleach:
Part of #3720. Lays the pure-function service (`buildAgent`, `restartAgent`, `warmCache`, `shutdownAgent`) for BCSC-side Credo agent initialization in v4.1, where BCSC becomes the base shell and needs its own clean init path. The legacy `useBCAgentSetup` + `Splash` path is untouched — it still serves the BCWallet mode.

This is PR 1 of a stacked set. PR 2 will add the ViewModel, provider, and RootStack wiring to activate the service.


app/src/bcsc-theme/features/agent/services/agent-service.ts
MacDeluca, line 1: nit: tsdoc on these functions would be helpful

jleach: Done.

bryce-mcmath (approved):

PR #3725, opened 2026-04-21
feat: wire BCSC agent init via ViewModel + provider

jleach:
Part of #3720 (PR 2 of 2, stacked on #3724). Activates the agent service from PR1 inside the BCSC shell via `useAgentSetupViewModel` + `BCSCAgentProvider`, mounted in `BCSCRootStack` between the `!didAuthenticate` and `!verified` branches so VerifyStack, MainStack, and any future credential-consuming feature run inside the live agent context.

On failure, the provider calls `emitErrorModal` (once per error transition) with 2901/2902 strings and a Retry action; the LoadingScreen stays mounted underneath. `isAppError`/`instanceof AppError` is used to preserve typed errors thrown by the service and wrap anything else as 2901 with the original as `cause`.

### DIFF
--- a/app/src/bcsc-theme/features/agent/BCSCAgentProvider.tsx
+++ b/app/src/bcsc-theme/features/agent/BCSCAgentProvider.tsx
@@ -1,41 +1,24 @@
-import { useErrorAlert } from '@/contexts/ErrorAlertContext'
 import { AgentProvider } from '@bifold/core'
-import React, { PropsWithChildren, useEffect, useRef } from 'react'
+import React, { PropsWithChildren } from 'react'
 import { useTranslation } from 'react-i18next'
 
 import { LoadingScreen } from '../../contexts/BCSCLoadingContext'
 
 import useAgentSetupViewModel from './useAgentSetupViewModel'
 
+// Non-blocking on error: init failures are logged by the ViewModel but we do
+// not surface a modal — children render without an active agent context, and
+// downstream credential features will surface their own errors when invoked.
 const BCSCAgentProvider: React.FC<PropsWithChildren> = ({ children }) => {
   const { t } = useTranslation()
-  const { emitErrorModal } = useErrorAlert()
-  const { agent, status, error, retry } = useAgentSetupViewModel()
-  const lastErrorCodeRef = useRef<number | null>(null)
-
-  useEffect(() => {
-    if (status !== 'error' || !error) {
-      lastErrorCodeRef.current = null
-      return
-    }
-
-    if (lastErrorCodeRef.current === error.statusCode) {
-      return
-    }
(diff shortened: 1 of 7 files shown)


app/src/bcsc-theme/navigators/RootStack.tsx
```diff
@@ -65,15 +66,27 @@ const BCSCRootStack: React.FC = () => {
...
+  if (verified === false) {
     return (
       <BCSCActivityProvider>
         <VerifyStack />
       </BCSCActivityProvider>
     )
   }
 
-  if (store.bcscSecure.verified === true) {
+  if (verified === true) {
```
MacDeluca: thought: These changes are going to conflict with this https://github.com/bcgov/bc-wallet-mobile/pull/3732/changes#diff-cd4337025e4f3a3957a7b764c1f702b95748857e080ef89463c434ba7699d96f

suggestion: The VerifyStack doesn't require the Agent and therefore we don't need to wrap with BCSCAgentProvider. We can probably just wrap the MainStack for now?

jleach: Conflict with #3732: rebased PR1 onto `release/v4.1.0` (which now includes #3732), then re-stacked PR2 on top. Resolved the `RootStack.tsx` conflict to keep the new `verifiedStatus === IN_PROGRESS` condition.

 Wrapping only MainStack: agreed, VerifyStack doesn't need the agent yet — moved BCSCAgentProvider down to wrap just BCSCMainStack. Also let me drop the AuthenticatedStack sub-component, so the diff vs release is now basically one line.

app/src/bcsc-theme/features/agent/BCSCAgentProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+    lastErrorCodeRef.current = error.statusCode
+
+    const title = t(`Error.Title${error.statusCode}`)
+    const description = t(`Error.Message${error.statusCode}`)
+    emitErrorModal(title, description, error, {
+      action: {
+        text: t('Init.Retry'),
+        onPress: retry,
+      },
+    })
```
bryce-mcmath: Should agent errors put up a modal to users or should they just log? The user-facing error might be kind of a divergence from the non-blocking approach

bryce-mcmath (changes requested):
I think maybe non-blocking means avoiding fullscreen errors unless necessary — users can still see if the wallet features are available on relevant tabs now that the status is available

app/src/bcsc-theme/features/agent/BCSCAgentProvider.tsx
jleach: Done.

bryce-mcmath (approved):

PR #3726, opened 2026-04-21
feat: non-bcsc e2e 

MacDeluca:
# Summary of Changes

  - New e2e suite for Non-BCSC card verification flow — covers the full path: onboarding → dual ID collection (driver's licence + passport) → residential address → email confirmation with disposable email via Guerrilla Mail
  - Fixed DropdownWithValidation modal — replaced inner Pressable with accessible={false} with a SafeAreaView, fixing iOS accessibility tree being collapsed and hiding all dropdown options from Appium
  - Added email helper (e2e/src/helpers/email.ts) — creates temp Guerrilla Mail addresses, polls for incoming emails, extracts 6-digit confirmation codes from HTML bodies
  - Test config updates — fail-fast on first failure (bail: 1 + mocha bail: true), added firstName/lastName to test users, fixed testID values for evidence form inputs (*Pressable → *Input), added province dropdown option ID
  - Added testID to SetupStep subtext for verifying dynamic content like confirmed email addresses

# Testing Instructions
- Make sure to create new builds for Android and iOS and place in /apps dir.
```bash
yarn test:ios:device --spec test/bcsc/full-regression/verify-non-bcsc-card.spec.ts
yarn test:android:device --spec test/bcsc/full-regression/verify-non-bcsc-card.spec.ts
```

# Acceptance Criteria

Test suite passes on iOS and Android and email address automatically confirmed.

# Related Issues


e2e/src/testIDs.ts
```diff
@@ -120,11 +120,14 @@ export const BCSC_TestIDs = {
   EvidenceIDCollection: {
     Continue: 'com.ariesbifold:id/EvidenceIDCollectionContinue',
     Cancel: 'com.ariesbifold:id/EvidenceIDCollectionCancel',
+    DriversLicenseEvidenceOption: `com.ariesbifold:id/EvidenceTypeListItem B.C. driver's licence`,
+    PassportEvidenceOption: `com.ariesbifold:id/EvidenceTypeListItem Canadian Passport`,
```
jleach, line 124: These testIDs couple the e2e suite to i18n strings (`B.C. driver's licence`, `Canadian Passport`) and embed spaces + an apostrophe into the resource ID. Two concerns:

1. **Breaks on copy changes** — any label tweak in translations will silently break these specs.
2. **Fragile as Android `resourceId`** — spaces/apostrophes in resource IDs aren't universally supported by UiAutomator selectors depending on how the source builds them.

Suggest adding stable, label-free keys upstream (e.g. `EvidenceTypeListItem-DriversLicence`, `EvidenceTypeListItem-Passport`) in the component that renders these options, and referencing those here instead.

MacDeluca: These testID's are driven by the API response, ie: 'Canadian Passport' is returned from the API and injected into the testID function -> `com.ariesbifold:id/EvidenceTypeListItem Canadian Passport`.

Is there an alternative way of doing this with dynamic values?

app/src/bcsc-theme/components/DropdownWithValidation.tsx
```diff
@@ -193,42 +199,28 @@ export const DropdownWithValidation = <T extends string | number>({
       ) : null}
 
       <Modal visible={isOpen} transparent animationType="slide" onRequestClose={handleClose}>
-        <Pressable
-          style={styles.modalOverlay}
```
jleach, line 197: Small nit: now that this `Pressable` overlay is gone, `styles.modalOverlay` (line 74 on main) is unused — safe to drop whenever.

e2e/src/testIDs.ts
jleach: I see. We'd probably need to remove whitespaces or something. Maybe ok for now.

jleach:
Looks like tap-outside-to-close is gone, but when I switch between your branch and main I don't really see any option to tap outside the sheet either way. Was it ever actually used? Testing with the province dropdown on Step 3.

MacDeluca:
> Looks like tap-outside-to-close is gone, but when I switch between your branch and main I don't really see any option to tap outside the sheet either way. Was it ever actually used? Testing with the province dropdown on Step 3.

Yea I didn't quite understand what was going on. The modal appears to be full screen so tap outside would only include the header. I just simplified it to only close on press of the button OR pressing a province in the list.

I did update the styling slightly to give some visual feedback when the user presses an item in the list.

note: Only reason I touched it was because the e2e test suite was having difficulty pressing the items in the list with all the nested Pressable elements.

jleach (approved):
LGTM. Tested changes to the modal on Step 3 (Province). Still working as expected.

Issue #3794, opened 2026-05-04
Potential missing nickname issue


bryce-mcmath:
### Description

On iOS, the prod BCSC app installed in 2023 and used infrequently since, might fail to populate the nickname when upgraded to v4. It's hard to know for certain as the situation that produced this result was a bit odd — we accidentally installed a BC Wallet build overtop first, then tried to recreate the same situation with a freshly installed v3 but we weren't able to. Worth installing an early 2023 version of the app, verifying by send video, then upgrading to the latest main of v4.

PR #3795, opened 2026-05-05
fix: use displayName as fallback for missing nickname from earlier v3

bryce-mcmath:
# Summary of Changes

Two changes:
- use displayName as fallback nickname in initial account check
- change button to "Continue" rather than "Continue setup" so if someone verified manages to migrate without a displayName or nickname, they don't mistakenly think they need to re-verify

# Testing Instructions
In ias-ios:
```zsh
git checkout 3.5.1_489
# install, get verified
git checkout 3.12.0_618
# install, unlock app
# then build this branch overtop, confirm nickname gets set to displayName (will be given name or family name)
```

Then, do a fresh install of this branch and create an account but don't yet create a nickname. Cycle the app and check that the button shown says "Continue" not "Continue setup"

# Acceptance Criteria

Fallback works, button label changes

N/A

# Related Issues

#3794

### DIFF
--- a/app/src/bcsc-theme/api/hooks/useInitializeAccountStatus.test.tsx
+++ b/app/src/bcsc-theme/api/hooks/useInitializeAccountStatus.test.tsx
@@ -167,9 +167,9 @@ describe('useInitializeAccountStatus', () => {
     expect(mockDispatch).not.toHaveBeenCalledWith(expect.objectContaining({ type: BCDispatchAction.ADD_NICKNAME }))
   })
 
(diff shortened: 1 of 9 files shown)


jleach:
Nice-to-have, not blocking: `Global.ContinueSetup` is now orphaned in the locale files (`en`, `fr`, `pt-br`) after the rename. Fine to clean up here or punt to a later change.

jleach (approved):
Tested on iOS. Worked as expected. Last name used as nickname for fallback.

Issue #3796, opened 2026-05-05
QR process fails Removal of Source


RaulConstan:
### Description of problem
The Remove process repeats on the Source phone.

NOTE: the process is repeatable iOS to Android and Android to iOS.

### Expected behavior
Per v3

### Steps to reproduce
1. Source = iOS 26.3.1 BCSC v4.0.0 build 6365
2. Target = Android 16 BCSC v4.0.0 build 6365
See video

### Screenshots and/or log output

https://github.com/user-attachments/assets/1d49de84-7567-43b1-aeca-042c091556e6

### Environment
SIT

Build #:
6365

Android Device Model:

iOS Device Model:

gvodden:
I just recreated this on my iphone 11.

It wasn't clear to me from the ticket that the screen was backing up a step on its own, but my iphone 11 just did that.
I tapped the first "remove account" button and the screen proceeded to the confirmation buttons, and then in about 2 seconds, it automatically backed up a screen.

gvodden:
https://github.com/user-attachments/assets/3de26f4c-5876-47c3-8658-251e2d8382e4
my video showing the screen going back without interaction, and when I did tap the confirmation quick enough, it worked.

gvodden:
https://github.com/user-attachments/assets/857c2eb3-d15a-48e2-8663-97dbad6e50a0
second video - same experience going the other way

jleach:
Fixed in 6368.

gvodden:
looks good to me in 6368

RaulConstan:
confirmed in 6376.

PR #3797, opened 2026-05-06
fix: interval cleanup on transfer account

bryce-mcmath:
# Summary of Changes

Intervals in the transfer account feature weren't being cleaned up properly, resulting in repeated navigation back to the same screen

# Testing Instructions

Transfer some accounts and try to slowly go through the remove account process on the transferrer

# Acceptance Criteria

No unwanted navigation

N/A

# Related Issues

#3796


al-rosenthal (approved):
tested on both iOS and android and it works as expected

jleach (approved):

PR #3800, opened 2026-05-06
fix: birthdate off-by-one in non-BCSC verification request

jleach:
Non-BCSC users entering DOB `2000-01-01` had `1999-12-31` sent to IAS. Cause: `new Date('YYYY-MM-DD')` parses as UTC midnight per ECMAScript, so in any TZ west of UTC the stored Date represented the prior calendar day, and the residential-address submission then formatted it back in local time.

The BCSC card flow already dodged this via `moment(value, format).toDate()`; the non-photo path slipped in with a plain `new Date(...)`.

Pulled the parsing into a shared `parseBirthdateToLocalDate` helper so both DOB-entry paths build local-midnight Dates from the same code, plus a TZ-invariant regression test (asserts local Date components instead of relying on `TZ=GMT` masking the bug).

# Manual Testin

1. Do a BCSC activation, should accept DoB as expected.

2. Do a Non-BCSC activation, the DoB you enter for ID 1 should be shown in the `idcheck` process. I did send video.


al-rosenthal (approved):
I've tested this locally and it works as expected

bryce-mcmath (approved):

Issue #3804, opened 2026-05-06
Verify In Person instructions are showing BCSC language for Non-BCSC flow.


gvodden:
### Expected behavior
Screen should not reference BC Services Card or Card Serial number, and should tell user to bring the ID types that were selected during setup.

Actual Behaviour:  Screen instructs the user to bring a BC Services Card, and shows the first document number entered as a "card serial number"

### Steps to reproduce
Compare v4 (left) with v3.15 (right):

<img width="958" height="1140" alt="Image" src="https://github.com/user-attachments/assets/0de4008b-6b06-41ba-82c5-ce0af0fbf705" />

### Environment

Build #:

Android Device Model:

iOS Device Model:

bryce-mcmath:
will be in 6376

RaulConstan:
Confirmed in 6376 non-BCSC flow:

<img width="462" height="1128" alt="Image" src="https://github.com/user-attachments/assets/af9c033e-5457-4627-b73a-8f3a7aa4d347" />

RaulConstan:
Confirmed in 6376 BCSC flow:

<img width="462" height="1128" alt="Image" src="https://github.com/user-attachments/assets/21664867-f583-4e8b-afbb-d9d36fd30892" />

PR #3806, opened 2026-05-06
fix: copy on in-person verification screen

bryce-mcmath:
# Summary of Changes

Non-BCSC specific copy for verify in person screen

# Testing Instructions

Do a non-BCSC verify in person flow

# Acceptance Criteria

Copy looks correct as per related issue

IMG_0456

# Related Issues

#3804

### DIFF
--- a/app/src/localization/en/index.ts
+++ b/app/src/localization/en/index.ts
@@ -757,6 +757,7 @@ const translation = {
       "WhereToGoLink": "A Service BC Location",
       "WhatToBring": "What to bring",
       "ThisDevice": "This device",
+      "YourID": "Your ID, including",
       "YourBCServicesCard": "Your BC Services Card - if it's a non-photo card, bring your additional ID too",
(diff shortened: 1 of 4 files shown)


app/src/localization/en/index.ts
```diff
@@ -757,6 +757,7 @@ const translation = {
       "WhereToGoLink": "A Service BC Location",
       "WhatToBring": "What to bring",
       "ThisDevice": "This device",
+      "YourID": "Your ID, including",
```
al-rosenthal, line 760: no French and portuguese versions?

jleach (approved):
Confirmed on iOS, copy changed.

app/src/localization/en/index.ts
jleach: +1

bryce-mcmath: app is english only until phase 2 or later

al-rosenthal (approved):
ran locally and copy is changed

Issue #3835, opened 2026-05-08
Incorrect time used for Service Hours on Video Call


RaulConstan:
### Description of problem
The v4.0.0 6402 app calls:
[08/May/2026:07:58:58 -0700] 216.232.79.208 "GET /video/v2/service_hours HTTP/1.1" 200 516 "BCServicesCard/4.0.0 (Android 16; Build 6402)" "-" 1850 3329 12976 +
	
[08/May/2026:07:58:58 -0700] 216.232.79.208 "GET /video/v2/destinations HTTP/1.1" 200 84 "BCServicesCard/4.0.0 (Android 16; Build 6402)" "-" 1849 2896 4088 +

This results in a mismatch with the IAS server hours of operation for Video Call.

### Expected behavior
v4 and IAS are on the same time periods

### Steps to reproduce
v4 is off by 1 hour than the IAS PROD
1. Complete setup steps to Step 4 BCSC flow
2. Click step 5 and select Video Call between 7:30 am and 8:00 am

### Screenshots and/or log output

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/9ab8d515-e506-4428-bae9-0273065c1699" />

### Environment
PRODUCTION

Build #:
6402

Android Device Model:
S25 (Android 16)

iOS Device Model:

jleach:
Probably related to the date parsing as resolved in #3800. If so, try and use the same helper so we consolidate date / TS parsing to a single location the code.

bryce-mcmath:
will be in 6428

RaulConstan:
Confirmed in 6434 (PROD Video Call)

Issue #3834, opened 2026-05-08
Camera zooms 2x after the back button used


RaulConstan:
### Description of problem
The camera zooms after scanning the first time and using the back button to re-scan a CSN in step 2.

### Expected behavior
Honestly not sure - Clecio did talk about a 2x zoom but not clear when it's supposed to happen.

### Steps to reproduce
1. Create new account
2. Step 2: scan a CSN
3. Click back button
4. Get to the camera scan again
5. Observe the 2x zoom the on the 2nd camera scan.

### Screenshots and/or log output

https://github.com/user-attachments/assets/b9659001-cdcb-4910-b1d4-4655f81f266e

### Environment

Build #:
6402

Android Device Model:
S25 Android 16.

iOS Device Model:

jleach:
Chatted about this internally. Plan is to drop the initial zoom from 2x to 1x on the scan-serial screen — matches v3 and the first-mount state, so coming back won't surprise-zoom anymore.

Leaving the component default at 2x in case it's useful elsewhere down the road.

bryce-mcmath:
will be in 6428

bryce-mcmath:
apk available here: https://github.com/bcgov/bc-wallet-mobile/actions/runs/25583706406

bryce-mcmath:
closing as per comments on #3844

PR #3837, opened 2026-05-08
fix: timezone and initial zoom

bryce-mcmath:
# Summary of Changes

- Timezone corrected (BC is now permanently +7, no more DST changing)
- Initial zoom always 1x now (there is a bug on Android that is fixed in v5 of react native vision camera that will enable us to use other initial zooms without issue)

# Testing Instructions

Card serial scanning. Timezone is hard to verify except at 5pm - 6pm or 7:30am-830am

# Acceptance Criteria

Works

N/A

# Related Issues

#3834 
#3835

### DIFF
--- a/app/src/bcsc-theme/utils/service-hours-formatter.ts
+++ b/app/src/bcsc-theme/utils/service-hours-formatter.ts
@@ -199,7 +199,7 @@ const getCurrentTimeInTimezone = (timezone: string): Date => {
   const now = new Date()
   if (timezone === PACIFIC_TIMEZONE) {
     const utcTime = now.getTime() + now.getTimezoneOffset() * 60000
-    const pacificOffset = -8 * 60 * 60000
+    const pacificOffset = -7 * 60 * 60000
     return new Date(utcTime + pacificOffset)
   }
   return now
--- a/app/src/bcsc-theme/components/CodeScanningCamera.tsx
+++ b/app/src/bcsc-theme/components/CodeScanningCamera.tsx
@@ -128,7 +128,7 @@ const CodeScanningCamera: React.FC<CodeScanningCameraProps> = ({
   showBarcodeHighlight = false,
   enableScanZones = false,
   scanZones,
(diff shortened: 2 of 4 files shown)


app/src/bcsc-theme/utils/service-hours-formatter.ts
```diff
@@ -199,7 +199,7 @@ const getCurrentTimeInTimezone = (timezone: string): Date => {
   const now = new Date()
   if (timezone === PACIFIC_TIMEZONE) {
     const utcTime = now.getTime() + now.getTimezoneOffset() * 60000
-    const pacificOffset = -8 * 60 * 60000
+    const pacificOffset = -7 * 60 * 60000
```
jleach, line 202: I ran into a UTC offset issue in PR #3800 and used a helper with moment. No option for that here? I think you can ignore Copilot's comments since BC does not change the TZ any more but, but I think impact users who are traveling - say living in Ontario who still does the TZ change? Or is that only for display purposes and people need to compute the TZ offset manually?

jleach (changes requested):

app/src/bcsc-theme/utils/service-hours-formatter.ts
bryce-mcmath: The times are specified to be "Pacific Time" to the user, so it will be the same regardless of where the user is. If someone in Ontario tries to call at 7:29am Pacific, it will show them the Closed view, at 7:30am it will make the call. I'm also trying to keep changes to a minimum.

al-rosenthal: it's used to calculate if the current time is within the working hours of the live call service so we could probably swap that out with the helper function

jleach: Ok. Good enough for now.

jleach (approved):

al-rosenthal (approved):

Issue #3841, opened 2026-05-08
Older unused Android app has missing token issue after upgrade to v4


bryce-mcmath:
### Description

A fully verified app from v3.10 or earlier will result in a missing token bug in v4. We need a similar fallback check for the older storage location the way we do for the registration token (provider style)

bryce-mcmath:
will be in 6490

jleach:
No notable issues. Will need to be tested by external members.

PR #3854, opened 2026-05-12
fix(wip): fallback refresh token check

bryce-mcmath:
# Summary of Changes

# Testing Instructions

# Acceptance Criteria

# Related Issues

#3841


packages/bcsc-core/android/src/main/java/com/bcsccore/storage/NativeCompatibleStorage.kt
```diff
@@ -361,6 +361,13 @@ class NativeCompatibleStorage(
             tokensFile.delete()
         }
 
+        // Also delete V3 providers file to prevent orphaned token reuse
```
al-rosenthal, line 364: this feels risky, if we are potentially missing something else from this providers file and delete it, there is no recovery option

PR #3865, opened 2026-05-13
fix: fallback for missing clientId, issuer

bryce-mcmath:
# Summary of Changes

Older versions of v3 migrated through many different versions can have empty clientID / issuer fields on their account object. This fallback catches that scenario, and writes it to their account going forward. This prevents a failing startup where client ID and issuer are required for a token fetch

# Testing Instructions

I have not found a reproducible migration path yet. For now, good enough to test fresh installs and simple migrations from latest v3.

# Acceptance Criteria

Fixes the app startup issue that we thought was related to tokens

N/A

# Related Issues

#3841

### DIFF
--- a/app/src/bcsc-theme/hooks/useSecureActions.tsx
+++ b/app/src/bcsc-theme/hooks/useSecureActions.tsx
@@ -19,6 +19,7 @@ import {
   EvidenceType,
   getAccount,
   getAccountFlags,
+  getAccountSecurityMethod,
   getAuthorizationRequest,
   getCredential,
   getEvidence,
@@ -27,6 +28,7 @@ import {
   NativeAuthorizationRequest,
   NativeSavedService,
   PhotoMetadata,
+  setAccount,
   setAccountFlags,
   setAuthorizationRequest,
   setCredential,
@@ -733,6 +735,7 @@ export const useSecureActions = () => {
 
       // Load all data from native storage in parallel
       const [
+        account,
         authRequest,
         refreshTokenObj,
         registrationAccessTokenObj,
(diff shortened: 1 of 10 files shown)


app/src/bcsc-theme/hooks/useSecureActions.tsx
```diff
@@ -761,6 +765,41 @@ export const useSecureActions = () => {
...
+                ...account,
+                issuer: recoveredIssuer,
+                clientID: recoveredClientID,
+                securityMethod,
+              }
+              logger.info('[hydrateSecureState] Recovered issuer/clientID from authorization request; updating account')
+              await setAccount(updatedAccount)
+            } else {
+              logger.warn(
+                `[hydrateSecureState] Could not recover account fields from authorization request: issuer=${recoveredIssuer ? 'OK' : 'MISSING'}, clientID=${recoveredClientID ? 'OK' : 'MISSING'}`
```
al-rosenthal, line 792: should this throw if it can't recover from authRequest?

bryce-mcmath: my thinking is no, in the case that they have created an account but done nothing else, not even nickname. then it's harmless because as soon as they add nickname it will add the missing fields to their account object and doesn't need to do any refresh token requests (not verified yet)

jleach:
One small thing if you've got the appetite: a Kotlin unit test on `readAuthorizationRequestFromV3Provider` with a fixture missing the nested `authorizationRequest` (just `{ "clientRegistration": { "client_id": "abc" } }`) would lock this in.

al-rosenthal (dismissed):
tested this locally, migration and fresh builds both work

jleach (dismissed):
LGTM. Tested on Android. New setup OK, Restart OK, Login to BC Parks OK..

al-rosenthal (approved):

jleach (approved):

Issue #3923, opened 2026-05-26
Accessibility pairing code entry allows for excess characters


bryce-mcmath:
### Description

When using VoiceControl / Voiceover on iOS, the manual pairing entry enables the user to enter more than six characters, resulting in a failed request (and an ugly error message)

### Expected behaviour

Voice controlled character entry works the same way as typical keyboard usage

franko-c:
my read is that `handleChangeCode` at `ManualPairing.tsx:33` strips non-alphanumeric but doesn't cap to `PAIRING_CODE_LENGTH` — VoiceControl dictation bypasses the native keyboard's maxLength enforcement. adding `.slice(0, PAIRING_CODE_LENGTH)` after the `.toUpperCase()` should close the gap.

bryce-mcmath:
6743

RaulConstan:
Confirmed in 6764.

PR #3924, opened 2026-05-26
fix: cut off excess characters in code input

bryce-mcmath:
# Summary of Changes

Previously, users using Voice Control could enter more than the max six characters into the pairing code input (email confirmation input had the same issue as they are a common component.) This change simply truncates any additional hidden characters that voice control put past the input. So if you say via Voice Control, "H T R D E F F F F", previously it would attempt to send "HTRDEFFFF", whereas now it only sends "HTRDEF".

# Testing Instructions

Get to a manual pairing code entry, turn on Voice Control ("Hey Siri, turn on Voice Control"), make the input active, and then whatever the pairing code is, speak them, but also speak extra letters after. You should still log in successfully.

# Acceptance Criteria

Pairing code login via Voice Control works.

N/A

# Related Issues

#3923


KjartanE (commented):
Image

Not working perfectly, when icon is at end of input. When I am speaking "F" nothing happens but suggests I say "Tap F". Also on the email confirmation screen. I can speak letters into the input fields, or if I say the number "one" it types out those letters

KjartanE (dismissed):

KjartanE (approved):
Working way better! Nice work! 😻

Issue #3888, opened 2026-05-20
Remove setup steps screen


bryce-mcmath:
### Description

In 4.1, the setup steps screen should be removed. Resuming verification should pick up where the user left off using the same logic the setup steps screen currently uses to determine which button is active.

Issue #3822, opened 2026-05-07
4.1 stepper (dev)


bryce-mcmath:
### Description

As per Figma, each stack screen in stack navigators should be configurable with a progress bar, which can be created using a custom header component.

## Implementation Notes

- Figma has two different looks for this. Check with UX to see what one they want implemented.
- If this is overly long and complicated push back. It is eye candy and can be deferred.

PR #3927, opened 2026-05-27
feat: remove setup steps and add stepper

bryce-mcmath:
# Summary of Changes

This PR removes the SetupSteps screen and translates the same logic into direct navigation and "resuming" verification from any point from the home tab. It also adds the fancy stepper.

Note: there is currently no way to navigate back to the MainStack once you've begun/resumed verifying. This should be tackled in a follow up PR after conversation with UX team. I also suspect that the push notification handlers for successful video verification won't work as expected from the MainStack either. Another ticket.

# Testing Instructions

Try out different verify flows.

# Acceptance Criteria

Works without the setup steps screen.

https://github.com/user-attachments/assets/5e9d30cb-64d4-469b-9d61-6ce011754b18

# Related Issues

#3822 
#3888


al-rosenthal (changes requested):
a few things I noticed in testing:
- when sending a video in for verification, if the request is rejected, the app sees the rejection message but send the user back to the status screen. The request probably needs to be deleted from the apps side so the user is navigated back to the verification options
- During the other card flow, I got to the email verification screen, got the email code, closed and reopened the app and was at the verification selection screen. This was all done within the 5 minute time limit 
- unrelated to the PR, but if I wait on the email code screen for more than 5 minutes, the code still works

bryce-mcmath:
> * when sending a video in for verification, if the request is rejected, the app sees the rejection message but send the user back to the status screen. The request probably needs to be deleted from the apps side so the user is navigated back to the verification options

fixed this

> * During the other card flow, I got to the email verification screen, got the email code, closed and reopened the app and was at the verification selection screen. This was all done within the 5 minute time limit

this revealed another bug, as discussed it will be put into another ticket and PR

> * unrelated to the PR, but if I wait on the email code screen for more than 5 minutes, the code still works

there is no functional 5 minute email code behaviour (yet?), we can talk to UX but I think that text was just a creative wireframe addition

al-rosenthal (approved):
navigation issue fixed, everything else worked as expected. Other issues are out of scope of the work and will be done at a later date

PR #3935, opened 2026-05-28
fix: stale wallet

KjartanE:
# Summary of Changes

Fixes a stale-wallet bug where signing out and back in (without restarting the app) left the agent broken — scanned credential offers would hang and the mediator never reconnected until a full app restart.

**Root cause:** `BCSCAgentProvider` (which owns the Credo agent) is only rendered while authenticated, so signing out unmounts it — but nothing shut the agent down on unmount. It lingered as a zombie holding the shared Askar wallet open and its mediator live-session socket alive. The next sign-in built a new agent that collided with the zombie on the same wallet/mediator, so offers were delivered to the dead session.

**Fix:**
- Tear the agent down when the provider unmounts (sign-out), via a dedicated unmount-only effect.
- Serialize wallet open/close through a module-level queue in `agent-service` so a sign-in rebuild waits for the previous agent's shutdown before reopening the shared wallet. The old and new agents live in separate hook instances (unmount → remount), so this can't be coordinated with component refs.

# Testing Instructions

1. Sign in and confirm the wallet works (e.g. receive a credential).
2. Sign out via Settings.
3. Sign back in **without closing the app**.
4. Scan a QR credential offer — it should be received and the notification banner should appear, with no force-restart needed.

Unit tests: `yarn jest src/bcsc-theme/features/agent`

# Acceptance Criteria

- After sign-out → sign-in (no app restart), the agent reconnects to the mediator and can receive credentials.
- Scanning a credential offer after re-login surfaces the offer/notification without a force-restart.
- No regression to first-launch agent init, retry, or factory reset.

N/A

# Related Issues

N/A

### DIFF
--- a/app/src/bcsc-theme/features/agent/services/agent-service.test.ts
+++ b/app/src/bcsc-theme/features/agent/services/agent-service.test.ts
@@ -58,6 +58,7 @@ jest.mock('@credo-ts/indy-vdr', () => ({
 import {
   AgentWalletSecret,
   buildAgent,
+  initializeAgent,
   loadCachedLedgers,
   restartAgent,
   shutdownAgent,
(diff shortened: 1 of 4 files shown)


bryce-mcmath (approved):
neat trick

PR #3944, opened 2026-05-29
fix: agent initialization

KjartanE:
# Summary of Changes

Hardens the BCSC agent lifecycle so init, shutdown, and wallet-store deletion no longer collide on the shared Askar wallet, fixing the load/reset errors. Also fixes an empty section header in BCSC Settings.

**Serialize all wallet operations (`agent-service.ts`)**

- Added a `deleteWalletStore()` helper that routes `askar.deleteStore()` through the same `enqueueWalletOp` queue as init/shutdown, so a store delete can never run concurrently with an open/close handle (the source of transaction / `onCloseContext` errors). Unlike `shutdownAgent`, it does not swallow errors — callers decide if a failed delete is fatal.
- Made `shutdownAgent` idempotent: it re-checks `agent.isInitialized` inside the serialized op. A reset path may have already shut the agent down, and a second `agent.shutdown()` runs against an already-closed store, throwing `"There is no open store"`.

**Remove concurrent-cache contention (`bc-agent-modules.ts`)**

- Registered an in-memory `CacheModule` (`InMemoryLruCache`, limit 500) to replace Credo's default storage-backed cache. The default rewrites a single Askar record on every `set()` (assumes serial access), but BCSC init writes the DID→ledger cache concurrently (parallel `warmCache` + the live mediator pickup loop), causing `"Duplicate entry"` / transaction rollbacks. These caches are re-resolvable, so dropping persistence is safe and removes the contention.

**Warm cache cold-start (`agent-service.ts`)**

- `warmCache` now pre-resolves each unique issuer DID once, sequentially, before the parallel cred-def/schema reads. Many cred-defs share an issuer DID, so resolving them concurrently fired N simultaneous cold-cache lookups for the same DID; pre-warming makes the parallel reads cache hits. Per-DID failures are non-fatal.

**Factory reset (`useFactoryReset.tsx`)**

- Shut the agent down *before* deleting the store (deleting first leaves the agent's Askar context pointing at a removed store, so shutdown's `onCloseContext` throws). Both ops now go through the shared `shutdownAgent` / `deleteWalletStore` helpers and the wallet-op queue.

**View model (`useAgentSetupViewModel.ts`)**

- Wallet-reset and recovery paths now delete the store via the shared `deleteWalletStore()` helper instead of calling `askar.deleteStore()` directly, so deletion is serialized with init/shutdown.

**Settings empty header (`SettingsContent.tsx`)**

- The Features (Contacts) section header was rendered unconditionally, with the Contacts action conditionally nested inside, so an empty header showed when no `onContacts` handler was provided. Moved the `onContacts` guard to wrap the whole header + card, so the section only renders when there is an action.

Unit tests added/updated for `agent-service`, `useFactoryReset`, `useAgentSetupViewModel`, and `bc-agent-modules`.

# Testing Instructions

Automated:

- `cd app && yarn test` — covers `agent-service.test.ts`, `useFactoryReset.test.ts`, `useAgentSetupViewModel.test.ts`, and `bc-agent-modules.test.ts`.

Manual:

1. Fresh install → onboard and reach the home/account screen → confirm the agent initializes with no Askar errors in the logs (`"Duplicate entry"`, transaction rollback, `"There is no open store"`).
2. Reset wallet (Settings) and re-onboard → agent rebuilds cleanly on the same device, no duplicate-store error.
3. Factory reset → app does not crash; wallet store is deleted and state cleared even if shutdown reports an error.
4. Open BCSC Settings where the Contacts feature is unavailable → the Features/Contacts section header is not rendered; where it is available → the header and Contacts action both show.

# Acceptance Criteria

- Agent init, shutdown, and store deletion are serialized; no Askar concurrency errors during init, wallet reset, or factory reset.
- Re-onboarding after a reset succeeds without a duplicate-store error.
- Factory reset always completes (store deleted, state cleared) regardless of shutdown errors.
- BCSC Settings does not render the Features (Contacts) section header when no Contacts action is available.
- All unit tests pass.

N/A

# Related Issues

N/A


al-rosenthal (commented):
I was able to consistently trigger an error with agent initialization:

Image

I was able to do this by resetting my wallet then removing my account before the wallet has re initialized, you'll see the above error every time the app opens and the wallet attempts to initialize. You can fix this state by resetting the wallet again. 


Aside from this error though, no more errors on start up!

al-rosenthal (approved):

Issue #4016, opened 2026-06-09
ERR_2822: Account not found — app state missing or corrupted


jleach:
### Edit:
Same error with different code is occurring:
```
{
  "message": "Device Authentication Failed",
  "data": {
    "title": "Device Authentication Failed",
    "description": "Unable to authenticate using the device's built-in authentication.",
    "code": 2707,
    "message": "Device authentication failed — biometric or passcode verification did not succeed\nDebug: [device.device_authentication_error.2707] E_ACCOUNT_NOT_FOUND: No account found\nScreen: Account Select"
  }
}
```

## Problem Report

**Error code:** 2822
**User-facing message:** Problem with App
**User-facing description:** The app does not appear to be installed correctly. Please remove the app from your device and add it again. (error 2822)

## Debug Info

```
[general.account_not_found.2822] No account found for the current user — state may be corrupted or missing
```

## Log Sample

```
2026-06-08 10:27:31.210 error {
  "message": "Problem with App",
  "data": {
    "title": "Problem with App",
    "description": "The app does not appear to be installed correctly. Please remove the app from your device and add it again. (error 2822)",
    "code": 2822,
    "message": "No account found for the current user — state may be corrupted or missing\nDebug: [general.account_not_found.2822]"
  }
}
```

## Notes

- Count: 1 occurrence
- Account record absent at point of use — could be a failed initial write, partial state corruption, or secure storage loss
- Possibly part of a storage failure cluster:
  - #4015 (ERR_2600): disk-full write failure — could prevent account data from ever being persisted
  - #4014 (ERR_2502): null token on read — consistent with missing/incomplete stored state

Issue #4015, opened 2026-06-09
ERR_2600: Storage write failure due to device out of disk space (error 100)


jleach:
## Problem Report

**Error code:** 2600
**User-facing message:** Problem with App
**User-facing description:** The app does not appear to be installed correctly. Please remove the app from your device and add it again. (error 100)

## Debug Info

```
[storage.err_100_failed_to_write_local_storage.2600] AsyncStorage or secure storage write operation failed
An unexpected File IO error occurred — volume "User" is out of space
```

## Log Sample

```
2026-06-08 15:20:47.416 error {
  "message": "Problem with App",
  "data": {
    "title": "Problem with App",
    "description": "The app does not appear to be installed correctly. Please remove the app from your device and add it again. (error 100)",
    "code": 2600,
    "message": "AsyncStorage or secure storage write operation failed\nDebug: [storage.err_100_failed_to_write_local_storage.2600] An unexpected File IO error occurred! Error: You can't save the file \"[redacted].jpg\" because the volume \"User\" is out of space."
  }
}
```

## Notes

- Count: 1 occurrence
- Root cause is clear: device storage is full; the write fails at the OS level
- User-facing message ("reinstall the app") is misleading — reinstalling won't help if the device has no free space; should detect low-storage conditions and surface a more actionable message
- Possibly related to #4014 (ERR_2502: null token) — a disk-full write failure in a prior session could leave a token unwritten, causing a null read on next launch

jleach:
Also possibly related to #4016 (ERR_2822: account not found) — a disk-full write failure could prevent account state from being persisted, leading to this error on next launch.

cmathies16:
Investigate clean up logic.

knguyenBC:
Provide to devs plain language for this error to replace current user-facing description

PR #4025, opened 2026-06-09
fix: error 2502 token hydration

### LINKED ISSUE #4014: ERR_2502: Token unexpectedly null — storage read failure (error 119)
## Problem Report

**Error code:** 2502
**User-facing message:** Problem with App
**User-facing description:** The app does not appear to be installed correctly. Please remove the app from your device and add it again. (error 119)

## Debug Info

```
[token.err_119_token_unexpectedly_null.2502] Token reference was null at point of use — storage read may have failed
```

## Log Sample

```
2026-06-09 04:56:47.713 error {
  "message": "Problem with App",
  "data": {
    "title": "Problem with App",
    "description": "The app does not appear to be installed correctly. Please remove the app from your device and add it again. (error 119)",
    "code": 2502,
    "message": "Token reference was null at point of use — storage read may have failed\nDebug: [token.err_119_token_unexpectedly_null.2502]"
  }
}
```

## Notes

- Count: 2 occurrences
- Token read from secure storage returned null at point of use — could indicate secure storage corruption, a failed write on a prior session, or a race condition during token retrieval
- User-facing guidance is to reinstall, suggesting this is unrecoverable without resetting state

KjartanE:
# Summary of Changes

Fixes **ERR_2502 (`TOKEN_NULL`)** — the "Problem with App… reinstall (error 119)" message users hit on launch — and makes the app self-heal when the failure was transient.

`BCSCApiClient.tokens` is in-memory only and rebuilt each session by `hydrateSecureState()` via the stored refresh token. That step can fail silently (network failure during the startup refresh — e.g. airplane mode mid-unlock — or running before the client is ready), leaving the cache empty. The next read threw `TOKEN_NULL`, surfacing a misleading "reinstall the app" prompt for a fully recoverable state, and the session then stayed broken until an app restart.

**Token recovery**

- Added `BCSCApiClient.recoverTokens()` — rebuilds the in-memory token cache from the refresh token in secure storage when it's empty. Throws `TOKEN_NULL` **only** when no refresh token exists (the genuinely unrecoverable case — reinstall is then correct).
- Routed both throw sites (`ensureValidTokens()` and `getCachedIdTokenMetadata()`) through the recovery path. Recoverable cases now self-heal, or surface an accurate network error instead of a reinstall prompt.

**Post-failure healing** (offline launch previously bricked the session until restart)

- `recoverTokens()` now emits `TOKENS_REFRESHED` after a successful rebuild. The event had listeners (`BCSCAccountContext`, `useSystemChecks`) but **no emitter anywhere** — they were dead code.
- `BCSCIdTokenContext`: `refreshData` was a no-op when the first load failed; it's now always functional. Also listens for `TOKENS_REFRESHED` (parity with the account provider) and reloads from the freshly cached tokens.
- New `useRetryOnReconnect` hook: ID-token and account providers retry automatically when connectivity is restored, removing the dead `!apiClient.tokens` recovery hack that threw before it could ever recover.

# Testing Instructions

1. Onboard/verify so a refresh token is stored, then force-quit the app.
2. Launch, trigger the FaceID/PIN prompt, enable airplane mode, then complete unlock.
   - **Before:** "Problem with App… reinstall (error 119)" and the session stayed broken after reconnecting.
   - **After:** no false 119 — failures surface as network errors (`no_internet.2100`); on disabling airplane mode, ID-token and account data reload automatically (look for `Token cache empty; rebuilding from stored refresh token` then `Tokens refreshed, reloading…` in logs).
3. Confirm normal online launch and ID-token-backed screens (Services, system-check banners) still load.
4. Unit tests: `cd app && yarn typecheck && yarn test src/bcsc-theme`.

# Acceptance Criteria

- A missing in-memory token cache with a valid stored refresh token recovers transparently instead of throwing `TOKEN_NULL`.
- `TOKEN_NULL` / reinstall messaging only appears when no refresh token is recoverable.
- After an offline launch, the app heals on reconnect without a restart (account + ID token data reload).
- No regression to normal launch, token refresh, or authenticated requests; typecheck, lint, and tests pass.

N/A — no UI changes.

# Related Issues

Closes #4014 (ERR_2502). Also removes the #4015 → #4014 knock-on (a missing token from a prior failed write no longer triggers a false 119). Low-disk messaging for #4015 (ERR_2600) and #4016 (ERR_2822) are out of scope.

### DIFF
--- a/app/src/bcsc-theme/api/client.test.ts
+++ b/app/src/bcsc-theme/api/client.test.ts
@@ -6,7 +6,7 @@ import { localization } from '@/localization'
 import { initLanguages } from '@bifold/core'
 import { AxiosError } from 'axios'
 import { jwtDecode } from 'jwt-decode'
-import { getAccount } from 'react-native-bcsc-core'
+import { getAccount, getToken, TokenType } from 'react-native-bcsc-core'
 
 jest.mock('jwt-decode', () => ({
   jwtDecode: jest.fn(),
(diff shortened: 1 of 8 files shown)


al-rosenthal (approved):
works as described

PR #4028, opened 2026-06-10
fix: revert BCSC PROD app icon to v3

### LINKED ISSUE #4019: Revert the app icon to for PROD to match v3
Revert the BCSC v4.0 PROD icon back to the previous version used in v3.x for both iOS and Android.

jleach:
Reverts the BC Services Card **production** app icon back to the v3 design — the white ID card on a blue background — for both iOS and Android.

**Scope:** only the `bcsc-prod` variant. `dev` / `qa` / `test` and BC Wallet are unchanged.

- **iOS** — uses the v3 marketing icon as a single app icon (no dark/tinted variants, same as v3).
- **Android** — restores v3's original adaptive icon (vector foreground/background), legacy icons, and Play Store icon.

The v3 assets were taken from the original native BC Services Card app, so this is a faithful restore rather than a regenerated icon.

Closes #4019.


KjartanE (approved):
Good to go

Issue #3992, opened 2026-06-08
ERR_2707: Android Device Authentication Failed


RaulConstan:
### Description of problem
Updating from 3.15.0 to 4.0.0 (6575) triggers an error.  The user is blocked from getting past the error.

Remote Logging Code: 832874

### Expected behavior
No errors

### Steps to reproduce
1. Have 3.15.0 from PlayStore installed with a verified ID on it
2. Be part of the rollout for 4.0.0
3. Update

### Screenshots and/or log output

<img width="225" height="564" alt="Image" src="https://github.com/user-attachments/assets/dee9c1ea-ecaa-4764-86cd-b3422b229c07" />

https://github.com/user-attachments/assets/434fa9bb-a57d-44ed-b035-54ca36675c5b

### Environment
PRODUCTION

Build #:
4.0.0 6575

Android Device Model:
S20 (Android 13)

iOS Device Model:

jleach:
Possibly related: #4013 (ERR_2111) also involves an invalid device code returning `invalid_grant` during the OIDC device flow. May be the same root cause or a related failure path.

hannahgov:
@bryce-mcmath "Some of your data on this device couldn't be read. To keep using the app, you'll need to re-verify or transfer from another device to use your identity again."

RaulConstan:
Confirmed fixed in 6809:

https://github.com/user-attachments/assets/4ea4c37c-bb06-4754-a755-6bd9ceb7bc2e

PR #4038, opened 2026-06-12
fix: corrupted token file

bryce-mcmath:
# Summary of Changes

This PR address the three different facets of the corrupted token file issue:
- prevents future corrupted tokens by making writes atomic
- returns null rather than throwing if a token is 0 bytes
- recovers when in the corrupted token situation

# Testing Instructions

Convert, make sure happy paths still work, then once you're verified, run the following commands:

```sh
adb shell run-as ca.bc.gov.id.servicescard.dev ls files/sit (get your account ID)
adb shell run-as ca.bc.gov.id.servicescard.dev cp /dev/null files/sit/<account ID>/tokens (corrupt your token file)
```
 
Relaunch app, ensure you are directed through the recovery flow and end up back on step 2 of the setup steps.

# Acceptance Criteria

Happy paths still work and recovery works too

https://github.com/user-attachments/assets/e96133a0-c911-4f89-9730-a3430d3b2b66

# Related Issues

#3992

### DIFF
--- a/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
+++ b/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
@@ -463,12 +463,20 @@ class BcscCoreModule(
 
                             promise.resolve(null)
                         } catch (e: Exception) {
-                            Log.e(NAME, "Failed to parse decrypted JSON content: ${e.message}", e)
-                            promise.reject("E_STORAGE_ERROR", "Failed to parse decrypted token data: ${e.message}", e)
+                            Log.e(
+                                NAME,
+                                "Failed to parse/migrate decrypted token content for type $tokenType — returning null",
+                                e,
+                            )
+                            promise.resolve(null)
                         }
                     } else {
-                        Log.e(NAME, "Decrypted content is not valid JSON")
-                        promise.reject("E_STORAGE_ERROR", "Decrypted token content is not valid JSON")
+                        Log.w(
+                            NAME,
+                            "Decrypted token content is empty or not valid JSON at $tokenFilePath — " +
+                                "returning null for type $tokenType",
+                        )
+                        promise.resolve(null)
                     }
                 } else {
                     Log.w(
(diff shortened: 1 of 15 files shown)


packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
bryce-mcmath, line 1: If decryption fails (0-byte file or non-JSON cases) return null rather than throwing to enable the recovery process

jleach:
Still DRAFT but `useSessionRecovery` could do with a few tests.

packages/bcsc-core/android/src/main/java/com/bcsccore/storage/NativeCompatibleStorage.kt
```diff
@@ -238,21 +239,51 @@ class NativeCompatibleStorage(
...
+
+            if (!tempFile.renameTo(file)) {
+                // renameTo can fail if the destination already exists on some platforms;
+                // fall back to delete-then-rename.
+                file.delete()
+                if (!tempFile.renameTo(file)) {
+                    throw IOException("Failed to rename temp file to ${file.absolutePath}")
+                }
             }
             true
```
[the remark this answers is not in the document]
bryce-mcmath: addressed. it still built fine for me originally but updated anyway to be safer

bryce-mcmath:
> Still DRAFT but `useSessionRecovery` could do with a few tests.

done

packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
```diff
@@ -467,8 +467,12 @@ class BcscCoreModule(
                             promise.reject("E_STORAGE_ERROR", "Failed to parse decrypted token data: ${e.message}", e)
```
jleach: Should we watch for failed token parsing? Might be a rare edge case if JSON parsing works but not token. Can circle back some ohter time if its uber rare.

bryce-mcmath: Haven't seen any instances of this error but could probably return null here as well, what do you think? until we come up with a way to send native logs to Loki it's essentially trading knowing where an error is for handling it gracefully without knowing it happened. I kind of think that's a good trade

jleach (dismissed):
Manual testing on a Pixel 7a worked. Tried from a warm and cold start. Both took me to the "Restore your account" and then Step 2.

hannahgov:
@bryce-mcmath "Some of your data on this device couldn't be read. To keep using the app, you'll need to re-verify or transfer from another device to use your identity again."

al-rosenthal (approved):
tested this locally, re verification worked, qr transfer worked

jleach (approved):

jleach:
Still works on a Pixel 7a.

PR #4062, opened 2026-06-17
fix: add diagnostic logging to jwe decrypt functions

KjartanE:
# Summary of Changes

**This is diagnostics, not a fix.** It does not change decryption behaviour — a user hitting **error 2507 (unable to decrypt JWE)** will still hit it. What it changes is our ability to tell _why_ from a field report alone.

## What we're trying to diagnose

2507 is thrown whenever the app fails to decrypt the `/userinfo` JWE the server returns (seen on the Home and Login Request screens). It has been trickling in during the staged production rollout (~10 reports / 7 days) and is expected to scale with the rollout percentage. The problem is that today **every distinct failure collapses into the same opaque 2507** — "no key", "wrong key", "unsupported algorithm", and "corrupt/truncated ciphertext" are indistinguishable in the report, so we cannot tell which real fix to build. The native path also surfaced a bare `Unable to decode payload` with no key inventory, and on iOS the unhelpful `localizedDescription` instead of the real `JOSEException` text.

This PR makes 2507 reports **self-classifying**. Every decode reject now appends a compact, bridge-safe summary `[keys, newest, jweParts, jweAlg, jweEnc, jweKid, kidMatchesLocal]` plus the underlying error message. Read straight off a report, those fields separate the root causes:

| Signature in the report                                                 | Cause it points to                                                                            |
| ----------------------------------------------------------------------- | --------------------------------------------------------------------------------------------- |
| `kidMatchesLocal=false`, `jweKid` is a real key id                      | Server encrypted to a key this device doesn't hold/select (rotation, migration, multi-device) |
| `[keys=0]` (persists)                                                   | No usable private key (restore to a new device, keystore wiped)                               |
| iOS `E_KEYCHAIN_UNAVAILABLE`, or transient `[keys=0]` that recovers     | Keychain locked / pre-first-unlock — retryable, not a real decrypt failure                    |
| `jweEnc`/`jweAlg` not in the supported set (e.g. `A256GCM`, `RSA-OAEP`) | Server changed the algorithm (iOS supports only RSA1_5 + CBC-HMAC)                            |
| `jweParts!=5` or `jweAlg=?`                                             | Corrupted / replaced JWE in transit (proxy, captive portal, truncation)                       |

## What changed

- **Android `BcscCoreModule.decodePayload`** and **iOS `BcscCore.decodePayload`**: build the diagnostics summary up front and append it (plus the underlying reason) to every reject path. The decryption attempt itself is untouched.
- **iOS** additionally distinguishes a temporarily-unavailable keychain (`E_KEYCHAIN_UNAVAILABLE`, retryable) from a genuine decryption failure, and reads the real `JOSEException` text instead of `localizedDescription`.
- The diagnostics already reach problem reports via `AppError.fullMessage` (`technicalMessage` = native code + message); no JS change was needed beyond a regression test that locks that contract.

# Testing Instructions

- Run the `appError` unit suite (`app/src/errors/appError.test.ts`). The new test confirms the native diagnostics + error code survive into `AppError.fullMessage` for 2507 (asserts `statusCode 2507`, the `token.err_110_unable_to_decrypt_jwe.2507` tag, `E_PAYLOAD_DECODE_ERROR`, and fields like `kidMatchesLocal=false`, `jweEnc=A256CBC-HS512`).
- Native (manual, Android + iOS): trigger a decode failure during get-user-info (e.g. force a key/`kid` mismatch) and confirm the problem-report message contains the `[keys=…, jweKid=…, kidMatchesLocal=…]` summary and the underlying reason.
- iOS: confirm a locked-device / keychain-unavailable failure surfaces as `E_KEYCHAIN_UNAVAILABLE` rather than a generic `E_PAYLOAD_DECODE_ERROR`.

# Acceptance Criteria

- 2507 problem reports include the decode diagnostics summary and the underlying error message.
- "No key" / "wrong key" / "unsupported alg" / "corrupt JWE" are distinguishable from a report alone.
- iOS keychain-unavailable failures report distinctly (`E_KEYCHAIN_UNAVAILABLE`) and are not conflated with decryption failures.
- Success path and existing decode behaviour are unchanged; the diagnostics builder never throws (unreadable fields → `?`).

# Next Steps

This is the first half of a deliberate **diagnostics-first** approach — ship observability, then fix the cause the data actually shows. After this has been in the field for ~1–2 weeks:

1. **Read the 2507 distribution** using the table above to find the dominant cause.
2. **Build the targeted fix** (none ships in this PR):
   - _Right key held but not selected_ (`kidMatchesLocal=false` with the kid present locally) → kid-aware + try-all-local-keys decryption in `decodePayload`, and enforce the iOS HMAC tag (currently computed but not checked). Self-contained, no server coordination.
   - _Key / server desync_ → wire the server-JWKS reconcile-and-retry (today it only runs on a token-endpoint 401) into the 2507 path.
   - _No key present_ (`[keys=0]`) → route to a clean re-registration flow instead of "Something went wrong", and stop Android silently seeding a fresh `rsa1` (which turns a recoverable no-key into a permanent mismatch).
   - _Unsupported alg/enc_ → coordinate with the backend before any `alg`/`enc` change (iOS supports only RSA1_5 + CBC-HMAC).
3. Full investigation and the complete ~25-scenario hypothesis catalog live in `tmp/err_2507_investigation.md`.

N/A

# Related Issues

N/A

### DIFF
--- a/app/src/bcsc-theme/services/hooks/useUserService.test.tsx
+++ b/app/src/bcsc-theme/services/hooks/useUserService.test.tsx
@@ -38,7 +38,7 @@ describe('useUserService', () => {
 
       await expect(result.current.getUserInfo()).rejects.toThrow(mockError)
       expect(userApi.getUserInfo).toHaveBeenCalled()
-      expect(mockAlerts.failedToDeserializeJsonAlert).toHaveBeenCalled()
+      expect(mockAlerts.failedToDeserializeJsonAlert).toHaveBeenCalledWith(mockError)
     })
 
     it('should show alert on JWE decryption error and rethrow error', async () => {
@@ -55,7 +55,10 @@ describe('useUserService', () => {
 
       await expect(result.current.getUserInfo()).rejects.toThrow(mockError)
       expect(userApi.getUserInfo).toHaveBeenCalled()
-      expect(mockAlerts.unableToDecryptJweAlert).toHaveBeenCalled()
+      // Regression: the alert must receive the ORIGINAL AppError. Calling it bare makes
+      // ensureAppError(undefined, …) rebuild a cause-less error, dropping the native decode
+      // diagnostics (E_…_ERROR + [keys=…]) from the problem report.
+      expect(mockAlerts.unableToDecryptJweAlert).toHaveBeenCalledWith(mockError)
     })
 
     it('should show alert on claims set error and rethrow error', async () => {
@@ -72,7 +75,7 @@ describe('useUserService', () => {
 
       await expect(result.current.getUserInfo()).rejects.toThrow(mockError)
       expect(userApi.getUserInfo).toHaveBeenCalled()
-      expect(mockAlerts.failedToGetClaimsSetAlert).toHaveBeenCalled()
+      expect(mockAlerts.failedToGetClaimsSetAlert).toHaveBeenCalledWith(mockError)
     })
 
     it('should show alert on JWS parse error and rethrow error', async () => {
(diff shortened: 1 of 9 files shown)


al-rosenthal (approved):

PR #4076, opened 2026-06-19
feat: report receipt

KjartanE:
# Summary of Changes

Adds a "Report this problem" action to the error modal. When tapped, it sends an incident report to Loki (Grafana) and surfaces a short, human-readable reference code (e.g. `7K2P-9XQF`) that the user can copy and share with support. Support can then locate the incident by that code, giving a way to correlate a user-reported error with backend logs without collecting PII. Reporting is best-effort — transport failures are swallowed so the user always gets a code.

# Testing Instructions

1. Trigger an error that surfaces the BCSC error modal.
2. Tap **Report this problem** — the button should switch to **Reported** (with a check) and a reference code (`XXXX-XXXX`) should appear with a "Share this code with support" hint.
3. Tap **Copy** — it briefly shows **Copied** (~2s) and the code is on the clipboard.
4. (Optional) In Grafana/Loki, confirm an `incident-report` entry arrives with a matching `report_id`.

# Acceptance Criteria

- "Report this problem" appears on the error modal when reporting is enabled.
- Tapping it sends an `incident-report` to Loki with the reference code as `report_id` and displays that code.
- The code is copyable; the confirmation shows briefly, then reverts.
- A reporting/transport failure neither crashes nor blocks the code from being shown.
- New strings are localized in en, fr, and pt-br.

image

# Related Issues

N/A

### DIFF
--- a/app/src/errors/components/ErrorInfoCard.test.tsx
+++ b/app/src/errors/components/ErrorInfoCard.test.tsx
@@ -1,3 +1,4 @@
+import Clipboard from '@react-native-clipboard/clipboard'
 import { fireEvent, render, waitFor } from '@testing-library/react-native'
 import React from 'react'
 import { ErrorInfoCard, ErrorInfoCardProps, colors as errorInfoCardColors } from './ErrorInfoCard'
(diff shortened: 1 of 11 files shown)


al-rosenthal (approved):
tested this locally and it works as expected

Issue #4069, opened 2026-06-18
Prevent audio desync caused by bluetooth devices


bryce-mcmath:
### Description

Some bluetooth devices are causing delays between audio and video in send-video requests

gvodden:
unable to reproduce this issue on v4.0.3 (6923) on multiple iPhones (6s, 11, 15).
Looks good to me!

PR #4083, opened 2026-06-22
fix: prevent audio desync in recorded video

bryce-mcmath:
# Summary of Changes

There were some reports of users using bluetooth devices having out of sync audio, this patch update prevents bluetooth usage in send video

I decided not to remove bluetooth support from live call since there have been zero reports of audio desync there and it's a different mechanism
 
# Testing Instructions

Use bluetooth headset, put your bluetooth headset in a drawer or another room (still connected) and then do a send-video request

# Acceptance Criteria

It doesn't record from the mic in your bluetooth headset

N/A

# Related Issues

#4069


KjartanE (commented):
Tested locally on iOS device, music did not stop. But video did record from device microphone

KjartanE (approved):
Working as intended

Issue #4072, opened 2026-06-18
Scheduled notifications for verifications still in progress


bryce-mcmath:
### Description

v3 schedules two local notifications when a verification is started to remind users to complete before their 7 days are up. We are missing that in v4. They should be scheduled and also cancelled once the user is verified. 
https://citz-cdt.atlassian.net/wiki/spaces/BMS/pages/301576844/Notification+Schedule+and+Re-schedule+logic

There are also rules around extensions of device code expiries we aren't matching here
https://citz-cdt.atlassian.net/browse/IASP-15043

So:
- schedule the notifications, cancel them once verified
- check approval status on VerifyStackSystemCheck
- update new expiry based on rules above

PR #4087, opened 2026-06-22
feat: scheduled local notis

bryce-mcmath:
# Summary of Changes

This PR adds scheduled local notification reminders to finish verification. They are cancelled when verification is complete or the account is removed. They are best effort, and don't throw, just log, if they fail. A new dependency is added, notifee, which is the standard React Native package for this type of feature.

# Testing Instructions

In verificationReminders.ts, change this variable to true:
```
const DEBUG_QUICK_REMINDER = true
```

Then, in a converted local build, get past entering your card serial number and birthdate. As soon as that's done and you're propelled to a later step, background or fully close the app and wait 20 seconds. You should receive the notification.

# Acceptance Criteria

Scheduled local notifications are received on both Android and iOS

Android & iOS
Screenshot_20260623_132401_One UI HomeIMG_0484

# Related Issues

#4072


al-rosenthal (approved):
tested this locally and works as expected on android and iOS

PR #4107, opened 2026-06-26
feat: back home button

KjartanE:
# Summary of Changes

Adds a **Back to home** option to the verification flow's help menu so users can leave an in-progress verification and return to the home screen **without losing progress** — they can resume from the home "Start/Continue verification" notification. The verification menu is now its own variant with a vertical-ellipsis (⋮) "more options" trigger offering Report a problem / Back to home / Restart verification; every other stack keeps the question-mark (?) menu (Learn more / Give feedback / Report a problem) unchanged.

Leaving is handled by a new `useLeaveVerification` hook, which moves verification status out of `IN_PROGRESS` so the RootStack renders the main stack instead of the verify stack. Splitting the help menu into default vs. verify variants also moved the shared report-problem modal wiring into a `useReportProblem` hook used by both.

# Testing Instructions

1. Start identity verification so the verify stack is shown.
2. On a verification screen, open the **⋮** menu (top-right) and tap **Back to home**.
3. Confirm you land on Home with no lost data and a "Continue verification" notification appears.
4. Tap the notification and confirm verification resumes at the step you left off.
5. Reopen the ⋮ menu: confirm **Report a problem** still opens the report modal and **Restart verification** still works.
6. On a non-verify screen (e.g. Home/Settings help), confirm the trigger is still the **?** icon with Learn more / Give feedback / Report a problem.

# Acceptance Criteria

- [ ] Verification help trigger is the ⋮ icon; all other stacks keep the ? icon.
- [ ] Verify menu shows Report a problem, Back to home, and — once verification has started — Restart verification.
- [ ] Back to home exits to Home, preserves progress, and verification can be resumed from the home notification.
- [ ] Default menu (Learn more / Give feedback / Report a problem) and the report-problem modal are unchanged for other stacks.

image

# Related Issues

N/A

### DIFF
--- a/app/src/bcsc-theme/components/FloatingHelpMenuHeaderButton.tsx
+++ b/app/src/bcsc-theme/components/FloatingHelpMenuHeaderButton.tsx
@@ -4,16 +4,25 @@ import { useNavigation } from '@react-navigation/native'
 import { StackNavigationProp } from '@react-navigation/stack'
 import React, { ReactNode, useCallback, useRef, useState } from 'react'
 import { useTranslation } from 'react-i18next'
+import { useLeaveVerification } from '../hooks/useLeaveVerification'
 import { useRestartVerification } from '../hooks/useRestartVerification'
 import { BCSCScreens } from '../types/navigators'
 import FloatingHelpMenu, { FloatingHelpMenuRef } from './FloatingHelpMenu'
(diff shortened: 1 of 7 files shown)


al-rosenthal (approved):
ran this locally and works as expected on both iOS and android

PR #4142, opened 2026-07-01
fix: guard InternetDisconnected navigate during root-stack swap

### LINKED ISSUE #3980: Unhandled NAVIGATE action for BCSCNoInternet after biometric setup
After completing the biometric/fingerprint setup step during onboarding, the following error appears in the debug console:

```
The action 'NAVIGATE' with payload {"name":"BCSCNoInternet"} was not handled by any navigator.
```

**Steps to reproduce:**
0. Disable WiFi
1. Complete onboarding up to the "Secure your app" step
2. Choose fingerprint / biometrics
3. Observe the debug console

**Expected:** No navigation error
**Actual:** Unhandled NAVIGATE action logged to console

jleach:
## What

`InternetStatusSystemCheck.onFail()` now only navigates to the `BCSCNoInternet`
(`InternetDisconnected`) modal when that route is actually registered in the currently-mounted
navigator. When no navigator owns the route, it skips the navigate instead of dispatching it.

Fixes #3980.

## Why

With WiFi disabled, completing the **fingerprint/biometric** "Secure your app" step during BCSC
onboarding logged:

```
The action 'NAVIGATE' with payload {"name":"BCSCNoInternet"} was not handled by any navigator.
```

Root cause:

- `useSystemChecks(STARTUP)` is mounted in `BCSCRootStack`, which sits directly under the single
  `NavigationContainer`, so its `useNavigation()` resolves to the **container (root) ref**.
- The biometric path calls `setupDeviceSecurity()`, which shows the OS fingerprint dialog and
  backgrounds→foregrounds the app. Returning to foreground fires the `AppState 'active'` internet
  check at the same instant `handleSuccessfulAuth` dispatches `SUCCESSFUL_AUTH` — atomically
  flipping `hasAccount` + `didAuthenticate`, which makes `BCSCRootStack` unmount `OnboardingStack`
  and mount `VerifyStack`.
- During that remount there is a transient window where no child navigator is mounted, so the
  container-level `navigate('BCSCNoInternet')` has nothing to handle it → the logged error.

The PIN path uses the same `handleSuccessfulAuth` but never spawns an OS dialog, so no foreground
check fires during the swap — which is why the error is biometric-only. `BCSCNoInternet` is
registered on all four stacks, so in steady state the navigate succeeds; only the transient
"no navigator mounted" window (and the non-navigator `LoadingScreen` branch) lack it.

## Decisions

- **Guard at the throw site, not the trigger.** Added a `canShowModal` check
  (`state?.routeNames?.includes(...)`) in `onFail()`. This is idempotent and localized —
  consistent with the repo's "favour idempotency over erroring" principle — rather than gating the
  `AppState`/`NetInfo` listeners in `useSystemChecks`.
- **Correctness is race-free.** `onFail()` reads the navigation state **once** and both guards plus
  the `navigate` operate on that same synchronous snapshot — if `routeNames` reports the route is
  available, that same navigator handles the dispatch.
- **No new plumbing.** The existing `SystemCheckNavigation.getState(): NavigationState | undefined`
  already exposes `routeNames` (confirmed in `@react-navigation/routers` types) — no interface,
  hook, or navigator-registration changes needed.
- **Accepted tradeoff.** During the transient swap window the modal is not shown (it wasn't shown
  before either — the navigate failed). It re-appears on the next `NetInfo` change or app-foreground
  event. Note this is event-driven only: nothing re-runs the check merely because a stack finishes
  mounting, so a device that goes offline exactly in the window stays modal-less until the next
  connectivity or foreground event — same as pre-PR behavior, scope here is limited to removing the
  spurious error.
- **Sibling checks deferred to #4147.** Post-review finding: `ServerStatusSystemCheck` has the
  identical unguarded navigate on the **same** foreground trigger (`recheckServerStatus()` runs
  right after the internet check in the AppState handler), and `UpdateAppSystemCheck` is unguarded
  in the mount batch. Rather than copy the private getter into each class, #4147 tracks promoting
  the guard to a shared `canNavigate` helper in `system-checks.ts` and adopting it in all three
  checks — kept out of this PR to preserve a small, device-testable diff.

## Tests

- `app/src/services/system-checks/system-checks.test.ts` — updated the existing `onFail` mock to
  include `routeNames`; added two cases: no navigator mounted (`getState()` undefined) and a
  navigator that doesn't register the route → `navigate` not called, warning still logged.
- `yarn test src/services/system-checks/system-checks.test.ts` (32/32), `yarn typecheck`, and
  `yarn eslint` on changed files all pass.

## Verification (manual — pending on-device)

1. Disable WiFi/cellular.
2. Fresh onboarding → "Secure your app" → choose **fingerprint/biometrics**.
3. Debug console no longer logs the unhandled `NAVIGATE ... BCSCNoInternet` error; onboarding
   proceeds to the Verify stack.
4. Sanity: with WiFi off on a normal stack (e.g. Home), background→foreground still surfaces the
   InternetDisconnected modal.


KjartanE (approved):
Tested locally on iOS, works as intended

PR #4178, opened 2026-07-07
feat(bcsc): self-heal device signing-key desync and confirm registration

### LINKED ISSUE #4166: Verify device signing-key registration and self-heal key/server desyncs by matching public keys
## What

The wallet authenticates to the identity server by signing requests with a device key. In an uncommon but real case, the key the app signs with can drift out of sync with the key the server has on record. When that happens the app can't refresh its session, and the only way back today is a full card re-setup.

Two improvements make this self-correcting:

1. **Confirm registration actually landed.** After the app registers or rotates its key, confirm the server has stored the key the app is about to sign with — *before* the app starts relying on it. A silent registration miss is then caught at the moment it happens instead of surfacing days or weeks later.

2. **Recover on the device instead of forcing a re-setup.** When a desync is detected, have the app ask the server which key it holds, find the matching key already on the device, switch to it, and clear out the leftovers — rather than sending the person back through card setup.

## Why

- **Fewer people sent through a full re-setup** for a problem the app can resolve itself.
- **Problems caught early, when they're cheap to fix** (at registration time) rather than later when the person is already stuck.
- **Lower support load** — fewer "I have to set up my card again" reports.
- **More resilient to long-lived / legacy devices** whose on-device key labels differ from the server's naming.
- Continuous-improvement hardening of the sign-in path; no user-facing behaviour change in the happy path.

### Shared foundation

- **Match on public key material, not `kid`.** The server relabels registered keys under its own naming (e.g. `rsa1`), so a device kid (e.g. a URL-style `…/device//kid/2`) will never string-match the server's kid even when it's the same key. Compare the **RSA modulus `n`** (or, preferably, an **RFC-7638 JWK thumbprint**) — that is exactly what determines whether a signature will validate.
- **Prerequisite — expose local public key material.** `getAllKeys()` currently returns `{ id, keyType, keySize, created }` with no public component (`packages/bcsc-core/src/index.ts`, iOS `BcscCore.swift` `getAllKeys`). Add a native accessor to return the public modulus / JWK / thumbprint per alias (iOS `BcscCore.swift`, Android `BcscCoreModule.kt`). Server side already exposes it.
- **Two endpoints, two credentials — this is what makes curative recovery possible.** `POST /device/token` authenticates with `private_key_jwt` (the signing key that may be desynced). `GET /device/register/{client_id}` authenticates with the **registration_access_token** (Bearer), independent of the signing key — so it keeps working during a signing-key desync. Recovery depends on this side channel (reg token must still be valid; if it isn't, fall through to re-setup).

### Part 1 — confirm registration after a key write

- After `createRegistration` / `updateRegistration` (`app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx`), confirm the server's jwks contains the active signing key's `n`. The register/update **response already echoes `jwks.keys[].n`**, so the happy-path check needs no extra round-trip; a follow-up `GET /device/register/{client_id}` is the stronger belt-and-suspenders confirmation.
- **Only promote a newly generated key to the active signing key after confirmation.** On failure, keep signing with the previously-confirmed key and retry later — never strand the device.
- **iOS sequencing caveat:** the active key is "newest by `kSecAttrCreationDate`" (read-only), so *generating* a new key makes it active immediately — before confirmation. The signer must therefore select the confirmed key explicitly, or the unconfirmed new key must be deleted so "newest" falls back to the prior key. Android selects by newest `createdAt` — handle equivalently.

### Part 2 — curative recovery (rework `performKeyRecovery`)

`app/src/bcsc-theme/utils/key-recovery.ts` (called from `app/src/bcsc-theme/hooks/useSecureActions.tsx`):

- On a `/device/token` 401: `GET /device/register/{client_id}` (reg token), read `jwks.keys[].n`.
- Compare each local key's `n` against the server set:
  - **Match found** → `setActiveKeyAlias(match)`, then prune local keys whose `n` is **not** in the server set (leaving the one confirmed key).
  - **No match** → the registered private key is not on the device; fall back to **re-registration** (PUT a fresh key with the reg token, or re-setup). **Do not prune** in this branch.
- **Encode consistently.** Derive the local `n` with the **same routine the registration path uses** (`JWK.jwkFromSecKey` → `Base64URL.encode`), or compare thumbprints — otherwise the same key can produce a different `n` string and false-mismatch (leading-zero / padding differences).
- **Keep existing guardrails:** refuse to prune on an empty/failed jwks response (current `failed_empty_jwks` guard); native delete continues to refuse removal of the final remaining alias.
- **Backward compatible:** modulus matching strictly generalizes today's kid-string match — devices whose kid already matches keep working.

## Acceptance / verification

- A device holding the server's registered key under a differently-labelled alias **recovers without a re-setup** (switches to the matching key, prunes the rest).
- A device holding **no** matching key falls back to re-registration and **does not** delete its keys.
- A key write that the server did not store is **detected at registration time**, and the app keeps using the previously-confirmed key.
- Modulus/thumbprint encoding round-trips between the registration path and the recovery path (unit test) so identical keys never false-mismatch.
- Empty/failed `GET /device/register` never triggers a prune.

jleach:
## What

The wallet signs its requests to the identity server with a device key. In rare cases that key can drift out of sync with what the server has on file, and today the only fix is a full card re-setup. This change makes that self-correcting: (1) after a device registers for the first time, the app double-checks the signing key it just sent actually landed on the server before relying on it, and (2) if the app ever detects it's signing with a key the server no longer recognizes, it now tries to recover on its own — asking the server which key it holds, switching to the matching local key, tidying up stale leftover keys, and if nothing matches, re-registering the current key — instead of forcing the user to redo card setup.

## Why

A small number of users hit a state where the app can't refresh its session because its signing key and the server's records disagree (most often after upgrading from an older app version that could leave more than one key on the device). Today the only way out is a full re-setup of the BC Services Card. This change lets the app detect and fix the common causes of that drift by itself, and adds an early check so a similar drift is caught right at initial registration instead of surfacing later as a broken session.

## Decisions

- **Compare on public-key material (modulus bytes), never on `kid`.** Device key identifiers have drifted across past migrations and the server verifies a signature by trying every key it has on file, so the RSA modulus is what actually determines validity — a `kid` string match can both activate the wrong local key and delete the right one. All decoding is tolerant of the different encodings iOS and Android (and legacy registrations) use, and never throws — undecodable input is always treated as "no match," never an error.
- **Initial-registration confirmation is scoped to `createRegistration` only**, not `updateRegistration` — this follows the issue's explicit scope decision (v4 doesn't rotate keys, so there's only one key in play at initial registration; the separate recovery flow is what catches a desync from any other cause).
- **Hard-fail on a definite mismatch.** If the sent key's modulus is decodably absent from the server's response (and the server's response itself contains at least one decodable key), registration now fails loudly with a new error code (`ERR_121`) so the existing setup-failure screen shows and the user can retry. If either side is unparseable/empty (a server-shape surprise, not a confirmed problem), the app logs a warning and proceeds rather than blocking setup.
- **Recovery never deletes a key it can't evaluate.** If a local key's public data can't be read for any reason, it's left alone rather than treated as confirmed-stale — deletion only happens when the server's response positively confirms a key is no longer recognized.
- **A hard gate after cleanup.** After realigning keys, the app re-checks that the intended key is unambiguously the one that will actually be used to sign. If that can't be confirmed, recovery reports failure rather than a false success.
- **No-match fallback re-registers automatically** (via a PUT to the registration endpoint) rather than deferring that to a future change, per the locked decision in the issue.
- Reused the existing shared OAuth scope value in the new re-registration path instead of hardcoding a new copy, and cross-referenced the three now-existing copies of that string so a future change to it doesn't get missed in one spot.

## Manual Testing

1. You'll see the app confirm it has checked key registration. Finish settign up the app, quit, return. It should work as expected. `[RegistrationApi] Confirmed sent signing key modulus is present in server jwks`
2. Upgrade from 3.x or 4.0. The app should work as expected. Quit, restart, and again it should work as expected.

## Tests

- New unit tests for the key-normalization helper covering the exact iOS/Android encoding differences, padding differences, and malformed input.
- Extended the registration-hook tests to cover: key confirmed among multiple server keys in a different encoding, a definite mismatch throwing the new error, and unparseable body/response cases proceeding with a warning.
- Rewrote the recovery-flow tests for the new modulus-based matching, including a regression case for the old kid-string-collision bug, the new "never delete an unevaluable key" safety behavior, and the hard post-cleanup gate (including when cleanup itself is rejected or the keystore reports empty).
- Added coverage in the secure-state hook for all three recovery outcomes (recovered / no match / failed) and for persisting a rotated registration token.
- Ran a structured self-review pass (8 independent angles: correctness, removed-behavior, cross-file impact, reuse, simplification, efficiency, architecture, and repo conventions) and fixed the findings that mattered — the most important being the "never delete a key we can't evaluate" safety behavior above.
- Full suite: 254 test files / 2378 tests passing. Typecheck, TS lint (both packages), Kotlin lint, and formatting all clean.

## Verification

This PR cannot be smoke-tested on-device from this environment (no simulator/emulator access). Manual verification still needed before release:
- **Fresh registration**: complete a normal card setup and confirm no ERR_121 / failure screen appears (the happy path should be unaffected).
- **Simulated desync recovery**: on a device with a stale/mismatched signing key (e.g. via a test account with multiple keys), confirm the app self-heals on next launch instead of requiring re-setup, and that a genuinely unmatched key triggers the automatic re-registration path.

Native lockfiles were not regenerated (`ios:setup`/`android:setup`) since only method bodies were added to existing native modules — no dependency or podspec/gradle changes.

Fixes #4166


jleach:
This is what recoverly looks like:

```
Hydrating secure state from native storage...  
console.js:661 [POST] https://idsit.gov.bc.ca/device/token  
console.js:661 [BCSCApiClient] Server Error: Unknown error code (unauthorized) {
  "name": "AppError",
  "technicalMessage": "Signature did not validate for presented JWT authentication.",
  "code": "network.unknown_server_error.2111",
  "timestamp": "2026-07-09T20:55:59.495Z",
  "handled": false,
  "screen": "Device Authentication Prep",
  "url": "/device/token/",
  "method": "POST",
  "cause": {
    "name": "AxiosError",
    "code": "unauthorized",
    "message": "Signature did not validate for presented JWT authentication.",
    "method": "POST",
    "status": 401,
    "url": "https://idsit.gov.bc.ca/device/token/",
    "isTimeout": false,
    "isNetworkError": false,
    "request": {
      "headers": {
        "Accept": "application/json, text/plain, */*",
        "Content-Type": "application/x-www-form-urlencoded",
        "User-Agent": "BCServicesCard/4.0.0 (Android 16; Build 3708)"
      },
      "data": "grant_type=refresh_token&client_id=1b4186bf-0e45-457a-88e2-2fdea20ceaa4&client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer&client_assertion=..."
    },
    "response": {
      "headers": {
        "access-control-allow-origin": "*",
        "cache-control": "no-store",
        "connection": "Keep-Alive",
        "content-type": "application/json",
        "date": "Thu, 09 Jul 2026 20:55:58 GMT",
        "keep-alive": "timeout=5, max=100",
        "pragma": "no-cache",
        "server": "Apache",
        "strict-transport-security": "max-age=31536000; includeSubDomains, max-age=31536000 ; includeSubDomains",
        "transfer-encoding": "chunked",
        "www-authenticate": "Form realm=\"oauth\", error=\"unauthorized\", error_description=\"Signature did not validate for presented JWT authentication.\"",
        "x-content-type-options": "nosniff, nosniff",
        "x-frame-options": "DENY",
        "x-xss-protection": "0"
      },
      "data": {
        "error": "unauthorized",
        "error_description": "Signature did not validate for presented JWT authentication."
      }
    }
  }
} 
console.js:661 [BCSCApiClient] No error handling policy for: {
  "endpoint": "https://idsit.gov.bc.ca/device/token",
  "appEvent": "unknown_server_error",
  "errorCode": "unauthorized"
} 
console.js:661 [hydrateSecureState] Failed to refresh tokens with stored refresh token: Server Error: Unknown error code (unauthorized)  
console.js:661 [hydrateSecureState] Attempting key recovery in case of signing key mismatch...  
console.js:661 [performKeyRecovery] event=triggered client_id=1b4186bf-0e45-457a-88e2-2fdea20ceaa4 — probing server for its registered signing keys  
console.js:661 [GET] https://idsit.gov.bc.ca/device/register/1b4186bf-0e45-457a-88e2-2fdea20ceaa4  
console.js:661 [performKeyRecovery] server returned 1 key(s), 1 with a decodable modulus [wpY81bzd...9G2AtJsU (len=683)] (serverKids=["rsa2"])  
console.js:661 [performKeyRecovery] checking 1 local key(s) ["rsa1"] against the server set for a modulus match  
console.js:661 [performKeyRecovery] checking local key 'rsa1' (modulus=mhFAUh2H...4r0CfnME (len=683)): no match  
console.js:661 [performKeyRecovery] event=failed_no_match no local key modulus matches server jwks (local=["rsa1"], serverKids=["rsa2"])  
console.js:661 [hydrateSecureState] No local key matches server jwks; attempting to re-register the newest local key...  
console.js:661 [PushTokens] Successfully retrieved notification tokens for registration  
console.js:661 [PUT] https://idsit.gov.bc.ca/device/register/1b4186bf-0e45-457a-88e2-2fdea20ceaa4  
console.js:661 [reRegisterNewestKey] event=succeeded re-registered newest local key with server  
console.js:661 [POST] https://idsit.gov.bc.ca/device/token  
console.js:661 [hydrateSecureState] token refresh succeeded after re-registration  
console.js:661 Persisting refresh token  
console.js:661 Persisting registration access token  
console.js:661 Persisting access token  
console.js:661 Tokens persisted to native storage successfully  
console.js:661 Secure state hydrated successfully  
```

KjartanE (approved):
Tested locally, works as expected

Issue #3969, opened 2026-06-04
Enter serial number manually button should be available all the time during camera scan of BCSC 


knguyenBC:
### Description of feature / user story
As a user who is unable to scan my BCSC (accessibility)
I want to be able to enter my serial number manually
So that I can continue the verification process

### Acceptance Criteria
Given the user is prompted to scan a BCSC
When the camera is prompted
Then the "Enter Serial Number Manually" button should immediately be present
And the button design should be the secondary button design on dark background

### Wireframes or relevant image assets / links

TBD

PR #4179, opened 2026-07-08
fix: enter serial manually appears immediately

bryce-mcmath:
# Summary of Changes

Button appears always, not just after the timer. Timer still exists for changing help text

# Testing Instructions

Open ScanSerialScreen, check the button is always there.

# Acceptance Criteria

Button is there from the jump

N/A

# Related Issues

#3969


jleach (approved):
Confirmed on Android. I don't think Copilots comment is correct. This looks correct as per Figma.

PR #4182, opened 2026-07-08
fix: prevent duplicate evidence image on retake causing /documents 400

### LINKED ISSUE #4159: Evidence submission 400 errors during ID verification: diagnosis and improvements
## What

During BC Services Card identity verification, a small number of evidence submissions (ID document photos, liveness videos, and file uploads) are rejected by the server or fail to complete. When this happens the app shows a single generic message — *"Server rejected the request — check request payload"* — regardless of the actual cause, so the user has no idea what to fix and typically retries into the same failure.

We analyzed production server logs (2026-06-28/29) alongside the server source and now have a code-verified explanation for every category of failure. The most actionable finding is app-side: **for several ID types the app submits one more image than the server accepts**, producing an immediate, unrecoverable rejection.

This issue captures the diagnosis and the concrete improvements it points to — an app-side fix, a clearer user-facing error, and an observability ask for the backend team.

## Why

- These failures land **mid-verification**, so an affected user cannot finish onboarding.
- The error message is identical and generic for every cause, so it gives the user no path forward and gives support no signal.
- The image-count case is a **dead end**: retrying re-sends the same bad payload and fails the same way every time (we observed the same devices retrying 3–8×).
- Volumes are low today, but each of these is a lost verification, and the same generic-error pattern hides how often it happens. Fixing the image-count case removes a whole class of unrecoverable failures, and clearer errors + tracking protect the verification completion rate going forward.

## Technical details

The server code explains why the video/upload reasons were missing from the logs and lets us name each cause.

### The key mechanism (server `GlobalExceptionHandler`)

Every 400 is logged at `logger.debug` — and prod runs at INFO — so **the 400 reason is invisible unless the service method logged an INFO line before throwing.** That is the whole reason `/documents` had a visible reason and the others did not. The 400-mapped exceptions are `InvalidParameterException` and `MessageDigestMismatchException`; the response body is `{"message":"<msg>"}`.

(For reference: `ResourceNotFoundException`→404, `UnauthorizedException`→401, `AlreadyRegisteredException`→409, and the catch-all `Exception`→500 is the only one logged at ERROR — which is why the DB/ClientAbort 500s were the loudest thing in the logs.)

### Each evidence 400, explained

| Endpoint | Both-day total | Root cause | Confidence |
|---|---|---|---|
| `POST /documents` | 48 | `validateEvidenceImages` → **"unexpected images count"**: app sent the wrong number of images for the doc type | Verified — INFO-logged, 12/12 exact match on 06-29 |
| `POST /videos` | 62 | `assertEvidenceVideo` rejects video **metadata**: bad content-type / duration out of min–max range / stale capture date / liveness-prompt errors | Cause set narrowed from code; exact reason not loggable (all DEBUG) |
| `PUT /uploads/{ref}` | 56 | **Interrupted binary upload** — `ClientAbortException`/`EOFException` reading the body → 400 | Verified — 22/22 match a ClientAbortException at same node+second |
| `PUT /verifications` | 7 | `VerificationAPI` param validation (id / upload_uris / card type / sha256) | From code; 06-28 only, no app logs that day |

Two things the code proved that the logs alone could not:

1. **All 48 `/documents` 400s (both days) are the same error.** Every one returned an identical 48-byte body ⇒ same message ⇒ the 36 on 06-28 (no app logs) are *also* "unexpected images count," not just probably.
2. **`/uploads` and `/videos` are fundamentally different failures.** Uploads are a **transport** problem (dropped connection mid-PUT of a multi-MB image/video), not a bad payload — digest/size/MIME mismatch and session-not-found (a 404) were ruled out. Videos are a genuine **metadata validation** failure (the request reaches the controller), just DEBUG-logged so the specific field is invisible.

### The image-count pattern (the clearest app-side bug)

Every observed `/documents` rejection was the app sending **exactly one extra image** for the type:

| ID type | Server expects | App sent |
|---|---|---|
| Foreign Passport | 1 | 2 |
| Canadian Passport | 1 | 2 |
| BC Drivers Licence | 2 | 3 |

### Uploads are low-rate network noise, not a payload bug

Over both days: **7864 successful uploads (200)** vs **56 aborts (400)** + **21 timeouts (408)** ≈ **0.7% abort rate**, spread across **32 distinct devices** (iOS + Android), each abort occurring **7s–2.7min** into the transfer. That is the signature of interrupted large uploads over mobile networks — not the app deliberately stopping or sending a truncated body.

### App-side improvements this points to

1. **Fix the extra-image submission** for the affected doc types (Foreign Passport, Canadian Passport, BC Drivers Licence) — the app is including one image beyond what the type accepts.
2. **Make the 400 error actionable.** The app currently collapses every 400 to `ERR_209_BAD_REQUEST` / "Server rejected the request — check request payload" and ignores the server response body. For the image-count case the server returns a specific `{"message":"unexpected images count"}` we could translate into real guidance (e.g., "Remove the extra photo and try again"). At minimum, log/telemeter the server message so these stop being invisible.
3. **Consider a retry/telemetry path for interrupted uploads** so a dropped multi-MB upload can resume or auto-retry instead of surfacing as a hard failure.

### Cross-team ask for the backend (IAS)

The reason these were hard to diagnose is `GlobalExceptionHandler` logging all 400s at `debug`. **Bumping the `InvalidParameterException` handler to `info`/`warn` (including `e.getMessage()`)** would surface the video and verification 400 reasons in prod with zero guesswork. (This is a server change in the IAS repo; noted here so the mobile fix and the server observability change stay linked.)

### Open questions

- The `evidence.video.duration.seconds.{min,max}` limits are externalized to deployment config (not in the repo). Worth confirming against the app's ~19s videos, since duration is one of the possible `/videos` rejection reasons.
- The 06-28 video/upload/verification reasons could not be confirmed from logs (no app-level logs retained for that day); causes above are code-derived.

_Source: analysis of IAS production access + application logs (2026-06-28/29) cross-referenced with the `device-parent/evidence` module. No user data is included in this issue._

jleach:
## What

A subset of ID verification retries were hitting an unrecoverable "server rejected the request" error partway through, with no way for the user to get past it. This happened when someone went back to retake or re-confirm a photo of their ID (e.g. their driver's licence or passport) after already finishing that step — the app kept the old photo *and* the new one instead of replacing it, then tried to upload both. The server rejects that as an invalid submission, and because the extra photo was saved to the device, every subsequent retry hit the same wall.

This fix stops the duplicate from being created, and also detects and cleans up the extra photo for anyone who's already stuck in this state, so they can complete verification without reinstalling or clearing data.

## Why

This is one of the confirmed causes behind the evidence-submission 400 errors described in #4159 — specifically the "off-by-one" pattern where the app sends one more image than the ID type expects (e.g. a driver's licence upload with 3 images instead of 2). It's a dead end for the affected user with no self-service recovery, so fixing both the cause and the already-affected cases matters.

## Decisions

- **Root cause**: `EvidenceCaptureScreen`'s accept handler appended the newly captured photo to the in-memory array by index-less push. Navigating back from Evidence ID Collection to retake/re-accept a side leaves the screen mounted with stale state, so a second accept/retake on the same side appended a duplicate instead of overwriting it. Fixed by keying the write to the current side's index, so re-accepting or retaking a side always replaces that side's entry. Back navigation and retake itself are unchanged — legitimate UX — only the data write is fixed.
- **Healing already-corrupted state**: a corrupted n+1 array, once persisted, previously had no recovery path — it still passed the existing completeness check and was uploaded as-is on every retry. Added `clampEvidenceImagesToSides` (in `card-utils.ts`), a read-side dedupe applied in `useEvidenceUpload`'s upload path that trims metadata down to the card's expected number of sides, keeping the most recently captured photo per side. It doesn't rewrite persisted storage — it only affects what gets sent to the server — so it's a low-risk, defense-in-depth safety net rather than a data migration.
- Consolidated the `front`/`back` legacy label normalization (previously a private map duplicated in `useEvidenceUpload`) into one exported helper in `card-utils.ts`, and fixed a latent side-effect of the same bug where a duplicate label could cause the same binary file to be uploaded twice.
- Out of scope, deferred to a follow-up: user-facing error copy for this failure (currently a generic message; the server's response is already remote-logged), and the other 400 sources noted in #4159 (video/upload/verification endpoints).

## Manual Testing

Do a send-video submission, retake the selfie 2 or more times; retake the video 2 or more times. Submit. The server will always get the last selfie and last video.

## Tests

- Extended `EvidenceCaptureScreen.flow.test.tsx` with two regression tests that reproduce the exact reported patterns (re-accepting a single-sided card's only photo; retaking a two-sided card's second photo) and assert the persisted metadata never exceeds the expected count. Verified these tests fail against the pre-fix code (2 and 3 entries respectively) and pass with the fix.
- Extended `card-utils.test.ts` with cases for the new `clampEvidenceImagesToSides`/`normalizeEvidenceImageLabel` helpers (over-count dedupe, exact-count passthrough, legacy label mixing, missing/empty side data, all-distinct over-count slicing).
- Added `useEvidenceUpload.test.tsx` covering the upload path end-to-end: a corrupted 3-entry driver's licence record heals to 2 images with a warning log, and a healthy 2-of-2 record passes through unchanged with no warning.
- Full `yarn check` suite run: typecheck, TS lint (app + bcsc-core), Prettier format-check, and the full Jest suite (257 suites / 2383 tests) all pass with no snapshot changes. The Swift lint step reports pre-existing issues in `packages/bcsc-core/ios` files untouched by this change (not part of this diff).

Fixes #4159


bryce-mcmath (dismissed):
Not blocking, but I don't know if evidence helper methods belong in card-utils. Still, approved

bryce-mcmath (approved):

Issue #3293, opened 2026-02-19
Critical TODOs and FIXMEs


bryce-mcmath:
### Description

List of TODOs and FIXMEs that must be addressed before release, organized by version target. Items verified against current source on 2026-03-28.

## Version 4.0 — Phase 1 (BCSC Related)

| # | File:Line | Description | Issue | Owner | Notes |
|---|-----------|-------------|-------|-------|-------|
| 1 | `service-hours-formatter.ts:190` | Silently fails on malformed API response | #3573 | — | |
| 2 | `LiveCallScreen.tsx:69` | API throws unhandled error when user ineligible for live call | #3582 | Bryce | |
| 3 | `bcsc-credential.ts:27` | Hardcoded `bcscReason` is not a valid value per spec | #3588 | Mac | |
| 4 | `navigators.ts:55,61` | `CancelledReview` & `DualIdentificationRequired` screens have no V3 mapping — verify Snowplow tracking | #3589 | Mac | |

## Version 4.1 — Phase 2 (Credential Related)

| # | File:Line | Description | Issue | Owner | Notes |
|---|-----------|-------------|-------|-------|-------|
| 1 | `attestation.ts:36` | Untyped subscription; Credo doesn't export event subscription type | — | — | |
| 2 | `attestation.ts:339` | Missing handling for credential offer completing during attestation | — | — | |
| 3 | `attestation.ts:663` | No validation that credential attribute count matches expected schema | — | Jason | |
| 4 | `client.ts:329` | Business logic mixed into API client; should be separated | #3431 | — | |
| 5 | `PersonCredentialLoading.tsx:176` | No error handling when BCSC auth fails during person credential loading | — | Jason | |
| 6 | `BCIDHelper.ts:17` | Hardcoded `_internal/legacyDid` because AFJ doesn't export it | — | Jason | |
| 7 | `bc-agent-modules.ts:132` | Credo version mismatch (0.5.15 vs 0.5.19) causes suppressed type error | — | Bryce | |
| 8 | `bc-agent-modules.ts:141` | Same Credo version mismatch type error in another module registration | — | Bryce | |
| 9 | `credentials.ts:131` | AFJ bug: credential format key with `undefined` value still gets processed | — | — | |
| 10 | `platform-context-retriever.ts:30` | Some platform context retrieval methods stubbed out | #3541 | Mac | |
| 11 | `navigators.ts:64` | One V4 screen maps to multiple V3 screens | #3409 | Mac | |
[...]

PR #4216, opened 2026-07-10
refactor: extract environment and analytics helpers from store

### LINKED ISSUE #3583: Extract environment and analytics helpers from store.tsx
## Description

`store.tsx` is ~790 lines and contains environment/analytics utility functions that don't belong in the store. These helpers (around line 291) should be extracted to a separate file to keep the store focused on state management.

## Reference

Originally tracked in #3293

jleach:
## What

`store.tsx` (state management for the app) had grown to include a chunk of
IAS environment/analytics configuration logic that doesn't belong there.
This PR moves that logic into its own file, and moves the small `Mode`
(BC Wallet vs. BC Services Card) enum into the app's shared constants file.
No behavior changes — the app works exactly the same before and after.

## Why

`store.tsx` was ~840 lines and mixing state-management concerns with
environment/analytics configuration made it harder to find and maintain
either one. This was already flagged with a `TODO` in the code and tracked
back to #3293. Splitting it out keeps the store focused on state and makes
the environment configuration easier to find and test on its own.

## Decisions

- New `app/src/utils/environment.ts` holds `IASEnvironment` (interface +
  const) and `getInitialEnvironment()`, plus the module-private
  `getAnalyticsAppId`/`createIASEnvironment` helpers — moved verbatim from
  `store.tsx`. It imports from `@/constants` only, never from `@/store`,
  since `store.tsx` calls `getInitialEnvironment()` at module scope
  (`developerState`) and a `store → environment → store` import cycle would
  crash on load.
- Per direction from the issue owner, `Mode` (`BCWallet` / `BCSC`) landed in
  `app/src/constants.ts` rather than in the new `environment.ts` file, since
  it's a general app-mode concept rather than an IAS-environment concept.
- Clean cut, no compatibility re-exports: every consumer of `IASEnvironment`,
  `getInitialEnvironment`, or `Mode` was retargeted to import from the new
  locations directly instead of via `store.tsx`.
- Found two additional call sites outside `app/src/` (`app/App.tsx`,
  `app/container-imp.ts`) that also needed retargeting — caught by
  `yarn typecheck`, not part of the original file list.
- One test (`MainStack.test.tsx`) hand-mocks `@/constants` with a partial
  stub; it needed `Mode` added to that stub now that `store.tsx` resolves
  `Mode` through `@/constants` transitively.

## Tests

- New `app/src/utils/environment.test.ts`: the existing `getInitialEnvironment`
  coverage moved over unchanged, plus new assertions locking in
  `IASEnvironment`'s computed `analyticsAppId`/`iasApiBaseUrl` values.
- `store.test.ts` slimmed to just the reducer test (the environment tests
  moved out with the code they were testing).
- Full suite: `yarn typecheck`, `yarn lint:ts`, `yarn format:ts:check`, and
  `yarn test` (264/264 suites, 2444/2445 tests, 1 pre-existing `todo`) all
  pass. No snapshots changed.

Fixes #3583


al-rosenthal (approved):
ran locally and was able to verify and go through wallet features without issue

Issue #4114, opened 2026-06-30
Update app icon for 4.1.x


cvarjao:
### Background
App icon has been reviewed by GCPE and changes were made. 
The app and app stores will need to be updated with the new app icon.

### Design
download app icon for app stores [here](https://bcgov-my.sharepoint.com/:f:/r/personal/kimberly_nguyen_gov_bc_ca/Documents/Design/BCSC%20app%20icon/BCSC%204.1%20app%20icon_appstores?csf=1&web=1&e=yhK3E7): 

Affected locations:
- App stores
- [Splash and unlock screen ](https://www.figma.com/design/Zbf4Z4BF8yWCTEKGkkVvMW/BCSC-Phase-2---Android-iPhone?node-id=4161-8134&t=7Ai1JuxsUzSO1fpo-4)

jleach:
## Implementation plan — BCSC v4.1 app icon

Source assets are already staged, verified, and corrected in the `bc-wallet-mobile-icons` working dir (1024² per platform; prod bg `#013366`; iOS light opaque / dark transparent / tinted grayscale; Android foreground + mono). Decisions locked in: prod ships the **full** set (light+dark+tinted, adaptive fg/bg/mono), undoing the #4028 prod v3 revert.

**Target:** `bcsc` build target only (all four variants: dev, qa, test, prod). BC Wallet is untouched — the change lives entirely in `variants/**` (+ an optional generation script in `scripts/`), so the committed default tree and bcwallet builds are unaffected.

### Approach

Replace the icon *bytes* in the existing overlay structure (`variants/_bcsc-base` + `variants/bcsc-{dev,qa,test}`) with the GCPE-approved v4.1 artwork, and structurally un-revert `variants/bcsc-prod` (PR #4028, commit `cd6e1608d`) so prod rejoins the adaptive/full-appearance scheme with a solid `#013366` background. All targets are generated from the verified 1024×1024 sources via a small Python/Pillow script — only downscaling, compositing, masking, and alpha-flattening are needed (sources already exceed every target size). Android boot splash (`launch_screen.xml`, `android:src="@mipmap/ic_launcher"`) inherits the new icon automatically; iOS `LaunchScreen.storyboard` has no icon. The in-app splash/unlock mark (`BCAnimatedLoadingIcon.tsx`, a hand-coded animated SVG of the *old* circular mark) is a code redesign from Figma node 4161-8134 and is **deferred to PR 2 of the stack**.

### Steps

[...]

Issue #3468, opened 2026-03-17
Email missing in Account screen info right after verification


bryce-mcmath:
### Description of problem
The email field on the Account screen is empty after first verification, but appears after subsequent restarts

### Expected behavior
Shows up initially

### Steps to reproduce
Verify and go to account tab

### Environment

Build #:

Android Device Model:

iOS Device Model:

bryce-mcmath:
5305

PR #4230, opened 2026-07-11
fix: use BCSC notification icon for BCSC build target

### LINKED ISSUE #4221: Indicator icon for the BCSC app is still the BC Wallet
### Description of problem
The indicator icon (the tiny icon that appears in the phone's notification bar) for the BCSC app is still the BC Wallet icon. 

### Expected behavior
The indicator icon should be the black & white version of the BCSC app.

### Steps to reproduce
Receive a notification from the BCSC app and check the phones notification bar at the top

Image

### Environment

Build #: 4.0.2 (6447)

Android Device Model:

iOS Device Model:

jleach:
## What

The small icon shown in the phone's status bar / notification shade for BCSC push notifications was still the BC Wallet icon, even on a BC Services Card build. This change swaps that icon for a white silhouette of the BCSC ID-card glyph, matching the app it's actually notifying from. BC Wallet's own notification icon is unchanged.

## Why

Users receiving a BCSC notification saw the wrong brand in the status bar, which is confusing and looks like a bug (because it is one) — see #4221.

## Decisions

- Asset-only, Android-only change; no Kotlin/manifest/native code touched. Both places that reference the icon by name — the FCM `default_notification_icon` meta-data in `AndroidManifest.xml` and `BcscCoreModule.kt`'s `showLocalNotification` — resolve `ic_notification` by resource name, so swapping the drawable bytes fixes both call sites.
- iOS is out of scope: iOS has no separate "small icon" concept for notifications — it shows the app icon, which is already correct per build target.
- Generated the new glyph from the GCPE-approved `android_foreground_tinted.png` source (issue #4114's art drop) via `scripts/icons/generate-bcsc-icons.py`, extended to also emit `ic_notification.png` at the 5 Android densities into `variants/_bcsc-base/overlay` (shared by all four `bcsc-*` variants via `BASE_VARIANT`), rather than hand-exporting one-off PNGs. Recolours the source's dark-grey opaque fill to pure white while preserving alpha *before* the final downscale, so no grey halo bleeds into the anti-aliased edges — Android supplies the actual colour via notification icon theming, so the asset itself must be a flat white silhouette.
- Sized the glyph near-full-bleed (~2dp padding in a 24dp frame, ~83% of the canvas) per Android's small-icon guidance — tighter than the ~44% used for the adaptive launcher foreground, since this one isn't circle-masked. Flagged as a QA-visual-check item: if it reads a little small/large on an actual device, it's a one-constant tweak (`NOTIFICATION_TARGET_PX` in the generator script).
- Left `default_notification_color` (`@color/primary`, `#003468`) unchanged — out of scope for this fix, no complaint about the tint colour, only the glyph shape.
- Preserved the current BC Wallet notification-icon bytes verbatim in a new `variants/bcwallet-prod/overlay/.../drawable-*/ic_notification.png` (byte-for-byte identical to what's on `main` today, confirmed via `shasum`) so BC Wallet builds are completely unaffected.

## Tests

Extended the generator script's built-in invariant checks with an `ic_notification.png` case (mode RGBA, alpha spans fully-transparent to fully-opaque, every non-transparent pixel pure white `(255,255,255)`), and ran it twice back-to-back to confirm the output is byte-identical (idempotent). Also did a manual pixel-level pass on the generated assets — composited the xxxhdpi glyph over a dark background and visually confirmed a clean white BCSC ID-card silhouette with no grey/dark fringing at the edges.

## Verification

- `shasum -a 256` of the 5 new `bcwallet-prod` overlay PNGs against the currently-committed BC Wallet PNGs on `main` — exact match, so BC Wallet is byte-for-byte unaffected.
- Full variant round-trip in an isolated worktree: applied `bcsc-dev` (only the 5 `ic_notification.png` + the script diff show up), applied it again (identical diff, confirms idempotency), switched to `bcwallet-prod` (wallet mark bytes restored, hash-verified), switched back to `bcsc-dev` (clean return to the same expected diff — no other file left touched).
- `yarn check`'s TS/lint/format/test-relevant steps (typecheck, eslint, ktlint, prettier format check, jest) all pass. The Swift lint/format-check steps fail locally on 3 pre-existing files in `packages/bcsc-core/ios` unrelated to this diff (no Swift touched) — a known local-environment false-fail, not from this change.
- Still needed: an actual on-device check with a native BCSC Android build and a real push notification, to confirm the glyph reads clearly at real status-bar size and the ~2dp padding looks right. The shipped 4.0.2 build can't be hot-patched with an asset change — this rides the next release.

Fixes #4221


jleach:
Some old icons were used for the BC Wallet which I believe retains the current icon set.

KjartanE (approved):
tested on iOS, working as expected

PR #4261, opened 2026-07-16
fix: send video prompts

KjartanE:
# Summary of Changes

IAS rotates the selfie-video prompts — and the request sha256 they're bound to — on every fetch, so a recording is only accepted against the set the server just issued. Retakes and back-navigation could send the user to the camera with a stale prompt set, letting one recording answer the same challenge repeatedly.

Prompt fetching is centralized in a new `useVideoPrompts` hook, called from every gateway into recording. Back-navigation out of the review and too-long screens no longer lands on the armed camera with stale prompts: VideoReview redirects back to the instructions screen (which refetches), and VideoTooLong blocks gesture/hardware back so its Retake/Cancel buttons are the only exits.

Also fixes several edge cases surfaced in review: the rotated sha is no longer committed alongside an empty prompt set, concurrent refreshes are de-duplicated, a stale request id is only recreated on a 500 (not on any error), and a failed fetch now shows a Retry instead of an endless spinner.

# Testing Instructions

1. Start selfie-video verification and record a video.
2. On the review screen, tap **Retake** → confirm you return to the camera with a fresh set of prompts (not the previous ones).
3. From the review screen, use the header back button and Android hardware back → confirm both land on the instructions screen and it refetches before recording.
4. Record a video longer than the max duration → on the "too long" screen, confirm back gestures are blocked and only Retake/Cancel work.
5. With the network off, arrive at the instructions screen → confirm a Retry button appears (not a stuck spinner) and that retrying recovers once the network is back.

# Acceptance Criteria

- Every path into recording issues a fresh prompt set; no path reaches the camera with a stale one.
- Back-navigation from the review and too-long screens cannot reach the armed camera with stale prompts.
- A failed prompt fetch is recoverable in-place (Retry), not a dead end.
- An empty prompt set never leaves the store with a mismatched sha/prompts pair.

N/A

# Related Issues

N/A

### DIFF
--- a/app/src/bcsc-theme/features/verify/_models/useVerificationMethodModel.test.ts
+++ b/app/src/bcsc-theme/features/verify/_models/useVerificationMethodModel.test.ts
@@ -194,7 +194,8 @@ describe('useVerificationMethodModel', () => {
         await result.current.handlePressSendVideo()
       })
 
-      expect(mockLogger.error).toHaveBeenCalledWith('Error sending video:', mockError)
+      expect(mockLogger.error).toHaveBeenCalledWith(expect.stringContaining('No verification request available'))
+      expect(mockVideoPromptsMissingAlert).toHaveBeenCalledTimes(1)
       expect(mockNavigation.navigate).not.toHaveBeenCalled()
       expect(mockDispatch).not.toHaveBeenCalled()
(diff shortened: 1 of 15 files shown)


jleach (approved):
Confirmed on Android. New prompts on retry.

Issue #4258, opened 2026-07-16
Nickname changes to "BC Services Card"


RaulConstan:
### Description of problem
The nickname changed from "Combo" to "BC Services Card" without my intervention.

### Expected behavior
Nickname does not change without user input.

### Steps to reproduce
1. Created a non-BCSC flow to match to DL (Data: DL = 55667788, Surname = RC0000080, Given = Combo, DoB = 2000/01/01)
2. Approved it
3. Checked Settings for the Nickname to be Combo
4. Lock the phone (to trigger the sign-in)
5. Unlock the app
6. Check the Nickname 
Expected:
No change
Actual:
Changed to "BC Services Card"

### Environment
SIT

Build #:
7389

Android Device Model:
Pixel 6 (Android 16)

iOS Device Model:

jleach:
# Plan: Fix nickname reverting to "BC Services Card" (#4258)

## Summary

Target: **bcsc** build target only (`app/src/bcsc-theme/` + `app/src/store.tsx` — no changes to `bcwallet-theme`, no native code, no `packages/bcsc-core` changes).

One PR, three parts:

1. **Fix #4258** — restore `selectedNickname` into the app store from the native account record (secure storage, which survives process death) when the AsyncStorage copy is empty, in `useInitializeAccountStatus`.
2. **Shared name helpers** — add `getShortDisplayName` (given→family mononym rule) and `getFullDisplayName` (mononym-safe full name join) to `account-utils.ts`.
3. **Refactor/fix call sites** — `useVerificationResponseViewModel` uses `getShortDisplayName`; `tokenToCredentialMetadata` in `BCSCIdTokenContext` uses `getFullDisplayName`, fixing the `"undefined Smith"` mononym bug.

## Root cause

Verified in code at main HEAD (`05c483a57`):

- "BC Services Card" is never stored as a nickname. It is the display fallback in `app/src/bcsc-theme/features/settings/SettingsContent.tsx:274` — `store.bcsc.selectedNickname?.trim() || t('BCSC.Title')`. The underlying event is `selectedNickname` becoming empty.
- The nickname exists in two places: `store.bcsc.selectedNickname` (persisted to AsyncStorage by the `UPDATE_NICKNAME` reducer case, `app/src/store.tsx:433-439`) and the native account record's `nickname` (secure storage, written by `setAccount` in `updateRegistration`, `app/src/bcsc-theme/api/hooks/useRegistrationApi.tsx:356-362`).
[...]

PR #4271, opened 2026-07-16
fix: restore nickname from account record and handle mononyms consistently

jleach:
## What

When a user with a mononym-style BC Services Card account (a name stored only in the "surname" field, like a business name — e.g. "Combo") locked their phone and came back into the app, their account nickname could silently revert to the generic "BC Services Card" label instead of staying as the name they'd set. This restores the correct nickname automatically when that happens, and closes a related gap where mononym accounts could get a broken "undefined \<name\>" value stored internally.

## Why

On Android, locking the phone can kill the app process. When the app restarts, its saved settings are rebuilt from local storage — but the nickname wasn't being restored from the durable native account record, so if that particular save was lost, the app fell back to a generic placeholder instead of the user's real nickname. This also surfaced a second, related bug: accounts with only one name on file (no separate given name) could end up with a corrupted "undefined \<name\>" value stored internally, which risked triggering an unnecessary "your account was updated" alert once the underlying bug was fixed.

## Decisions

- **Self-healing restore, not a persistence rewrite**: rather than chasing every possible race in how the nickname gets saved to local storage, the app now restores the nickname from the native secure-storage account record (which survives process death) on startup whenever the store's nickname is empty. This repairs the symptom regardless of which specific write was lost.
- **Shared name helpers**: added `getShortDisplayName` (nickname default: given name, falling back to the stored surname for mononym accounts) and `getFullDisplayName` (mononym-safe full name join) to `account-utils.ts`, reused in both the post-verification nickname flow and the credential metadata used for account-change alerts — replacing two separate ad hoc implementations that disagreed slightly on mononym handling.
- **One-time alert suppression for existing mononym users**: the old buggy template stored values like `"undefined Smith"` for existing mononym accounts. Fixing the template alone would make that stored value look different from the newly-computed correct value on next login, firing a spurious "your account was updated" alert. `compareCredentialMetadata` now normalizes away that specific legacy artifact (a leading `"undefined "` token, plus stray whitespace) before comparing, so existing mononym users don't see that one-time false alert. Genuinely different names still compare as different.
- Scope: `bcsc` build target only — no changes to `bcwallet-theme`, no native/platform code, no `packages/bcsc-core` changes.

## Tests

- `account-utils.test.ts`: unit tests for `getShortDisplayName`/`getFullDisplayName` covering two-name, mononym (given-name missing/blank), and empty inputs, including the issue's own repro values (`given: 'Combo'`, `family: 'RC0000080'`).
- `useInitializeAccountStatus.test.tsx`: covers the acceptance criterion directly — `hasAccount: true` with an empty nickname and a native account record restores the nickname without touching `hasAccount`; an already-complete store skips the native call entirely; a missing native account doesn't clobber or downgrade anything; an existing user-set nickname is never overwritten by a different native value.
- `useVerificationResponseViewModel.test.ts`: added mononym (family_name-only) and given-name-preferred cases for the post-verification nickname.
- New `BCSCIdTokenContext.test.ts`: pins `tokenToCredentialMetadata`'s mononym-safe `fullName` output, and proves the legacy `"undefined Smith"` vs corrected `"Smith"` case compares equal while genuinely different names still compare unequal.
- Full suite: `yarn typecheck`, TS lint/format (app + bcsc-core), and `yarn test` (275 suites / 2687 tests / 158 snapshots) all pass.

## Verification

1. Repro the original issue's steps (non-BCSC flow matched to a DL with a mononym name, e.g. Given = Combo), confirm the nickname is set correctly, lock the phone until Android kills the app, reopen, and confirm the nickname is unchanged instead of reverting to "BC Services Card".
2. For an existing account that already has the legacy `"undefined <name>"` artifact stored, confirm no "your account was updated" alert appears purely from this fix.

#4258

### DIFF
--- a/app/src/bcsc-theme/contexts/BCSCIdTokenContext.tsx
+++ b/app/src/bcsc-theme/contexts/BCSCIdTokenContext.tsx
@@ -44,8 +44,8 @@ export const tokenToCredentialMetadata = (token: IdToken): CredentialMetadata =>
  * {@link getFullDisplayName} output, avoiding a one-time false-positive
  * "account updated" alert for existing mononym users. See #4258.
  *
- * Narrowly scoped to leading "undefined" artifacts by design (matches the
- * literal lowercase string JS produces when interpolating `undefined`); a
+ * Narrowly scoped to leading/trailing "undefined" artifacts by design (matches
+ * the literal lowercase string JS produces when interpolating `undefined`); a
  * genuine name that happens to start with capitalized "Undefined" is safe,
  * since the match is case-sensitive. See BCSCIdTokenContext.test.ts for the
  * documented tradeoff around a literal lowercase "undefined" name part.
@@ -63,8 +63,10 @@ const normalizeLegacyFullName = (fullName: string | undefined): string => {
     return ''
   }
 
-  // Only one part (given_name) was absent, e.g. "undefined Smith" -> "Smith".
-  return collapsed.replace(/^undefined\s+/, '')
+  // Only one part was absent: given_name ("undefined Smith" -> "Smith") or
+  // family_name ("Jamie undefined" -> "Jamie") — the legacy template had
+  // exactly two slots, so leading + trailing strips cover the remaining artifacts.
+  return collapsed.replace(/^undefined\s+/, '').replace(/\s+undefined$/, '')
 }
 
 /**
--- a/app/src/bcsc-theme/contexts/BCSCIdTokenContext.tsx
+++ b/app/src/bcsc-theme/contexts/BCSCIdTokenContext.tsx
@@ -38,19 +38,34 @@ export const tokenToCredentialMetadata = (token: IdToken): CredentialMetadata =>
  * Normalizes a fullName value that may have been produced by the legacy
  * `${given_name} ${family_name}` template, which rendered a literal
  * "undefined " prefix (and/or stray whitespace) for mononym users whose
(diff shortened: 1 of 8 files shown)


app/src/bcsc-theme/contexts/BCSCIdTokenContext.tsx
```diff
@@ -33,10 +34,46 @@ export const tokenToCredentialMetadata = (token: IdToken): CredentialMetadata =>
...
+
+  // Both given_name and family_name were absent: the legacy template rendered this
+  // as "undefined undefined" (repeated once per absent part), which carries no real
+  // name text. Normalize it the same as getFullDisplayName({})'s '' output.
+  if (/^(undefined\s*)+$/.test(collapsed)) {
+    return ''
+  }
+
+  // Only one part (given_name) was absent, e.g. "undefined Smith" -> "Smith".
+  return collapsed.replace(/^undefined\s+/, '')
```
KjartanE: normalizeLegacyFullName handles leading undefined but not trailing
One-line fix at BCSCIdTokenContext.tsx:67:
`return collapsed.replace(/^undefined\s+/, '').replace(/\s+undefined$/, '')`
Plus a 'Jamie undefined' vs 'Jamie' test case.

KjartanE (approved):

app/src/bcsc-theme/contexts/BCSCIdTokenContext.tsx
jleach: Fixed.

PR #4280, opened 2026-07-17
chore: enhance error reporting

MacDeluca:
# Summary of Changes
  This PR standardizes error reporting on the app's AppError type, and merges separate screen/url/method fields into a single context object (addContext()).

  - ErrorModalPayload now carries the full AppError instead of duplicating its fields.
  - reportProblem() takes a typed object with an optional AppError, and logs a device install_id.
  - The error modal and report flows no longer hand-build report strings — they pass the AppError straight through and let toJSON() serialize it.

  Net effect: one error type flows end-to-end from modal to logger to Loki, instead of being re-translated at each step.
# Testing Instructions

# Acceptance Criteria

# Related Issues


app/src/constants.ts
```diff
@@ -188,6 +188,7 @@ export const VERIFY_DEVICE_ASSERTION_PATH = 'v3/mobile/assertion'
 // Error constants
 export const UNKNOWN_APP_ERROR_STATUS_CODE = 9999
 export const FATAL_UNRECOVERABLE_ERROR_STATUS_CODE = 9998
+export const USER_REPORT_ERROR_CODE = 0
```
jleach, line 191: I might have done 9997 since we seem to be stacking our catch all in the 99xx range. But not a blocker for sure.

jleach (dismissed):

app/src/constants.ts
MacDeluca: I agree, I left it the same as before (0) to prevent any Grafana dashboard regressions.

I can update it if we don't think that's a concern?

jleach (dismissed):

jleach (approved):

Issue #4273, opened 2026-07-17
Error Code: 2822 no account found for the current user when transferring account from another device


RaulConstan:
### Description of problem
Bare with me on this one... I think the steps are correct, but take them with a grain of salt.
The issue happened on the Android after I scanned the QR code from the iPhone.  Clicking Continue button showed the error.  It is intermittent (not able to repro anymore), and non-blocking (I think).

### Expected behavior
No errors

### Steps to reproduce
1. Source QR setup (RC0000080/ Jan 1, 2000)
2. Target scans QR Code
3. Target clicks Continue

### Screenshots and/or log output
The video does not show the error, just the remote id (maybe something can be mined):

https://github.com/user-attachments/assets/b24c10a2-84ed-40f0-a804-08262da00942

The error screen:

<img width="225" height="564" alt="Image" src="https://github.com/user-attachments/assets/f8bc27b5-58f5-4825-8f52-56f780fed62c" />

### Environment
SIT

Build #:
Source (7411) - Target (7389)

Android Device Model:
Pixel 6 (Android 16)

iOS Device Model:
14 Plus (26.3.1)

jleach:
Talked on refinement today. Will monitor this issue and escalate if we can reproduce it.

gvodden:
Reproduced it but the path was slightly different.
Transferred from Android to iPhone
Tapped Remove Account, and the error popped up.
Reported Problem: RP3B-9BQE

Seems like the app is checking the account after it has been removed?

jleach:
From reported Grafana logs. Probably a race condition? Something deleted data, while something is trying to use the data?

```
2026-07-21 09:16:21.961 ERROR {
  "message": "Something went wrong",
  "data": {
    "description": "We're having trouble with the app. Please try again later.\n\nIf the issue persists, close and re-open the app, then try again.",
    "code": 2822,
    "message": "No account found for the current user — state may be corrupted or missing\nDebug: [general.account_not_found.2822]\nScreen: Intro\nReport ID: 952e1395-92c9-4fbe-a912-5a6105274b2b",
[...]

PR #4309, opened 2026-07-21
fix: prevent 2822 account-not-found alert on source device after account removal post-transfer

jleach:
## What

When transferring an account to a new device, the source device shows a QR code while the transfer completes in the background. In one intermittent case, after the transfer finished and the user removed the account from the source device (via "Remove Account"), a critical "account not found" error popped up — even though everything had actually worked correctly. This was confusing and alarming for users, since it looked like a real problem right after they'd just successfully finished the transfer.

## Why

The QR-display screen refreshes its QR code roughly every 50 seconds by checking the account and generating a new signed token. If the screen was torn down (navigated away from, or the account was removed) while one of those checks was still in progress, the check could finish *after* teardown and — in rare timing windows — end up installing a new background timer that nothing was left to clean up. That orphaned timer kept ticking every 50 seconds; once the account was gone, its next check found nothing and raised the critical error from whatever screen happened to be showing at the time (typically the app's intro screen). This was reported in Grafana as error code 2822 and confirmed against the user's report (RP3B-9BQE).

## Decisions

- Root cause confirmed: `accountNotFoundAlert()` in `TransferQRDisplayScreen`'s `createToken()` is the only app-wide trigger for error 2822. The 50s QR-refresh interval could leak past unmount when an in-flight `createToken()` resolved after the screen's cleanup ran, because `refreshToken()`'s post-`createToken` check only guarded on `completedRef`, not on whether the screen was still mounted.
- Fix: added an `isMountedRef` latch, checked (alongside the existing `completedRef` transfer-completed latch) at the top of `createToken`, again immediately after the `await getAccount()` suspension point (state can change while that await is pending), in `refreshToken`'s post-`createToken` restart decision, and in the QR-refresh interval's tick callback.
- A `null` account seen after unmount or after the transfer already completed is now treated as an expected, idempotent no-op (logged at `info`), not an error. A `null` account seen while the screen is actively mounted and displaying the QR code still raises the alert unchanged — that case is a genuine problem and needed to keep working exactly as before.
- Scope was kept to the transferer (source device) QR screen only, per the plan — no changes to alerting/error-registry code, i18n, native code, or the transferee side. No render-output changes (confirmed no snapshot diff).

## Tests

Added three regression tests to `TransferQRDisplayScreen.test.tsx` using a deferred (manually-resolved) `getAccount` mock to simulate the race:
- An in-flight `createToken()` that resolves successfully *after* unmount no longer installs an orphaned refresh interval (`getAccount` called exactly once total, even after advancing fake timers 150s).
- An in-flight `createToken()` that resolves with `null` *after* unmount does not raise the alert.
- An in-flight `createToken()` that resolves with `null` *after* the transfer has already completed (attestation confirmed, navigated away) does not raise the alert.

Before committing these, I confirmed each one actually fails against the pre-fix code (e.g., `getAccount` called 4 times instead of 1 in the leaked-interval case) to make sure they genuinely exercise the bug, then confirmed all 15 tests in the file pass with the fix restored.

Full verification: `yarn typecheck`, `yarn lint:app`, `bcsc-core`'s `lint:ts`, both workspaces' TS `format:check`, and the full `yarn test` suite (280 suites / 2717 tests / 159 snapshots) all pass. `yarn check`'s Swift lint step fails on pre-existing `packages/bcsc-core/ios/BcscCore.swift` findings unrelated to this change (confirmed byte-identical to latest `origin/main`), consistent with a known local SwiftFormat false-fail.

## Verification

1. Start a transfer from an old device to a new device and let it complete normally on the source (old) device.
2. Immediately after the transfer completes, tap "Remove Account" on the source device.
3. Confirm no "account not found" critical error appears.
4. Leave the transfer QR screen open and idle for several minutes without removing the account — confirm the QR code still refreshes normally and no spurious errors appear.

#4273

### DIFF
--- a/app/src/bcsc-theme/features/account-transfer/transferer/TransferQRDisplayScreen.test.tsx
+++ b/app/src/bcsc-theme/features/account-transfer/transferer/TransferQRDisplayScreen.test.tsx
@@ -47,6 +47,16 @@ const createDeferredGetAccount = () => {
   return { promise, resolve }
 }
 
+// Same pattern as createDeferredGetAccount, for holding the createDeviceSignedJWT() await
+// "in flight" so a test can control exactly when the signed JWT comes back.
+const createDeferredJwt = () => {
+  let resolve!: (value: string) => void
+  const promise = new Promise<string>((res) => {
+    resolve = res
+  })
+  return { promise, resolve }
+}
+
 describe('TransferQRDisplayScreen', () => {
   let mockNavigation: any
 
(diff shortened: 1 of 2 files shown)


al-rosenthal (approved):
tested this locally multiple times and no error after removing account

Issue #4312, opened 2026-07-22
Mediator push notifications are received even when the app is foregrounded


bryce-mcmath:
### Description

With the new FCM-enabled mediators in dev and test, push notifications for digital credential messages, offers, proofs are currently being received even when the app is foregrounded

These should be silenced unless the app is closed or backgrounded

### Notes

Reach out to @bryce-mcmath for the correct MEDIATOR_URL to use to test

PR #4317, opened 2026-07-23
fix: prevent foreground non-bcsc notifications

bryce-mcmath:
# Summary of Changes

This PR prevents foreground notifications for non-BCSC notifications like mediator notifications

# Testing Instructions

Verify and confirm those notifications still work the same, then go through the showcase and ensure you don't get notifications during proof requests and such.

# Acceptance Criteria

Get notifications as appropriate

N/A

# Related Issues

#4312

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ lib
 local.properties
 .jest
 .npmrc
+.kotlin/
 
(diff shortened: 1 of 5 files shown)


al-rosenthal (approved):

Issue #4302, opened 2026-07-21
Combo card scan may discard licence details read moments before the scan completes


jleach:
## What

A combo card (BC Services Card and Driver's Licence on one card) carries two barcodes: a small one holding the card serial, and a larger, denser one holding the licence details including the birthdate.

Reading the scan logic suggests there may be cases where we read the larger barcode successfully, then discard what we read and ask the person to type their birthdate anyway.

If that happens in practice, the effect is an avoidable manual step during setup. Nothing is captured incorrectly and the scan still succeeds — the person just types in something the card had already given us.

## Why

We finish the scan based on what the camera can see in a single moment, rather than on everything it has seen.

We keep a running tally of how many times we've read each barcode, and that tally carries across moments. But when we decide the scan is complete, we only pass on the barcodes visible in that final moment. If the larger barcode was read a second earlier but isn't picked up in the instant we finish, the licence details go with it — even though we had them.

Worth being clear about what is *not* the problem, since both are easy to assume:

- We do handle both barcodes properly when both arrive together. The card-reading code merges them correctly, and if both are visible at the point we finish, we wait for both before completing.
- The barcode service we call for other card types can't cover the gap. It requires both barcodes itself, and isn't used by this screen.

There is an existing note in the code that catching both barcodes on a combo card can take several attempts on iOS, which is what makes this worth checking rather than dismissing.

## How to confirm

This needs a combo card and a device before any fix is designed:

1. Scan a combo card and see whether setup asks for the birthdate or fills it in.
2. Repeat enough times to get a feel for how often each outcome happens.
3. The scanner already logs each decode (`[CardScanner] decoding barcode` in dev builds), so the logs will show whether the larger barcode was read at some point before the scan completed.

If the birthdate is reliably filled in, this can be closed — the scenario exists in the logic but not in practice.

## Technical detail

**Where the scan ends** — `app/src/bcsc-theme/components/CodeScanningCamera.tsx`

- `determineScanState` (`utils/camera.ts`) evaluates only the codes decoded in the **current frame**. `qualifyingCodes` is derived from that frame; the persistent `barcodeReadings` map supplies counts, not membership.
[...]

Issue #4300, opened 2026-07-21
Scan camera autofocus is driven by a repeating timer that leaves little room for continuous autofocus


jleach:
## What

On the card scan screen, the camera doesn't bring itself into focus. People have to move the phone back and forth until the card happens to land in focus, and then the scan works normally.

To someone scanning a card, this looks like the camera is slow or unreliable. It isn't — once the card is sharp, reading it works fine. The camera just isn't doing the focusing part on its own.

## Why

We ask the camera to focus on a fixed spot every two and a half seconds, on a repeating timer.

That was a deliberate decision, and for a good reason: the barcodes on these cards are very small, and phones left to their own devices don't always focus on something that small held up close. Pointing the camera at the card region was an attempt to help.

The catch is that the instruction we send is a *focus once and hold* instruction, not a *keep focusing* one. Sending it again every couple of seconds means the phone is repeatedly told to stop adjusting and hold still — so its own continuous focusing never gets a turn. The thing we added to improve focus is what keeps the camera from focusing.

This is long-standing behaviour, not something that changed in v4.1. Raising it so the approach can be looked at deliberately rather than adjusted by feel.

## Technical detail

`app/src/bcsc-theme/components/CodeScanningCamera.tsx`

- `FOCUS_CYCLE_INTERVAL_MS = 2500`, with `doFocus` invoked from a `setInterval`
- `doFocus` calls `camera.current.focus({ x, y })` at the centre of a scan zone

In `react-native-vision-camera`, `focus()` is one-shot on both platforms:

- **iOS** (`ios/Core/CameraSession+Focus.swift`) sets `device.focusMode = .autoFocus` and enables subject-area-change monitoring. It returns to `.continuousAutoFocus` only via the `subjectAreaDidChange` handler, i.e. when the scene changes enough to fire that notification.
- **Android** (`android/.../CameraSession+Focus.kt`) builds a `FocusMeteringAction` and calls `startFocusAndMetering`, which pre-empts continuous AF for that action.

A supporting detail: the cycling code explicitly swallows `capture/focus-canceled` errors, which is what you would expect if new focus requests were routinely cancelling in-flight ones.

### The cycling is effectively a no-op

`doFocus` contains round-robin logic and a "prefer zones that have not yet detected a barcode" preference, which reads like an adaptive multi-zone system. In practice there is only ever one zone:

[...]

Issue #4256, opened 2026-07-16
Card Scan does not work for BCSC


RaulConstan:
### Description of problem
On Android, in v4.1.0, the camera scan does not pick up the info from the barcodes.

### Expected behavior
works

### Steps to reproduce
1. Scan a BCSC in "step 1"

### Screenshots and/or log output
Camera not working:

https://github.com/user-attachments/assets/9d74b8aa-6056-405c-917c-05842db84786

### Environment
SIT

Build #:
7389

Android Device Model:
Pixel 6 (Android 16)

iOS Device Model:

jleach:
Noted on review today:
- This can happen if the app goes to sleep with the camera on or if the user switches camera apps.

cmathies16:
Move error code to separate ticket.

jleach:
## Investigation summary

**The short version:** the app is reading the card correctly. The camera sees the barcode and decodes it successfully — and then the app sets the result aside because it decides the barcode isn't sitting in quite the right spot on screen.

We added that position check as a safety net. In practice it's now catching the very thing we're trying to land. To the user it's indistinguishable from "the camera is broken": they hold the card up, nothing happens, and there's no message explaining why. That's why this has been hard to characterise from the outside.

There's a second, smaller contributor stacked on top: if the phone sleeps or the user switches apps and comes back, the camera can return in a non-working state. Different cause, same visible symptom — which is part of why this looked like one intermittent problem rather than two distinct ones.

## Highest-value changes

**1. Stop requiring the barcode to be in the "right place."**

We already check the *content* of what we read — we can tell a BC Services Card from a driver's licence from a combo card by decoding it. The position check adds very little on top of that. If we successfully read a valid BC Services Card, it's a valid BC Services Card regardless of where in the frame it happened to sit.

We'd keep the on-screen card outline as a *guide* — it still helps people frame the card, and it still gives feedback — but it stops being a gate that can quietly reject a good scan. This removes a whole category of failure rather than tuning it. It's also mostly *removing* logic, which is unusually good news for effort and risk.

*Detail, for the record — this is also why the check is mis-tuned today, and why it reproduces in UAT but not for us:*

[...]

PR #4324, opened 2026-07-23
fix(bcsc): stop scan focus re-trigger storm and keep combo-card licence data

jleach:
## What

Card scanning on the BC Services Card screen was unreliable, especially on Android: the camera's auto-focus would refire rapidly while the app was still deciding whether a barcode was properly aligned, which could blur the frame right as it was trying to read a card. Separately, on combo cards (BC Services Card + driver's licence in one card), the scan could lock in and finish using only the easier serial-number barcode, silently dropping a driver's licence barcode it had *already* read a moment earlier — sending the user into a manual birthdate-entry step they shouldn't have needed.

This change fixes both:
- The camera's auto-focus nudge no longer restarts itself every time the on-screen alignment state flips; it keeps running on its own steady schedule and only nudges focus when nothing has been read in a while.
- When a scan locks in on the serial number alone, any driver's licence barcode that was already read moments before is now carried along instead of thrown away.

## Why

Users on Android were reporting that scanning a BC Services Card in "step 1" of verification wasn't picking up the barcode data reliably (issue #4256). Investigation (see #4300 and #4302) found two remaining causes after the earlier fix in #4286:
1. The focus-cycling logic was tied to the on-screen scanning/aligned state, so it restarted itself dozens of times a second as that state flickered — undermining the auto-focus instead of helping it.
2. A combo card's lock only strictly needs the barcode with the serial number; the driver's-licence barcode data the camera had *already* validated a moment before could be dropped on the floor, forcing a fallback to manual birthdate entry that shouldn't have been necessary.

## Decisions

- **Focus fix scope (rec 1 only).** Removed `scanState` from the focus-cycling `useEffect`'s dependency array so a scanning↔aligned flip no longer tears down and restarts the cycle (the actual re-trigger storm). The 2.5s periodic focus timer itself is **unchanged** — rec 2's "event-driven" design is already delivered by #4286's `FOCUS_SUPPRESS_AFTER_DETECTION_MS` suppression window (a tick only calls `focus()` once nothing has been decoded recently), so only the doc comments were reframed to describe that as the intended design. No platform gate was added — this applies on both iOS and Android.
- **One combined PR.** Per plan, this PR bundles both the focus fix (rec 1) and the combo-card accumulator fix (rec 3) plus all associated tests, rather than splitting into two PRs.
- **Accumulator merge ordering.** `mergeLockedCodesWithAccumulated` returns accumulated extras *before* current-frame codes. `useCardScanner`'s decode loop applies "later code wins" for same-kind data, so putting the current frame last means a fresher reading always overrides a stale accumulated one of the same kind — ordering alone resolves that case; no same-kind entries are proactively dropped.
- **Rec 2 (event-driven focus suppression) was already delivered by #4286** (`FOCUS_SUPPRESS_AFTER_DETECTION_MS`) — nothing further needed here beyond the comment reframing above.
- `useFocusEffect`'s focus-gained handler now explicitly calls `startFocusCycling()` on screen refocus, since that previously happened only as an incidental side effect of the (now-removed) `scanState` dependency.
- Left the lock threshold (`LOCK_READING_THRESHOLD`) and `determineScanState` untouched, per plan — only the *codes carried into the lock* changed, not what triggers a lock.
- **Dead-code cleanup (bundled).** Removed two provably-inert refs in `CodeScanningCamera.tsx`: `isProcessingScan` (write-only, never read — the real re-entrancy guard is `isLockedRef`) and `highlightTimeoutRef` (never assigned a timeout, so its clear-blocks were no-ops; the live highlight timer is the separate `clearHighlightTimeoutRef`). Deletions only, no behavioral change.
- **Known limitation — accumulator is time-scoped, not card-scoped.** The accumulator merged into a lock is bounded only by the accumulation window, not by card identity. If a user physically swaps to a *different* BC Services Card mid-scan within that window and the new card's serial locks before the prior card's licence barcode expires, the merge could pair a serial from one card with a birthdate from the other. This is contrived (a holder has one card) and self-correcting — the backend rejects the mismatched serial+birthdate (→ MismatchedSerial screen → retry), with no incorrect authorization or data corruption. The naive fix (clear on barcode-decay) would defeat the feature, since carrying a no-longer-visible barcode into the lock is its whole purpose; a card-identity-aware invalidation is deferred as out of scope. Documented in `mergeLockedCodesWithAccumulated`.

## Tests

- `CodeScanningCamera.test.tsx` — new `Focus cycling restart storm (regression #4256/#4300)` describe: a discriminating case showing the call count is unchanged across a scanning↔aligned flip once the suppression window has lapsed (where the old code would have fired an extra unsuppressed `focus()`), while the periodic idle-nudge tick still fires afterward; idle-nudge cadence preserved with zero detections; idle-nudge survives a lock → auto-confirm-rejected → reset cycle.
- `CodeScanningCamera.test.tsx` — new `Lock keeps accumulated licence barcode (regression #4256/#4302)` describe: a validated-but-not-locked PDF-417 rides along into a lock triggered by the serial alone; a second case confirms the accumulated barcode is correctly dropped once it's aged out past the accumulation window.
- `utils/camera.test.ts` — new `mergeLockedCodesWithAccumulated` unit tests: merge, dedupe, window expiry (including the exact boundary), ordering, and the empty-accumulator passthrough.
- `hooks/useCardScanner.test.tsx` — added a case asserting `scanCard` handles the serial-before-licence code order correctly.
- Full suite: `yarn check` (typecheck + lint + format + all 283 test suites / 2812 tests) passes clean; no snapshot changes.

## Verification

1. On a device (ideally Android), go to Add a Card → BC Services Card → scan a combo card (BC Services Card + driver's licence).
2. Confirm the scan locks in promptly without visible focus "hunting"/flicker, and that scanning a combo card resolves straight to the setup flow (birthdate pre-filled from the licence) rather than falling back to manual birthdate entry.
3. Confirm scanning a serial-only BC Services Card (no combo licence data) still works as before.

Also addresses #4300 (focus re-trigger storm) and #4302 (combo-card licence data retention) — leaving those open for the reporter to close.

### DIFF
--- a/app/src/bcsc-theme/components/CodeScanningCamera.test.tsx
+++ b/app/src/bcsc-theme/components/CodeScanningCamera.test.tsx
@@ -1921,4 +1921,275 @@ describe('CodeScanningCamera', () => {
       )
     })
   })
+
+  describe('Focus cycling restart storm (regression #4256/#4300)', () => {
+    beforeEach(() => {
+      jest.useFakeTimers()
+      // The module-level device mock factory returns a brand-new object on every call,
+      // so without pinning it, startFocusCycling's useCallback (deps include `device`)
+      // would get a new identity on every re-render regardless of cause — masking the
+      // exact thing under test here. Pin it to a single stable reference, matching how
+      // the real hook behaves when the underlying camera device hasn't changed.
+      mockedUseCameraDevice.mockReturnValue({
+        id: 'back',
+        supportsFocus: true,
+        minZoom: 1,
+        maxZoom: 8,
+        neutralZoom: 1,
+        hasTorch: true,
+      })
+    })
+
+    afterEach(() => {
+      jest.useRealTimers()
+      // Restore the default per-render factory for any tests outside this describe.
+      mockedUseCameraDevice.mockImplementation(() => ({
+        id: 'back',
+        supportsFocus: true,
+        minZoom: 1,
+        maxZoom: 8,
+        neutralZoom: 1,
+        hasTorch: true,
+      }))
+    })
+
+    const mockFrame = { width: 1920, height: 1080 }
+
+    it('does not restart on a scanning<->aligned flip, but the periodic idle-nudge still ticks (discriminating case)', async () => {
+      const { getByTestId } = render(
+        <BasicAppContext>
+          <CodeScanningCamera {...defaultProps} />
+        </BasicAppContext>
+      )
+
+      // t=0: layout starts cycling — one immediate, unsuppressed doFocus (lastDetectionAtRef
+      // starts at 0, so nothing looks "recent" yet).
+      const cameraContainer = getByTestId('camera-preview-container')
+      await act(async () => {
+        fireEvent(cameraContainer, 'layout', {
+          nativeEvent: { layout: { x: 0, y: 0, width: 400, height: 800 } },
+        })
+      })
+      const callsAfterLayout = mockFocus.mock.calls.length
+      expect(callsAfterLayout).toBeGreaterThan(0)
+
+      // t=100: a single detection is enough to flip scanning -> aligned (minCodesForAligned
+      // === 1 for the single-zone BCSC_SN_SCAN_ZONES).
+      await act(async () => {
+        jest.advanceTimersByTime(100)
+      })
+      await act(async () => {
+        mockCodeScannerCallback!(
+          [
+            {
+              type: 'pdf-417',
+              value: 'FLIP_TO_ALIGNED',
+              frame: { x: 100, y: 200, width: 200, height: 50 },
+              corners: [],
+            },
+          ],
+          mockFrame
+        )
+      })
+
+      // t=2200: advance 2100ms past that detection — the suppression window (2000ms) has
+      // now lapsed — then send a blank frame. The single tracked reading (count 1) decays
+      // straight to 0 and drops the state back to 'scanning'.
+      //
+      // OLD code: scanState was a dependency of the cycling effect, so this flip tore the
+      // cycle down and restarted it, firing an extra doFocus() right here — and since the
+      // suppression window had already lapsed, that restart's call was NOT suppressed.
+      // NEW code: scanState is not a dependency, so nothing restarts and the call count
+      // must be unchanged.
+      await act(async () => {
+        jest.advanceTimersByTime(2100)
+      })
+      await act(async () => {
+        mockCodeScannerCallback!([], mockFrame)
+      })
+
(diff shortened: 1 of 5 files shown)


jleach:
## On-device verification (Pixel 7a, this branch @ `d0d797f92`)

Built and flashed this PR to a Pixel 7a and instrumented the camera session + JS card scanner with `adb logcat`. Compared against the current-`main` build captured earlier on the same device.

### 1. Focus re-trigger storm — eliminated

`focus()` calls per scan (the pinned nudge at zone centre `(0.46, 0.50)`):

| Build | focus() calls |
|---|---|
| current `main` (storm) | ~22–24 in a single ~3.5 s read (gaps as low as 1–30 ms) |
| **this PR** | **2 per scan, both at init, then none** |

```
Scan 1: 23:28:24.258, .259   → then silence
Scan 2: 23:28:31.956, .956   → then silence
Scan 3: 23:28:37.583, .584   → then silence for 6 s
```

The two init calls are the mount nudge + the `useFocusEffect` restart; after that the cycle no longer restarts on scanning↔aligned flips.

### 2. Continuous AF now runs — the "too far, move it closer" case

Scan 3 was held too far away and moved in until it focused — the exact hunt-for-focus symptom from the report. Trace:

```
37.583  pinned focus() nudge ×2            (initial one-shot)
37.611  af_mode=Auto           TriggerStart → ActiveScan
38.083  af_mode=Auto           ScanSuccess  → FocusedLocked   (initial lock, card still too far)
42.664  af_mode=Auto           TriggerCancel → Inactive       (one-shot releases)
42.698  af_mode=ContinuousVideo ScanStart    → PassiveScan     ← continuous AF takes over
43.001  af_mode=ContinuousVideo ScanSuccess  → PassiveFocused
43.706  decode: BCServicesComboCardCardBarcode
```

Because the app no longer re-fires `focus()` on a sub-second cadence, the device's own continuous autofocus gets to run and re-focuses as the card is brought in. On `main` the storm kept preempting continuous AF back into one-shot `Auto`, which is what made focus feel unreliable.

### 3. Combo-card licence retention (rec 3) — holding

All 3 scans decoded `BCServicesComboCardCardBarcode` (both barcodes) — no serial-only completion, no manual-birthdate fallback. 0 AF failures across the session.

### Timings (init → decode)

| Scan | init → decode | Result |
|---|---|---|
| 1 | 0.75 s | combo |
| 2 | 1.56 s | combo |
| 3 (far, reframed) | 6.19 s | combo |

Note: this is a single-device pass on Android (Pixel 7a). The changes are cross-platform; an iOS smoke pass is still worth doing before release.

bryce-mcmath (approved):

PR #4331, opened 2026-07-26
fix: no option to remove account from the verify menu

### LINKED ISSUE #4292: No option to reset app from the "Choose How To Verify" screen
When you're on the "choose how to verify" screen, the 'burger' menu does not have the reset / remove account option.
### Description of problem
in v3.15, the user has the ability to reset the app (remove the account and start over) at any point in the process.
v4 doesn't have the same option, so the user is stuck in the verification stream unless they get approved, or uninstall/reinstall the app
### Steps to reproduce
enter the identity info until you are prompted to choose a verification method.
tap the menu icon and note the lack of the Remove Account option.

### Environment

Build #:7488
Android Device Model:

iOS Device Model: 15+

jleach:
## What

Adds a "Remove account" option to the hamburger settings menu shown during the identity verification flow. Selecting it takes the user to the same remove-account confirmation screen already used elsewhere in the app.

## Why

Before this change, once a user started verifying their identity there was no way to reset the app from the verification screens — the settings menu offered app security, notifications, and other options, but not "Remove account". A user who didn't want to continue, or who got stuck, had no in-app way to start over short of uninstalling and reinstalling. This matches a capability the app had in a previous release (v3.15).

## Decisions

- Wired `onRemoveAccount` in `VerifySettingsScreen` to navigate to `VerifyRemoveAccountConfirmation` — `SettingsContent` already renders the "Remove account" row when this prop is provided, so no new components or i18n keys were needed. This mirrors how `MainSettingsScreen` already provides the same option.
- The option appears in the hamburger settings menu (accessed via the menu icon on the "Choose how to verify" screen), matching where v3 placed it and where users expect settings-level actions.
- Originally implemented in the `?` help menu — reverted after realizing the correct placement is the hamburger settings menu, not the contextual help menu.

## Verification

1. Build the `bcsc` target and start identity verification until reaching "Choose how to verify".
2. Tap the hamburger menu icon (top left) to open Settings.
3. Confirm "Remove account" appears in the settings list.
4. Tap it, confirm on the following screen, and confirm the app resets to onboarding.
5. Confirm removing the account from Home/Settings still works (no regression).

Fixes #4292


bryce-mcmath (dismissed):
Needs snapshot fix, then I will reapprove

app/src/bcsc-theme/features/settings/VerifySettingsScreen.test.tsx
```diff
@@ -13,15 +17,29 @@ describe('VerifySettings', () => {
...
+  it('navigates to remove account confirmation when remove account is pressed', () => {
+    const { getByTestId } = render(
+      <BasicAppContext initialStateOverride={authenticatedState}>
         <BCSCLoadingProvider>
           <VerifySettingsScreen navigation={mockNavigation as never} />
         </BCSCLoadingProvider>
       </BasicAppContext>
     )
 
-    expect(tree).toMatchSnapshot()
```
bryce-mcmath, line 25: Curious, why remove the snapshot test?

jleach: When I first looked at it that seemed like test theatre. Much of the good stuff is gated behind isAuthenticated so the snapshot seemed like a rubber stamp. I went with focused tests. It does catch some useful stuff though like ContactUs, Help, etc. I'll add it back with both states for `isAuthentnicated` so we capture the all the things.

bryce-mcmath (approved):

PR #4332, opened 2026-07-26
refactor: rename reportUUID to installId

### LINKED ISSUE #4325: refactor: rename reportUUID to installId
# Rename reportUUID → installId

## Problem: 
reportUUID is easily confused with reportId (the per-report identifier like 7K2P-9XQF), even though it actually represents the install, not a specific report.

## Requirements:

- Rename reportUUID → installId across store, actions/reducers, and call sites
- Update any persisted storage keys/migrations if this value is persisted under its old name
- Check reportProblem flow specifically, since that's the main confusion point — confirm nothing there relies on the old key name string

## Note: 
worth a repo-wide search for reportUUID string literals (not just the variable) in case it's also used as a storage key, log field, or API payload key — those need separate handling if you don't want to break existing stored data or backend expectations.

jleach:
## What

Renames an internal identifier so that the value identifying **an app installation** can no longer be confused with the **per-report reference code** a person quotes to support (e.g. `7K2P-9XQF`).

No user-visible change: no screens, copy, flows, or analytics change, and nothing new is collected or sent. People who already have the app keep the same installation identity when they upgrade — support can still connect a new report to that install's earlier history.

## Why

When someone reports a problem, support works with two different values that sounded like the same thing: the reference code for that one report, and a long-lived id for that person's app installation. Because the names were nearly identical, it was easy for a change to wire the wrong one into the wrong place — a mistake that reads as correct in review and only surfaces later, as support being unable to trace a report back to the install it came from. Renaming removes that ambiguity at the source.

Carrying existing installs forward mattered as much as the rename. Handled naively, everyone's installation identity would have been quietly reset on upgrade, cutting support's link between a new report and that install's history — the exact continuity the value exists to provide. This change preserves it.

Technical debt cleanup; no functional change requested by the business.

---

## Decisions

- **Renamed everywhere the value flows**: the store field (`BCSCState.reportUUID` → `installId`), the dispatch action (`SET_REPORT_UUID` → `SET_INSTALL_ID`, string value `'bcsc/setReportUUID'` → `'bcsc/setInstallId'`), the system check class (`ReportUUIDSystemCheck` → `InstallIdSystemCheck`, moved via `git mv`), `ErrorAlertContext`, `ErrorModal`'s payload, and the `reportProblem()` call site in `ReportProblemModal.tsx`.
- **Read-side migration for existing installs**: the BCSC state blob is persisted to native secure storage and hydrated on every launch (`container-imp.ts`, `TOKENS.LOAD_STATE`). Without a migration, every existing install would silently lose its id on upgrade and the startup system check would mint a brand-new one — defeating the entire point of an install-scoped id. Added a pure `migrateBCSCState()` helper (`app/src/store.tsx`) mapping a persisted `reportUUID` forward to `installId` (preferring an already-migrated `installId` if both are somehow present), plus a best-effort write-back so the mapping only runs once per install. The write-back never throws — if it fails, the mapping simply re-runs next launch.
- **Action string value renamed too, with all three hardcoded test-mock literals updated in the same commit.** Had a mock kept the old literal while the reducer moved, the reducer would stop matching and those tests would pass vacuously.
- **Found and fixed two call sites the plan didn't list**: `ReportProblemModal.tsx` — the file the issue names as "the main confusion point" — had two live references to `store.bcsc.reportUUID` feeding its `reportProblem()` call. Skipping them would have left the actual report flow reading `undefined` after the store rename (and in fact would not have compiled).
- **`ErrorModal.tsx`'s Loki log-body line does not exist and was not added.** The plan called for renaming an appended `` `\nReport ID: ${error.reportUUID}` `` to `` `\nInstall ID: …` ``. That line exists nowhere in the current code — a prior refactor removed all string-concatenation from the report path (see the comment at `ErrorModal.test.tsx:224-226`), and `AppError` has no such property to read from. A repo-wide grep for `"Report ID:"` / `"Install ID:"` returns zero hits. The install id already reaches Loki correctly as a structured field (`install_id`, `app/src/utils/logger.ts`), not as message text — so only the real `ErrorModalPayload.reportUUID` → `installId` field rename applied.
- Rename is functionally scoped to the **bcsc** target; **bcwallet** compiles unchanged (the field is simply `undefined` there).

## Tests

- New tests in `app/src/store.test.ts` for `migrateBCSCState`: legacy-only blob, both fields present (installId wins, legacy stripped), already-migrated and empty blobs (idempotent no-ops), plus reducer tests for `CLEAR_BCSC` (preserves `installId`) and `SET_INSTALL_ID`.
- New `app/container-imp.test.ts` — the end-to-end proof. Builds `AppContainer` against a minimal fake DI container, runs the real `.init()`, invokes the real `TOKENS.LOAD_STATE` closure, and asserts the migrated `installId` on the dispatched payload (and that the write-back fires only when a migration actually happened).
- Existing suites updated in place for the rename: `InstallIdSystemCheck.test.ts` (renamed via `git mv`), `useCreateSystemChecks.test.tsx`, `ErrorAlertContext.test.tsx`, `ErrorModal.test.tsx`.

### Test-theater fix round (commits 2 & 3)

An adversarial review found zero code defects but three tests that stayed green even when the code they claimed to cover was deleted. All three fixed test-only (no production code changed), each verified by deleting the target line, confirming red, then restoring:

- **The migration wiring had no coverage at all.** Deleting `bcsc = bcscMigration.bcsc as BCSCState` in `container-imp.ts` — the mutation that silently re-identifies every upgrading install — left all 283 suites passing, because no test anywhere loaded `container-imp.ts`. `app/container-imp.test.ts` (above) closes this; it now fails on that deletion, on gutting `migrateBCSCState`, and on flipping the `installId ?? reportUUID` precedence.
- **`ErrorModal.test.tsx` asserted `installId: undefined`** (the fixture never set one), so `toHaveBeenCalledWith` matched a call omitting the key entirely — deleting `installId: payload.installId` from `handleReport` changed nothing. The fixture now carries a real id, and `ErrorAlertContext.test.tsx` pins the store → payload hop the same way.
- **An assertion-free test** (`should call setInstallId with a generated uuid` — a bare `onFail()` call, no `expect`) deleted; its sibling already covers the dispatch with real assertions.

## Verification

- `yarn check` (typecheck + lint + format check + full test suite) passes: 284 suites, 2831 tests, 159/159 snapshots unchanged.
- Repo-wide grep for `reportUUID` / `ReportUUID` / `SET_REPORT_UUID` / `setReportUUID` across `app/` and `packages/` returns zero hits outside the migration helper itself (which necessarily names the legacy key it reads and strips) and its tests.
- **Not yet verified — suggested manual follow-up before release**: on-device upgrade smoke test. Install a build with the old code so a legacy id is persisted, upgrade to this build, and confirm the same id survives as `installId` with no new one minted; confirm a fresh install still gets a newly-minted id.

Fixes #4325


app/src/store.tsx
```diff
@@ -325,6 +325,23 @@ export const initialBCSCState: BCSCState = {
...
+  // `& { installId?: string }` keeps installId in the return type even when T is inferred from a
+  // narrow literal that never mentions it (e.g. `{ reportUUID: 'x' }` in tests).
+): { bcsc: Omit<T, 'reportUUID'> & { installId?: string }; migrated: boolean } => {
+  const { reportUUID, ...rest } = persisted
+  if (reportUUID === undefined) {
+    return { bcsc: rest, migrated: false }
+  }
+  return { bcsc: { ...rest, installId: rest.installId ?? reportUUID }, migrated: true }
+}
+
```
MacDeluca, line 343: question: is reportUUID in 4.0.2?

This function is cool, it might be nice to include when a property was added in the documentation (commit hash / release). So it's clearer in the future when we can clean up migrated values.

jleach: Yep, shipped in 4.0.2. Suggestion is good, implemented.

MacDeluca (approved):
✅

PR #4337, opened 2026-07-27
fix: prevent zombie store recreation on factory reset mid-verification

### LINKED ISSUE #4336: Factory reset mid-verification leaves wallet in unrecoverable state
## What

When a user removes their account while in the verification flow (e.g. after taking a selfie), the app enters a broken state where it can no longer initialize the credential wallet or receive digital credentials. The only recovery is clearing app data or reinstalling. Removing the account from the Home screen works correctly.

```
console.js:661 Error creating store 'bc-wallet-bcsc' {
  "error": {
    "code": 1
  },
  "errorMessage": "Error creating database tables\nCaused by: error returned from database: (code: 1) table config already exists"
} 
anonymous @ console.js:661
...
```

## Why

Users need the ability to reset and start over at any point in the verification process — for example if they entered wrong information, want to switch devices, or simply changed their mind. The upcoming "Remove account" menu option in the verification flow (#4292 / PR #4331) makes this path easily reachable. If the underlying reset doesn't work reliably from verification, the feature can't ship.

This issue is pre-existing — the same `useFactoryReset` code path runs from Settings and Home — but it is latent there because the conditions that trigger the failure only arise during verification. Fixing it unblocks #4331.

## Technical details

<details>
<summary>Root cause analysis (device- and code-proven)</summary>

### Mechanism

The factory reset inadvertently causes a transient React Navigation route change that mounts Bifold's credential provider tree against a zombie agent whose wallet store was just deleted. Credo's storage layer auto-provisions a new store when it can't find one, recreating the wallet under the old key. When the user re-onboards, a new wallet key is derived, and the agent can never open the old-keyed store again.

### Step-by-step sequence

1. `useFactoryReset` shuts the agent down and calls `deleteWalletStore` → Askar's `deleteStore()` removes `sqlite.db` and deregisters the `Store` instance from the DI container. This part is clean.

2. `clearSecureState()` dispatches `CLEAR_SECURE_STATE`, which resets `bcscSecure` wholesale (`store.tsx:503`). This clears `verifiedStatus` out of `IN_PROGRESS`, so `showVerifyStack` (`RootStack.tsx:106`) flips to false — but `hasAccount` (only cleared by the later `CLEAR_BCSC` dispatch) and `didAuthenticate` (flipped last) are still true.

3. For that intermediate render, `RootStack` falls through to `BCSCMainStack`, which mounts `BifoldScope` (`MainStack.tsx:152`) → Bifold's `AgentProvider` — fed the still-non-null zombie agent from `BCSCAgentProvider`.

4. The `@bifold/react-hooks` provider tree fires its mount-time fetches: `BasicMessageProvider` (`findAllByQuery`), `ConnectionProvider` (`getAll`), `CredentialProvider` + `CredentialFormatDataProvider` (`getAll` ×2), `ProofProvider` + `ProofFormatDataProvider` (`getAll` ×2) = exactly six storage reads.

5. Each read → `AskarStoreManager.withSession` → `getInitializedStoreWithProfile` → store instance missing → `openStore()` throws `AskarStoreNotFoundError` → Credo's fallback silently calls `provisionStore()`, auto-creating a brand-new wallet store. Six concurrent provisions race; one succeeds and the rest fail with `table config already exists`.

6. The recreated store is keyed with the OLD wallet key (from the zombie agent's config). Re-onboarding derives a NEW key → `openStore` hits Askar `Encryption` error → `AGENT_INITIALIZATION_ERROR` permanently.

### Evidence (instrumented device run)

Diagnostic probes added to `AskarStoreManager.provisionStore` and all seven `AskarStorageService` entry points. Reproduced on a physical Android device with a fresh `pm clear`:

```
FactoryReset: Removing local account file...
[DIAG-OP] findByQuery BasicMessageRecord      ← useSecureActions.tsx:1128
[DIAG-OP] getAll ProofRecord                  ← (dispatch CLEAR_SECURE_STATE)
[DIAG-OP] getAll CredentialRecord
[DIAG-OP] getAll ProofRecord
[DIAG-OP] getAll CredentialRecord
[DIAG-OP] getAll ConnectionRecord
FactoryReset: Clearing secure and plain BCSC state...
6× [DIAG] provisionStore caller stack
Error creating store 'bc-wallet-bcsc' … table config already exists (×5+)
```

Six storage reads → six provision attempts. The record types match the `@bifold/react-hooks` provider tree mount-time fetches exactly.

Device inspection post-reset confirmed a fully recreated store (4096-byte `sqlite.db` + 78KB WAL, `integrity_check ok`, 1 profile, 1 item, 8 open file descriptors).

### Why Home works but verification doesn't

Same `useFactoryReset` code path. From Home, `MainStack`/`BifoldScope` is already mounted — the providers initialized on app launch and do no storage reads at reset time. No remount → no fetch burst → store stays deleted.

From the verify flow, the reset itself causes the `MainStack` mount (via the transient route change described above), which triggers the provider tree's mount-time fetches after the store was already deleted.

### Important: `BifoldScope` is only in `MainStack`

The readers are NOT "hooks mounted during verification." `BifoldScope` is rendered only inside `MainStack.tsx:152-532`, not in `VerifyStack`. The transient `MainStack` mount mid-reset is what creates them.

</details>

<details>
<summary>Fix directions</summary>

The core invariant: **no mounted consumer should be able to issue a storage operation against a deleted store.**

### (a) Reorder dispatches — prevent the transient MainStack mount

Flip `didAuthenticate` and/or `hasAccount` before (or atomically with) clearing secure state. `RootStack` then short-circuits to `OnboardingStack` or `AuthStack` before `showVerifyStack` is even evaluated, so `MainStack` never mounts.

### (b) Null the agent after shutdown — disable the zombie reads

After `shutdownAgent`, null the provider's agent reference so `BifoldScope` passes `undefined` to Bifold's `AgentProvider`. The provider tree then skips its fetches (the `AgentReadyGate` premise). Needs `useAgentSetupViewModel` to expose a teardown path that does `shutdownAgent` + `setAgent(null)`.

### (c) Delete the store last — after the tree has settled

Capture the wallet key before clearing state; delete by URI (via `purgeWalletStore`) after the provider subtree has unmounted. This is timing-fragile on its own.

**Recommended: (a) + (b) as belt-and-braces.** (a) eliminates the transient mount; (b) ensures any future transient mount is harmless.

**Existing ordering constraint** (`useFactoryReset.tsx:121-128`): shutdown must precede delete; delete must precede clearing the wallet key. The fix must respect this.

</details>

<details>
<summary>Key files</summary>

- `app/src/bcsc-theme/api/hooks/useFactoryReset.tsx` — the reset hook; dispatch ordering lives here
- `app/src/bcsc-theme/navigators/RootStack.tsx:83-106` — `showVerifyStack` derivation and the route that falls through to MainStack
- `app/src/bcsc-theme/navigators/MainStack.tsx:152` — where `BifoldScope` is mounted
- `app/src/bcsc-theme/features/agent/BCSCAgentProvider.tsx` — holds the zombie agent reference
- `app/src/bcsc-theme/features/agent/useAgentSetupViewModel.ts` — agent lifecycle; shutdown effect at line 88
- `app/src/bcsc-theme/features/agent/BifoldScope.tsx` — passes `agent` to Bifold's `AgentProvider`
- `app/src/bcsc-theme/features/agent/services/agent-service.ts:317` — `deleteWalletStore`
- `app/src/store.tsx:503` — `CLEAR_SECURE_STATE` reducer
- `node_modules/@credo-ts/askar/build/AskarStoreManager.mjs:33` — `getInitializedStoreWithProfile` open-or-provision fallback
- `node_modules/@bifold/react-hooks/build/` — provider mount-time fetches (`BasicMessageProvider`, `ConnectionProvider`, `CredentialProvider`, `ProofProvider`)

</details>

<details>
<summary>Dead ends (do not revisit)</summary>

These theories were investigated and definitively ruled out:

- **Race between factory reset and in-flight agent init.** Ruled out: logs show the clean live-agent branch ran; agent was fully initialized before the reset.
- **Orphaned SQLite WAL/SHM replayed into a fresh database.** Ruled out: a 4096-byte main file with schema in the WAL is a healthy un-checkpointed store, not a corpse. Both captures showed `integrity_check ok`, 1 profile, 1 item. The store on disk was freshly provisioned, not resurrected.
- **`Store.remove()` silently failing (discarded return value).** Ruled out: the `exists()` guard in `provisionStore` would have caught a surviving file. The file was genuinely deleted; it was recreated by Credo's auto-provision.

</details>

## Acceptance criteria

- [ ] Removing the account mid-verification (from the ? help menu) completes successfully
- [ ] After removal, the app returns to onboarding with no console errors about store creation
- [ ] Re-onboarding after removal produces a working agent that can receive credentials
- [ ] Removing the account from Home/Settings continues to work (no regression)
- [ ] Wallet store directory is clean after reset: `adb shell run-as <pkg> ls files/.afj/wallet/bc-wallet-bcsc/` returns empty or absent

jleach:
## What

Removing the account while in the middle of ID verification (e.g. after taking a selfie) could leave the app in a broken state where it could no longer set up the credential wallet or receive digital credentials — the only way out was clearing app data or reinstalling. Removing the account from Home/Settings already worked correctly; this only affected the mid-verification path.

This fix makes account removal reliable no matter where in the app the user triggers it, which is a prerequisite for the upcoming "Remove account" option inside the verification flow itself (#4292 / PR #4331).

## Why

People need to be able to reset and start over at any point during verification — wrong info entered, switching devices, or just changing their mind. If the underlying reset doesn't work reliably from inside verification, that "start over" option can't ship.

## Decisions

Root cause: factory reset mid-verification caused a brief, unintended navigation change that mounted Bifold's credential provider tree against an agent whose wallet store had just been deleted. Credo's storage layer auto-provisions a new store when it can't find one, recreating the wallet under the *old* key. Re-onboarding then derives a *new* key, and the agent can never open the old-keyed store again.

Implemented the two changes recommended in the issue, as belt-and-braces:

- **Reorder dispatches in `useFactoryReset`** so `CLEAR_BCSC` (which flips `hasAccount` to `false`) fires before `clearSecureState` (which resets `bcscSecure`, including `verifiedStatus`). `RootStack` short-circuits to `OnboardingStack` as soon as `hasAccount` is `false`, before it ever derives whether to show the verification stack — so the transient `MainStack` mount that triggered the bug can no longer happen.
- **Add a permanent `teardownAgent` to `useAgentSetupViewModel`**, threaded through `BCSCAgentProvider`, that shuts the agent down, deletes the wallet store, and nulls the provider's agent reference (unlike `resetWallet`, it never re-initializes). This is belt-and-braces: even if some other transient mount slips through in the future, Bifold's provider tree will see `agent: undefined` and skip the storage reads that previously raced the deleted store.

`useFactoryReset` now awaits `agentCtx.teardownAgent()` on the live-agent branch instead of calling `shutdownAgent`/`deleteWalletStore` directly — that ordering-sensitive logic now lives in one place. The no-live-agent orphaned-store purge branch (`purgeWalletStore`) is unchanged.

`teardownAgent` guards against a concurrent `resetWallet` (via the existing `resettingRef`) but deliberately does **not** check `initializingRef` — this is a destructive path and must always delete the store when a live agent exists, never silently no-op because an init happens to be mid-flight (that would reproduce the same orphaned-store bug through a different door). The mirror case — `teardownAgent` called while `resetWallet` is already running — is left unguarded to match today's existing behavior for concurrent `resetWallet` calls; `resetWallet`'s own path still deletes the store unconditionally, so the store isn't left behind either way.

## Tests

- New `describe('teardownAgent', ...)` in `useAgentSetupViewModel.test.ts`: shuts down + deletes + nulls the agent, does not trigger re-initialization, stops both the attestation and credential-provisioning monitors before shutdown, logs a warning and still nulls the agent when store deletion rejects, and no-ops when there's no live agent.
- `useFactoryReset.test.ts` updated to assert the reset awaits `agentCtx.teardownAgent()` to completion before continuing, and dispatches `CLEAR_BCSC` before `clearSecureState` before `DID_AUTHENTICATE`.
- `yarn check` (typecheck, lint, format, full test suite — 283 suites / 2832 tests) passes.

## Verification

1. Start verification (e.g. through taking a selfie / submitting an evidence photo).
2. Open the `?` help menu and remove the account.
3. Confirm the app returns cleanly to onboarding with no console errors about "table config already exists" / store creation.
4. Re-onboard and confirm the new wallet initializes and can receive a credential.
5. Repeat account removal from Home and from Settings to confirm no regression there.

Fixes #4336

### DIFF
--- a/app/src/bcsc-theme/api/hooks/useFactoryReset.test.ts
+++ b/app/src/bcsc-theme/api/hooks/useFactoryReset.test.ts
@@ -1,7 +1,7 @@
 import useApi from '@/bcsc-theme/api/hooks/useApi'
 import { useFactoryReset } from '@/bcsc-theme/api/hooks/useFactoryReset'
 import { useBCSCAgentSafe } from '@/bcsc-theme/features/agent/BCSCAgentProvider'
-import { deleteWalletStore, purgeWalletStore, shutdownAgent } from '@/bcsc-theme/features/agent/services/agent-service'
+import { purgeWalletStore } from '@/bcsc-theme/features/agent/services/agent-service'
 import { useBCSCApiClientState } from '@/bcsc-theme/hooks/useBCSCApiClient'
 import useSecureActions from '@/bcsc-theme/hooks/useSecureActions'
 import { BCDispatchAction } from '@/store'
@@ -21,11 +21,11 @@ jest.mock('@/bcsc-theme/features/agent/BCSCAgentProvider', () => ({
   useBCSCAgentSafe: jest.fn(),
 }))
 // Factory mock (not automock) so the agent-service module's heavy transitive deps
-// (Credo, indy-vdr-shared, etc.) are never loaded by this hook test.
+// (Credo, indy-vdr-shared, etc.) are never loaded by this hook test. Only
+// purgeWalletStore is called directly by useFactoryReset now — the live-agent
+// shutdown/delete path is delegated to agentCtx.teardownAgent (mocked separately).
 jest.mock('@/bcsc-theme/features/agent/services/agent-service', () => ({
-  deleteWalletStore: jest.fn().mockResolvedValue(undefined),
   purgeWalletStore: jest.fn().mockResolvedValue(undefined),
-  shutdownAgent: jest.fn().mockResolvedValue(undefined),
 }))
 jest.mock('react-native-config', () => ({ Config: { INDY_VDR_PROXY_URL: '' } }))
 
@@ -141,7 +141,7 @@ describe('useFactoryReset', () => {
     expect(deleteRegistrationMock).toHaveBeenCalledWith('native-token', 'test-client-id')
   })
 
-  it('should delete the wallet store and shut down the agent before clearing state', async () => {
+  it('awaits agentCtx.teardownAgent before clearing state when a live agent exists', async () => {
     const bcscCoreMock = jest.mocked(BcscCore)
     const useSecureActionsMock = jest.mocked(useSecureActions)
     const bifoldMock = jest.mocked(Bifold)
@@ -150,14 +150,13 @@ describe('useFactoryReset', () => {
 
     const clearSecureStateMock = jest.fn()
     const deleteSecureDataMock = jest.fn().mockResolvedValue(undefined)
+    const teardownAgentMock = jest.fn()
 
     const callOrder: string[] = []
-    // Both teardown ops now route through the serialized agent-service helpers.
-    jest.mocked(deleteWalletStore).mockImplementation(async () => {
-      callOrder.push('deleteStore')
-    })
-    jest.mocked(shutdownAgent).mockImplementation(async () => {
-      callOrder.push('shutdown')
+    // Agent shutdown + wallet store deletion are now owned by useAgentSetupViewModel's
+    // teardownAgent (see its own test suite); useFactoryReset just awaits it.
+    teardownAgentMock.mockImplementation(async () => {
+      callOrder.push('teardownAgent')
     })
     deleteSecureDataMock.mockImplementation(async () => {
       callOrder.push('deleteSecureData')
@@ -173,6 +172,7 @@ describe('useFactoryReset', () => {
       error: null,
       retry: jest.fn(),
       resetWallet: jest.fn(),
+      teardownAgent: teardownAgentMock,
     })
 
     useBCSCApiClientStateMock.mockReturnValue({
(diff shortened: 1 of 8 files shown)


al-rosenthal (approved):

PR #4389, opened 2026-07-31
feat: trigger DSC auto issuance by schema ID as well as cred def ID

### LINKED ISSUE #4383: Make DSC auto issuance triggered by schema ID not just cred def ID
### Description

Currently the BC Digital Services Card auto issuance is only triggered by cred def ID, it should also be triggered by schema IDs.

See AutoFetchCredentialConfig in constants.ts

jleach:
## What

The BC Digital Services Card auto-issuance flow — the one that quietly fetches a missing Person Credential when a verifier's proof request asks for it — currently only recognizes that request if the verifier restricted it by credential definition ID. This change also recognizes it when the verifier restricted it by schema ID instead, so the auto-issuance flow works either way.

## Why

Some verifiers write their proof requests using schema ID restrictions rather than credential definition ID restrictions. Those users would hit a proof request for a credential they don't have yet, without the app automatically fetching it for them first. This closes that gap.

## Decisions

- Added a `schemaIDs` field per environment to `AutoFetchCredentialConfig` in `app/src/constants.ts`, paired with each environment's existing cred def ID, using the Person schema IDs already present in `DIGITAL_SERVICES_CARD_SCHEMA_IDS`.
- Redefined `DIGITAL_SERVICES_CARD_SCHEMA_IDS` as derived from `AutoFetchCredentialConfig` (`Object.values(...).flatMap(...)`) instead of a separately maintained literal list, so the two can't drift apart.
- `AutoCredentialRule` gained an optional `triggerSchemaIds?: string[]`, kept optional for backward compatibility with any other rule that doesn't set it.
- Centralized the restriction-matching logic (cred_def_id OR schema_id) into one private `restrictionMatchesRule` helper on `AutoCredentialMonitor`, used by all three places that previously checked cred_def_id only.
- Matching stays exact-string, matching the existing cred def ID behavior (no qualified/unqualified schema ID normalization).
- `buildDigitalServicesCardCredentialRule` now also flattens and sets `triggerSchemaIds` from `AutoFetchCredentialConfig`.

## Tests

- Added tests in `app/src/services/auto-credential.test.ts` covering: schema-only-restriction triggers a workflow when credential missing; schema_id restriction that matches no rule doesn't trigger; schema-matched credential already in wallet doesn't trigger; a rule without `triggerSchemaIds` ignores schema_id restrictions (backward compat). Existing cred-def-triggered tests pass unchanged.
- Added a test in `app/src/bcsc-theme/services/digitalServicesCardProvisioner.test.ts` asserting `triggerSchemaIds` is the flattened set of `schemaIDs` from every `AutoFetchCredentialConfig` environment.
- `yarn check` (typecheck + lint + format + full test suite) passes: 287 suites, 2882 tests passed.

Fixes #4383

### DIFF
--- a/app/src/bcsc-theme/features/qr-core/ConnectionLoadingScreen.test.tsx
+++ b/app/src/bcsc-theme/features/qr-core/ConnectionLoadingScreen.test.tsx
@@ -1,15 +1,34 @@
 import { BCSCScreens, BCSCStacks } from '@/bcsc-theme/types/navigators'
-import { Connection } from '@bifold/core'
+import { Connection, CredentialProvisioningEventTypes, LoadingPlaceholder } from '@bifold/core'
 import { CommonActions } from '@react-navigation/native'
-import { render } from '@testing-library/react-native'
+import { act, render } from '@testing-library/react-native'
 import React from 'react'
-import { BackHandler } from 'react-native'
+import { BackHandler, DeviceEventEmitter } from 'react-native'
 
 import ConnectionLoadingScreen from './ConnectionLoadingScreen'
+
+const mockProvisioningMonitor = { workflowInProgress: false }
 
 jest.mock('react-i18next', () => ({ useTranslation: () => ({ t: (k: string) => k }) }))
 jest.mock('@bifold/core', () => ({
   Connection: jest.fn().mockReturnValue(null),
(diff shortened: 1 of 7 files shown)


jleach:
This work will be folded into #4408. Closing.

PR #4457, opened 2026-08-12
fix(bcsc): use dark foreground on the warning banner for WCAG contrast

### LINKED ISSUE #4200: Device Management banner - Contrast does not conform WCAG standards
### Description of problem

Image

### Environment
PROD

Build #:
7182

Android Device Model:
Pixel 6 (Android 16)

iOS Device Model:

opensource-joe:
# Summary of Changes

The warning banner renders white text and icons on its amber `#F8BB47` background. That is a contrast ratio of **1.72:1** — it fails the WCAG 2.1 AA minimum of 4.5:1 for text, and also the 3:1 minimum for the icon, which carries meaning here.

There was already a `type === 'warning'` branch in `AppBanner.tsx` trying to special-case this, but it reached for `ColorPalette.brand.secondaryBackground`, which is theme-dependent:

| Theme | `secondaryBackground` | Contrast on `#F8BB47` |
| --- | --- | --- |
| DarkTheme | `#01264C` | 8.82:1 ✅ |
| **LightTheme** | **`#FFFFFF`** | **1.72:1 ❌** |

BCSC builds boot on LightTheme (`defaultThemeName` in `App.tsx` resolves to `BCThemeNames.Light` when `BUILD_TARGET === Mode.BCSC`), so in the shipped app **both sides of that ternary resolved to white** and the special case silently did nothing. That matches the screenshot on #4200.

This switches the warning foreground to `ColorPalette.grayscale.darkGrey` (`#313132`) — the palette's off-black, and already the light theme's usual body text colour (`foreground`, `textOnWhite`, `onboardingBodyText` all use it). That gives **7.54:1**, clearing AA and AAA. The banner background is a fixed hex rather than a theme token, so this is stable across themes.

The colour is now resolved once into a `foregroundColor` const shared by the icon, title and description, rather than the same ternary repeated three times — this bug is exactly the sort that duplication invites.

This aligns with @knguyenBC's direction on the issue ("the text colour for the warning banner should be an off-black"). I could not open the linked Figma BCSC style guide, so **if the style guide specifies a particular off-black, that value should win over my choice of `darkGrey`** — the swap is a one-line change. Flagging too that the issue carries `needs/ux-wireframe`; happy to hold for UX sign-off if that label is still live.

# Testing Instructions

Automated:

```
cd app
yarn test src/bcsc-theme/components/AppBanner
```

The new `AppBanner.contrast.test.tsx` renders the warning banner and asserts the computed contrast ratio of the title, description and icon against `#F8BB47` clears the WCAG thresholds. It fails on `main` (reports exactly 1.72:1) and passes with this change.

Worth knowing why that test mocks `useTheme`: rendered bare, `AppBannerSection` picks up Bifold's built-in default theme, whose `secondaryBackground` happens to be dark (`#313132`) — so a naive test passes even with the bug present. The test pins the app's real `LightTheme` instead, which is what surfaces the defect.

Manual:

1. Run a BCSC build and trigger a `warning`-type banner (e.g. the Device Management / device limit banner).
2. Confirm the title, description and alert icon render in dark grey on the amber background and are comfortably legible.
3. Confirm `error`, `info` and `success` banners are unchanged (still white on their darker backgrounds).

# Acceptance Criteria

- [x] Warning banner text meets WCAG 2.1 AA (>= 4.5:1) against its background — now 7.54:1
- [x] Warning banner icon meets the non-text minimum (>= 3:1) — now 7.54:1
- [x] `error` / `info` / `success` banners visually unchanged
- [x] Regression test that fails without the fix
- [x] `yarn test src/bcsc-theme` passes (239 suites, 2401 tests), `eslint` and `tsc --noEmit` clean

N/A — I don't have a BCSC device/simulator build available, so I have not captured before/after screenshots of the running app. The change is verified numerically instead (contrast ratios computed from the exact hex values above, asserted in the new test). **A visual check against a real BCSC build is worth doing before merge**, particularly to confirm the dark grey reads well against the amber in situ.

# Related Issues

Fixes #4200

### DIFF
--- a/app/src/bcsc-theme/components/AppBanner.contrast.test.tsx
+++ b/app/src/bcsc-theme/components/AppBanner.contrast.test.tsx
@@ -1,0 +1,76 @@
+import { testIdWithKey } from '@bifold/core'
+import { render } from '@testing-library/react-native'
+import React from 'react'
+import { StyleSheet } from 'react-native'
+
+import { AppBannerSection, BCSCBanner } from './AppBanner'
+
+// AppBanner reads its colours from useTheme(). Rendered bare, that resolves to
+// Bifold's built-in default theme rather than the theme the app actually ships,
+// which is what let the original contrast bug through: the default theme's
+// `secondaryBackground` happens to be dark, while the app's LightTheme sets it
+// to white. BCSC builds boot on LightTheme (see `defaultThemeName` in App.tsx),
+// so pin the real thing here.
+jest.mock('@bifold/core', () => {
+  const actual = jest.requireActual('@bifold/core')
+
+  // Resolved lazily: '@/theme' pulls in '@bifold/core' itself, so requiring it
+  // in the factory body would deadlock on the partially-initialised module.
+  return { ...actual, useTheme: () => jest.requireActual('@/theme').LightTheme }
+})
+
+/**
+ * WCAG 2.1 relative luminance and contrast ratio.
+ * https://www.w3.org/TR/WCAG21/#dfn-relative-luminance
+ */
+const relativeLuminance = (hex: string): number => {
+  const channels = [1, 3, 5].map((offset) => parseInt(hex.slice(offset, offset + 2), 16) / 255)
+  const [r, g, b] = channels.map((c) => (c <= 0.03928 ? c / 12.92 : Math.pow((c + 0.055) / 1.055, 2.4)))
+
+  return 0.2126 * r + 0.7152 * g + 0.0722 * b
+}
+
+const contrastRatio = (foreground: string, background: string): number => {
+  const [lighter, darker] = [relativeLuminance(foreground), relativeLuminance(background)].sort((a, b) => b - a)
+
+  return (lighter + 0.05) / (darker + 0.05)
+}
+
+const colorOf = (element: { props: { style: unknown } }): string =>
+  (StyleSheet.flatten(element.props.style as never) as { color: string }).color
+
+// Matches the `warning` case of AppBanner's bannerColor()
+const WARNING_BACKGROUND = '#F8BB47'
+
+const WCAG_AA_TEXT = 4.5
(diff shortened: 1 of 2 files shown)


opensource-joe:
Signed off the commit and force-pushed — DCO passes now. Sorry for the noise; I missed that on the way in. Worth noting it isn't mentioned in `CONTRIBUTING.md`, the PR template or the README, so it's only discoverable once the check fails. Happy to add a line about it in a separate PR if that'd be useful.

On the `do not merge` label — no rush from me, and I don't want to push this along. If it's the UX question I raised in the description, that seems right to me: #4200 carries `needs/ux-wireframe`, and the linked BCSC style guide (which I can't open) may well name a specific off-black that should win over the `grayscale.darkGrey` I picked. It's a one-token change if so.

If it's held for something else, or you'd rather take this in-house, just say and I'll close it — no work wasted either way.

opensource-joe:
@jleach Two small things on this one, and the first matters more than the second.

**The `do not merge` label.** It was added a couple of days ago without a note, so I do not know whether it reflects something wrong with this change, a release window, or a pending decision about the BCSC theme. If there is something you want changed here I am glad to do it. If it is just timing, no action needed and I will leave it alone. Either way I would rather ask than guess.

**CI has not actually run.** As a first-time contributor here my workflows need a maintainer to release them, so four runs are sitting at `action_required`. This one is easy to miss because the PR reads as green: DCO and both SonarCloud checks report and pass, so nothing looks blocked unless you check the workflow runs directly. If the label is only about timing, approving those would let anything real surface now rather than later.

The branch is also behind the base now. Say the word and I will rebase.

For context on what is waiting, this fixes the BCSC warning banner using a light foreground on a light background, which fell below WCAG AA. The fix moves it to the dark foreground the themed component already provides rather than hardcoding a colour.

PR #4465, opened 2026-08-13
fix: upgrade react-native-incall-manager to 4.2.2 to fix Android call crash

### LINKED ISSUE #4438: App crashes when starting live video verification call on Android

## Issue

A small number of Android users are experiencing an app crash when starting the live video verification call in BC Services Card. The crash happens as the call screen opens, which means the person loses their in-progress verification and has to restart the app and begin the flow again.

By the time a user reaches this screen they have already granted camera, microphone and Bluetooth permissions and uploaded their identity documents, so the cost of the crash to the user is high — the whole verification attempt is lost.

The crash is timing-dependent (a race between two pieces of audio setup work), so it affects only a fraction of call attempts rather than everyone.

**Affected version:** 4.0.2 (Android)

## Exception

```
Exception java.util.ConcurrentModificationException:
  at java.util.HashMap$HashIterator.nextNode (HashMap.java:1604)
  at java.util.HashMap$KeyIterator.next (HashMap.java:1627)
  at com.zxcpoiu.incallmanager.InCallManagerModule.getAudioDeviceStatusMap (InCallManagerModule.java:1856)
  at com.zxcpoiu.incallmanager.InCallManagerModule.lambda$updateAudioDeviceState$3 (InCallManagerModule.java:1847)
  at com.zxcpoiu.incallmanager.InCallManagerModule.$r8$lambda$coo-J0VvxOmwyoT_lR_5RlLwOBQ (InCallManagerModule.java)
  at com.zxcpoiu.incallmanager.InCallManagerModule$$ExternalSyntheticLambda3.run (D8$$SyntheticClass)
  at android.os.Handler.handleCallback (Handler.java:1070)
  at android.os.Handler.dispatchMessage (Handler.java:125)
  at android.os.Looper.dispatchMessage (Looper.java:358)
  at android.os.Looper.loopOnce (Looper.java:288)
  at android.os.Looper.loop (Looper.java:392)
  at android.app.ActivityThread.main (ActivityThread.java:10346)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:638)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:972)
```

## Root cause

This is a defect in the third-party `react-native-incall-manager` library (version `~4.2.1`), not in our own code. Our call sequence reliably creates the conditions for it. The stack trace lines match the resolved 4.2.1 source exactly (`getAudioDeviceStatusMap` iteration at `InCallManagerModule.java:1856`, posted from `updateAudioDeviceState` at `:1847`), confirming the crashing artifact is the version we ship.

The library keeps a list of available audio devices (speaker, earpiece, Bluetooth headset, wired headset) in a plain `HashSet` that is not safe for use from two threads at once. When we open the call screen, `LiveCallScreen.tsx:229-230` makes two calls back to back:

```ts
InCallManager.setForceSpeakerphoneOn(false)
InCallManager.start({ media: 'video', auto: true })
```

- `setForceSpeakerphoneOn(false)` maps a boolean `false` to the library's "force speaker off" flag (`-1`), which selects the earpiece. When the earpiece is already in the device list, that schedules work on Android's main UI thread which reads through the audio device list to broadcast an `onAudioDeviceChanged` event.
- `start()` runs on React Native's native modules thread and empties that same list (`audioDevices.clear()` at `InCallManagerModule.java:575`).

When the second call empties the list while the first is still reading it, Android throws `ConcurrentModificationException`. Because it happens on the main UI thread and nothing catches it, the app terminates rather than degrading gracefully.

Two things narrow the window, which is why the crash is intermittent rather than universal:

- The two calls are only one line apart.
- The crash cannot happen on the very first call attempt in a fresh app process. The device list starts empty and is only populated once the library has run a device-state update; `stop()` does not clear it, so the list survives call teardown. The exposed cases are a second or later call attempt in the same process, or a first attempt where a wired-headset or Bluetooth event already populated the list.

Related finding in our own code: the comment above these two lines (`LiveCallScreen.tsx:226-228`) says the call "clears the forced-speaker override". It does not — passing boolean `false` *forces the earpiece*; clearing the override requires a non-boolean argument (flag `0`). The comment should be corrected in the same change; whether forcing the earpiece is actually the intended audio behaviour is a separate question and out of scope here.

## Proposed fix

**Upgrade `react-native-incall-manager` to 4.2.2.**

The library maintainers fixed this exact defect upstream — the relevant commits are titled "fix: ConcurrentModificationException in getAudioDeviceStatusMap" (PR react-native-webrtc/react-native-incall-manager#260) and shipped in 4.2.2 (published 2026-07-24). Their fix uses a thread-safe collection (`ConcurrentHashMap.newKeySet()`) *and* takes a defensive copy before reading the list.

Points in favour of this approach:

- Our dependency range is already `~4.2.1`, which admits 4.2.2, so this is a lockfile-only update with no package.json or source changes on our side.
- The upstream change touches Android only.
- We reviewed the rest of the 4.2.2 changes for regression risk. The behaviour changes are confined to ringtone handling, audio-setup locking, and a bump of the library's *default* `minSdkVersion` from 21 to 24. We do not use any of the ringtone APIs — our only usage is `setForceSpeakerphoneOn`, `start`, and `stop` — and our app already sets `minSdkVersion = 24`, so the default bump has no effect. That leaves no meaningful regression surface for us.

Considered and rejected:

- **Patching the library ourselves** via `yarn patch`. Viable as a fallback if the upgrade turns out to be blocked, but unnecessary when the fix is already published.
- **Removing the dependency.** Not an option; it is a real direct dependency with live usage in the video verification flow, and it is not merely a transitive requirement of `react-native-webrtc`.
- **Dropping or reordering our two JS calls.** Dropping `setForceSpeakerphoneOn(false)` would narrow the window we create ourselves, but it would not close the race — a Bluetooth headset connecting mid-call can still trigger it. The call ordering itself should be left alone (only the inaccurate comment fixed, per above).

## Verification

- Confirm the resolved version is 4.2.2 after the upgrade.
- Smoke test the BC Services Card live video verification call on Android: audio routing through speaker, wired headset, and Bluetooth headset, plus connecting and disconnecting a Bluetooth headset during an active call. Include a *second* call attempt in the same app session — that is the scenario the original race actually exposed.
- Monitor Play Store crash reporting for recurrence of `getAudioDeviceStatusMap` after release.

jleach:
## What

Some Android users hit an app crash the moment the live video verification call screen opened. When it happened, the person lost their in-progress identity verification — camera, microphone and Bluetooth permissions already granted, identity documents already uploaded — and had to restart the app and begin the whole flow again. This change picks up the calling library's official fix for that crash, and corrects a misleading comment in the call screen code.

## Why

The crash came from a defect in `react-native-incall-manager`, the third-party library the video verification call screen uses to manage call audio. Two pieces of audio setup that run back-to-back when the screen opens could collide on the library's internal list of audio devices, occasionally terminating the app. The library's maintainers have published a fix for exactly this defect. Taking it up closes the crash window with no change to our own call logic and no change to how audio behaves for the user.

## Decisions

- **Version bump only — no patch, no source workaround.** Our dependency range (`~4.2.1`) already admits 4.2.2, so this is a lockfile-only re-resolution. No `package.json` change and no `yarn patch` were needed.
- **The fix was confirmed in the resolved artifact**, not just in the version number: the installed `InCallManagerModule.java` uses a thread-safe device set (`ConcurrentHashMap.newKeySet()`) and takes a defensive copy before iterating in `getAudioDeviceStatusMap`. A diff of 4.2.1 against 4.2.2 shows the only Android changes are that thread-safety work plus a guard around ringtone teardown — no change to audio routing behaviour, so there is no regression surface for our usage.
- **iOS `Podfile.lock` regenerated** (`ReactNativeIncallManager` 4.2.1 → 4.2.2) so the pod version tracks the npm package, even though the underlying fix is Android-only.
- **The two `InCallManager` calls and their order are untouched.** Reordering or removing them would narrow but not close the race, and the question of what audio route the call screen *should* use is a separate one, deliberately out of scope here.
- **Comment-only change in `LiveCallScreen.tsx`.** The old comment said `setForceSpeakerphoneOn(false)` "clears the forced-speaker override" and thereby "allows audio to route to Bluetooth/wired headsets". Neither is true: passing boolean `false` maps to flag `-1`, which *forces the earpiece* (flag `0` is what clears the override), and `start()` re-initialises the library's routing state immediately afterwards, so the preceding call does not determine the initial route at all. The comment now describes the verified behaviour.
- **`gradle.lockfile` unchanged**, confirmed by running the Android setup step — 4.2.2 introduces no new Android dependencies.

## Tests

- No new automated tests: the defect is in third-party Android native code that Jest cannot reach, and the only change to our own source is a comment.
- `yarn check` (typecheck, lint, format, full Jest suite — 293 suites / 3031 tests / 162 snapshots) passes clean, with no snapshot changes.
- `yarn why react-native-incall-manager` confirms the resolved version is 4.2.2.

## Verification

Automated verification (typecheck, lint, format, tests, dependency resolution, and confirmation that the fix is present in the installed library) is complete and passing — see Tests above.

**Outstanding — manual testing on a real Android device is still required before merge:**

- Live video verification call in BC Services Card (`BUILD_TARGET=bcsc`) with audio routed through the speaker, a wired headset, and a Bluetooth headset.
- Connecting and disconnecting a Bluetooth headset during an active call.
- A **second** call attempt within the same app session — this is the scenario that originally exposed the crash, since the library's audio device list starts empty on the first call and only fills in after some device activity.
- BC Wallet target (`BUILD_TARGET=bcwallet`) — build-success check only; that target does not use this library at runtime.
- After release: watch Play Store crash reporting for any recurrence of `getAudioDeviceStatusMap`.

## Follow-up

While verifying the comment against the library source, it became clear that `setForceSpeakerphoneOn(false)` before `start()` has no effect on the call's initial audio route — `start()` resets the library's routing state immediately afterwards. Removing that call is tracked separately in #4471. It is kept out of this PR deliberately: the library's iOS and JavaScript sources are byte-identical between 4.2.1 and 4.2.2, so this change needs only an Android device pass, whereas touching the routing call would add a full iOS pass on top.

Fixes #4438

### DIFF
--- a/app/ios/Podfile.lock
+++ b/app/ios/Podfile.lock
@@ -2836,7 +2836,7 @@ PODS:
     - React-perflogger (= 0.81.5)
     - React-utils (= 0.81.5)
     - SocketRocket
-  - ReactNativeIncallManager (4.2.1):
+  - ReactNativeIncallManager (4.2.2):
     - React-Core
   - RNCAsyncStorage (2.2.0):
     - boost
@@ -3858,7 +3858,7 @@ SPEC CHECKSUMS:
   ReactAppDependencyProvider: 1bcd3527ac0390a1c898c114f81ff954be35ed79
   ReactCodegen: 6c26f8c25d0b5ae66f86a1cce1777076ac8bcbd8
   ReactCommon: 5f0e5c09a64a2717215dd84380e1a747810406f2
-  ReactNativeIncallManager: dccd3e7499caa3bb73d3acfedf4fb0360f1a87d5
+  ReactNativeIncallManager: c6e08cc005da0f73d5a1ecbd58ad33bc61d5d79f
   RNCAsyncStorage: 29f0230e1a25f36c20b05f65e2eb8958d6526e82
   RNCClipboard: 4b58c780f63676367640f23c8e114e9bd0cf86ac
   RNDeviceInfo: d3e91ffb33ee97a7982108476edb68cb3672efa6
--- a/app/src/bcsc-theme/features/verify/live-call/LiveCallScreen.tsx
+++ b/app/src/bcsc-theme/features/verify/live-call/LiveCallScreen.tsx
@@ -223,11 +223,8 @@ const LiveCallScreen = ({ navigation }: LiveCallScreenProps) => {
   useEffect(() => {
     if (flowState === VideoCallFlowState.IDLE) {
       startVideoCall()
-      // Passing boolean false maps to force-speaker flag -1 (speaker forced OFF,
-      // earpiece as default route) — it does NOT clear the override (a non-boolean
-      // arg / flag 0 does that). We set it before start() so start({ media: 'video' })
-      // does not force audio to the speaker on Android, allowing routing to
-      // Bluetooth/wired headsets.
+      // Boolean false maps to force-speaker flag -1 (force earpiece on Android); it does NOT clear
+      // the override (flag 0 does). start() re-initialises routing, so this does not set the initial route.
       InCallManager.setForceSpeakerphoneOn(false)
       InCallManager.start({ media: 'video', auto: true })
     }
--- a/app/src/bcsc-theme/features/verify/live-call/LiveCallScreen.tsx
+++ b/app/src/bcsc-theme/features/verify/live-call/LiveCallScreen.tsx
@@ -223,8 +223,7 @@ const LiveCallScreen = ({ navigation }: LiveCallScreenProps) => {
   useEffect(() => {
     if (flowState === VideoCallFlowState.IDLE) {
       startVideoCall()
-      // Boolean false maps to force-speaker flag -1 (force earpiece on Android); it does NOT clear
-      // the override (flag 0 does). start() re-initialises routing, so this does not set the initial route.
+      // No-op: start() re-initialises audio routing immediately after. Removal tracked in #4471.
       InCallManager.setForceSpeakerphoneOn(false)
       InCallManager.start({ media: 'video', auto: true })
     }
--- a/app/src/bcsc-theme/features/verify/live-call/LiveCallScreen.tsx
+++ b/app/src/bcsc-theme/features/verify/live-call/LiveCallScreen.tsx
@@ -223,9 +223,11 @@ const LiveCallScreen = ({ navigation }: LiveCallScreenProps) => {
   useEffect(() => {
     if (flowState === VideoCallFlowState.IDLE) {
       startVideoCall()
-      // Clear the forced-speaker override before start() so that start()
-      // does not force audio to the speaker on Android. This allows audio
-      // to route to Bluetooth/wired headsets
+      // Passing boolean false maps to force-speaker flag -1 (speaker forced OFF,
+      // earpiece as default route) — it does NOT clear the override (a non-boolean
+      // arg / flag 0 does that). We set it before start() so start({ media: 'video' })
+      // does not force audio to the speaker on Android, allowing routing to
+      // Bluetooth/wired headsets.
       InCallManager.setForceSpeakerphoneOn(false)
       InCallManager.start({ media: 'video', auto: true })
     }
--- a/yarn.lock
+++ b/yarn.lock
@@ -16583,11 +16583,11 @@ __metadata:
   linkType: hard
 
 "react-native-incall-manager@npm:~4.2.1":
-  version: 4.2.1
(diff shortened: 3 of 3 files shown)


al-rosenthal (approved):
I tested this out a lot. No crashes on multiple calls, it handles swapping Bluetooth devices mid call. The only thing I'm unable to test is a headset wired to the phone.

PR #4473, opened 2026-08-13
fix: add build number to Need help menu app version

### LINKED ISSUE #4429: The app version in the "Need help?" menu doesn't have the build number
### Description of problem
The app version in the "Need help?" menu doesn't have the build number. The build number should be available here just in case a person can't access their settings.

### Expected behavior
Should have the full version number and build

### Environment

Build #: 4.1.0 (7840)

Android Device Model: Google pixel 8

iOS Device Model: iPhone 12

### Workaround
The build number is in settings

jleach:
## What

The app version shown in the "Need help?" floating menu now includes the build number (e.g. `App version: 4.1.0 (7840)`), not just the version number.

## Why

Support sometimes needs the build number to diagnose an issue, and the "Need help?" menu is meant to be reachable even when someone can't get to Settings. Without the build number there, the workaround was digging into Settings — the exact thing the help menu should avoid requiring.

## Decisions

- Reused the existing `getBuildNumber()`/`getVersion()` pattern from `react-native-device-info` already used in Settings, the Developer screen, and `ErrorInfoCard`, rather than introducing a shared version/build utility — this keeps the change minimal and consistent with existing usage elsewhere in the codebase.
- Added a `build` interpolation variable to the existing `BCSC.HelpMenu.Version` i18n string in `en`/`fr`/`pt-br`, matching the `Version X.Y.Z (build)` paren format already shown in Settings.
- Scoped to the `bcsc` build target only — `FloatingHelpMenu` is a BCSC-theme component, not used by `bcwallet`.

## Tests

Added `app/src/bcsc-theme/components/FloatingHelpMenu.test.tsx` with two tests:

1. **Render test** — asserts the help menu renders `App version: 4.0.0 (142)` using the fixed values from the `react-native-device-info` jest mock, confirming the component wires `version`/`build` into the translation call. (The global `react-i18next` mock drops interpolation, so this test uses a local interpolating mock override, matching the pattern in `AgentReadyGate.test.tsx`.)
2. **Locale-string test** — imports the real `en`/`fr`/`pt-br` locale objects and asserts each `BCSC.HelpMenu.Version` string contains the `{{ build }}` placeholder (and `{{ version }}` for `en`). This guards the locale-file half of the change: reverting the locale copy back to version-only now fails the suite.

`yarn check` (typecheck, lint, format, full test suite) passes: 294 suites / 3034 tests / 162 snapshots.

## Verification

1. Open the app with `BUILD_TARGET=bcsc`.
2. From any screen with the floating "Need help?" button, open the help menu.
3. Confirm the version line at the bottom reads `App version: <version> (<build>)`, matching the build number shown in Settings.

Fixes #4429

### DIFF
--- a/app/src/bcsc-theme/components/FloatingHelpMenu.test.tsx
+++ b/app/src/bcsc-theme/components/FloatingHelpMenu.test.tsx
@@ -3,10 +3,7 @@ import { render, screen } from '@testing-library/react-native'
 import React from 'react'
 import FloatingHelpMenu from './FloatingHelpMenu'
 
-// The global react-i18next mock (see app/__mocks__/react-i18next.ts) drops interpolation
-// values and returns the raw key, which would hide a regression in the version/build
-// interpolation this test is meant to catch. Override locally with a mock that actually
-// interpolates the `version`/`build` options into the real Version copy.
+// The global react-i18next mock returns raw keys, so interpolate locally to assert version/build.
 jest.mock('react-i18next', () => ({
   useTranslation: () => ({
     t: (key: string, options?: Record<string, string>) => {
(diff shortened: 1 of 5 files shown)


al-rosenthal (approved):

PR #4474, opened 2026-08-13
fix: update links in 4.1

### LINKED ISSUE #4458: Fix links in 4.1
### Links to fix

The "Learn more about the wallet" button should link here: https://id.gov.bc.ca/static/help/topics.html 

Image

___

The "Support" link in the Report a problem drawer should link to: https://id.gov.bc.ca/static/help/contact-us.html
 
Image

___

The "Contact us" button in the menu should be a webview of this link: https://id.gov.bc.ca/static/help/contact-us.html
The "Accessibility" button in the menus should navigate users to this link: https://www2.gov.bc.ca/gov/content/governments/government-id/bcservicescardapp#accessibility

Image

jleach:
## What

Four links in the BC Services Card app pointed at the wrong or outdated pages. This fixes them:

- "Learn more about the wallet" now opens the correct BC Services Card help topics page.
- "Support" (in the Report a problem drawer) now opens the correct contact-us help page.
- "Contact us" (in the menu) now opens the contact-us help page in-app, instead of a separate static screen with outdated content.
- "Accessibility" (in the menus) now opens the BC Services Card government page, which has an Accessibility section.

## Why

Several links in the 4.1 line pointed at stale or incorrect destinations, so users looking for help, support, or accessibility information could land on the wrong page. This aligns every link with the current BC Services Card help centre and government pages.

## Decisions

- Added `?fromapp=1` to the in-app Contact us WebView URL, matching every other in-app `HelpCentreUrl` webview link — it strips the government site's header/footer chrome for a cleaner in-app view.
- Consolidated all remaining references to the old `www2` contact-us page (the Report a problem "Support" link, `ErrorInfoCard`, and `deviceAuthorizationError` device-error screens) onto the same new help-centre contact-us page, via a new `CONTACT_US_HELP_URL` constant.
- Since "Contact us" now opens the WebView instead of the dedicated `ContactUsScreen`, that screen, its four per-stack `*ContactUs` routes, and its locale strings were dead code and removed in this same PR rather than as a follow-up.

## Tests

- `yarn check` (typecheck, lint, format, full test suite) passes with no snapshot changes.
- New/updated tests assert the literal URLs (not the constants) for the four fixed links, so a future regression in the constant's value would be caught rather than passing tautologically.

Fixes #4458


MacDeluca (commented):
Some things I noticed (possibly out of scope of this ticket)

- MainStack: Settings -> Give us feedback / Accessibility / Terms of use
    - no icon indicating user is leaving the app (ie: "contact us" is a webview and these are external links)
    - need to decide if they should be webviews or external links, if swapping to webviews append `?fromapp=1` where applicable

- Report a problem uses `CONTACT_US_HELP_URL` 
    - no icon indicating user is leaving the app

- Missing `fromapp` query param:
    - BC_LOGIN_PRIVACY_URL

- Deprecated:
    - WHERE_TO_USE_URL
    - ACCOUNT_SERVICES_URL

jleach:
@MacDeluca I fixed the `fromapp` and deprecated URLs. Talked the first two points through with Kim. If we open a WebView - no link icon. If we open the device browser (Chrome / Safari) then yes, a link icon is needed. I'm fixing feedback, Accessibility, and Terms of use to have the icon and removing it from the Privacy -> Learn More button since that is a WebView.

MacDeluca (approved):
Love it

PR #4477, opened 2026-08-14
fix: run device-security unlock off the main thread on Android

### LINKED ISSUE #4440: App becomes unresponsive after biometric unlock on Android

Unrelated to the other reports filed under #4435 (#4438, #4439) — this one is in our own code and is an unresponsive-app report rather than a crash.

## Issue

After a user successfully authenticates with biometrics or device PIN to unlock BC Services Card, the app can freeze long enough for Android to show the "app isn't responding" dialog.

This lands at the worst possible moment: the person has just proven who they are and is waiting to get into the app. On slower devices, or when the device's secure key store is busy with other work, the freeze can be long enough that they cannot get in at all without retrying.

This is not a crash — nothing fails. The app is doing real work, just on the wrong thread, so the screen stops responding while it finishes.

**Affected version:** 4.0.2 (Android)

**Priority note:** we would rank this above the other 4.0.2 reports. The others cost the user a flow or a session; this one can block entry to the app.

## Report

Android captured a full thread dump (all threads) because the app was unresponsive. The relevant part is the main thread, which is blocked waiting on the device's secure key store:

```
"main" tid=1 Native
  #00  pc 0x00000000000de9c8  /apex/com.android.runtime/lib64/bionic/libc.so (__ioctl+8)
  #01  pc 0x0000000000082c1c  /apex/com.android.runtime/lib64/bionic/libc.so (ioctl+156)
  #02  pc 0x000000000005f194  /system/lib64/libbinder.so (android::IPCThreadState::transact+1428)
  #03  pc 0x0000000000085428  /system/lib64/libbinder.so (android::BpBinder::transact+280)
  #04  pc 0x00000000001d72e8  /system/lib64/libandroid_runtime.so (android_os_BinderProxy_transact+152)
  at android.os.BinderProxy.transactNative (Native method)
  at android.os.BinderProxy.transact (BinderProxy.java:602)
  at android.system.keystore2.IKeystoreSecurityLevel$Stub$Proxy.generateKey (IKeystoreSecurityLevel.java:307)
  at android.security.KeyStoreSecurityLevel.lambda$generateKey$0 (KeyStoreSecurityLevel.java:170)
  at android.security.KeyStoreSecurityLevel.$r8$lambda$nmKQL67jdmJaQALlqay99_HTGYQ (unavailable)
  at android.security.KeyStoreSecurityLevel$$ExternalSyntheticLambda2.execute (D8$$SyntheticClass)
  at android.security.KeyStoreSecurityLevel.retryBusyException (KeyStoreSecurityLevel.java:71)
  at android.security.KeyStoreSecurityLevel.generateKey (KeyStoreSecurityLevel.java:170)
  at android.security.keystore2.AndroidKeyStoreKeyGeneratorSpi.engineGenerateKey (AndroidKeyStoreKeyGeneratorSpi.java:402)
  at javax.crypto.KeyGenerator.generateKey (KeyGenerator.java:612)
  at androidx.security.crypto.MasterKeys.generateKey (MasterKeys.java:145)
  at androidx.security.crypto.MasterKeys.getOrCreate (MasterKeys.java:100)
  at com.bcsccore.storage.NativeCompatiblePinStorage.getSecret (NativeCompatiblePinStorage.kt:146)
  at com.bcsccore.storage.NativeCompatiblePinStorage.getPinHash (NativeCompatiblePinStorage.kt:311)
  at com.bcsccore.authentication.PinService.getPINHash (PinService.kt:190)
  at com.bcsccore.BcscCoreModule.unlockWithDeviceSecurity$lambda$21 (BcscCoreModule.kt:2674)
  at com.bcsccore.BcscCoreModule.$r8$lambda$TakacDU7WJcmXz6moPsTsr8Xuk0 (unavailable)
  at com.bcsccore.BcscCoreModule$$ExternalSyntheticLambda1.invoke (D8$$SyntheticClass)
  at com.bcsccore.authentication.device.DeviceAuthenticationServiceImpl$performDeviceAuthentication$1$biometricPrompt$1.onAuthenticationSucceeded (DeviceAuthenticationService.kt:122)
  at androidx.biometric.BiometricFragment$9.run (BiometricFragment.java:907)
  at android.os.Handler.handleCallback (Handler.java:1095)
  at android.os.Handler.dispatchMessageImpl (Handler.java:135)
  at android.os.Handler.dispatchMessage (Handler.java:125)
  at android.os.Looper.loopOnce (Looper.java:269)
  at android.os.Looper.loop (Looper.java:367)
  at android.app.ActivityThread.main (ActivityThread.java:9333)
  at java.lang.reflect.Method.invoke (Native method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:566)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:929)
```

Every other thread in the dump was idle and waiting, which confirms the app was not deadlocked or overloaded — it was simply blocked on this one piece of work. Line numbers in the stack trace reflect the 4.0.2 build; the file/line references below are against current `main` (the code involved lives in `packages/bcsc-core`).

## Root cause

The unlock work runs on Android's main (UI) thread, which is the thread responsible for keeping the screen responsive. Android shows the unresponsive-app dialog when that thread is blocked for several seconds.

`BcscCoreModule.kt:2828-2837` performs the PIN hash lookup directly inside the biometric success callback:

```kotlin
deviceAuthenticationService.performDeviceAuthentication(activity, title, subtitle) { authResult ->
    when (authResult) {
        DeviceAuthenticationResult.SUCCESS -> {
            val hashResult = pinService.getPINHash(accountID)   // secure key store work, main thread
```

Main-thread delivery here is explicit, not incidental: `DeviceAuthenticationService.kt` wraps the prompt in `runOnUiThread` and constructs the `BiometricPrompt` with `ContextCompat.getMainExecutor(context)`, so `onAuthenticationSucceeded` always fires on the main thread (visible in the report as `BiometricFragment$9.run` under `Handler.handleCallback` on `tid=1`), and nothing in our chain moves the work to a background thread afterwards — `BcscCoreModule.kt` contains no threading constructs at all.

The work itself is legitimately slow and cannot be made fast:

- `MasterKeys.getOrCreate` asks Android's hardware-backed key store to create or fetch the master key. That runs inside the device's secure hardware and its duration is outside our control. The `retryBusyException` frame shows the system was in its "key store is busy, retrying" loop, so the key store was contended when this happened.
- One branch is considerably heavier. When no stored PIN hash is found (`BcscCoreModule.kt:2845-2854`, the migration path for users coming from v3), the same main-thread callback first checks for an old-format PIN, then generates a PIN and derives a key using **210,000 PBKDF2 iterations** (`NativeCompatiblePinStorage.kt`, `ITERATION_COUNT`), then saves the result — which performs a *second* `MasterKeys.getOrCreate` via `saveSecret`. On a lower-end device the key derivation alone is a multi-second block, before any of the key store round trips are counted.

For reference, the other biometric call site (`BcscCoreModule.kt:2513`) only returns a true/false result, so it does no meaningful work on the main thread and is not affected.

## Proposed fix

**Move the unlock work off the main thread.**

Run the body of the `SUCCESS` branch on a background thread and settle the promise from there. React Native allows promises to be settled from any thread, so nothing about the JavaScript-facing contract changes. Nothing in that branch touches the UI — it is key store access, storage reads and writes, and key derivation — so it is safe to move.

Include in this change:

- A guard so the promise cannot be settled twice. React Native raises an error if that happens. It cannot happen today because the terminal callbacks are mutually exclusive, but running the work asynchronously widens the window enough that this should be explicit rather than assumed.
- A loading state in the UI while the unlock completes. Moving the work off the main thread removes the unresponsive-app dialog, but the user still waits the same amount of time. Without visible progress the screen merely looks stuck instead of triggering a system warning. This is the difference between "no longer reported as unresponsive" and actually feeling fixed.

Implementation note: `bcsc-core` does not currently declare a coroutines dependency and `BcscCoreModule` has no existing asynchronous code, so this change introduces the module's first background-threading pattern. A plain executor or thread keeps it dependency-free. Worth a moment's thought on the approach, since it sets the precedent for later work in this module.

## Out of scope

**Reducing the number of repeated secure key store operations.** The unlock path performs several separate key store round trips rather than one, and none of the intermediate results are reused:

- `NativeCompatiblePinStorage.kt` calls `MasterKeys.getOrCreate` + `EncryptedSharedPreferences.create` from scratch at three call sites (`getSecret`, `saveSecret`, `removePin`), and a single unlock can hit more than one of them (e.g. `getPinHash` → `getSecret`, then legacy-key migration → `saveSecret`).
- `NativeCompatibleStorage.kt` decrypts a fresh file on every read via `readEncryptedFile` — eight call sites in that file plus three more in `BcscCoreModule.kt`. The encryption object itself is lazily cached, but each decrypt still performs its own Android Keystore round trip (`KeyStore.getInstance` + `load` + `getKey` per call).

Reducing this is the larger performance win and would shorten the wait even after the work moves off the main thread. It is a bigger change though: it introduces shared state across many call sites, raises thread-safety questions, and interacts with PIN removal and key rotation. It is not needed to resolve the unresponsive-app reports and should be its own piece of work covering both files.

**Replacing the deprecated `MasterKeys` API** (superseded by `MasterKey.Builder`). Not a contributor here; separate maintenance work.

## Verification

This cannot be covered by unit tests — it needs a real device with biometrics or device PIN enrolled.

- Unlock with biometrics on a physical Android device and confirm the app remains responsive throughout and no unresponsive-app dialog appears.
- Exercise the migration path where no PIN hash is stored yet, since that branch does the heaviest work and benefits most.
- Prefer a lower-end or older device, where the underlying work is slowest and the problem is easiest to reproduce.
- Confirm the loading state is visible during the wait.
- Monitor Play Store reporting for recurrence of unresponsive-app reports on the unlock path after release.

jleach:
## What

After a user unlocks BC Services Card with biometrics or device PIN on Android, the app could freeze long enough for Android to show the "app isn't responding" dialog — right after the person proved who they are and is waiting to get in. On slower devices, or when the device's secure key store is busy, the freeze could be long enough that the person couldn't get in without retrying.

This change moves that work off the main thread so it can no longer trigger the freeze, and keeps the existing loading animation visible while the (unavoidably slow) work completes.

## Why

Unlocking does real security work that can't be skipped or made faster — the app has to prove to the device's secure key store that it may open the user's data, and for people migrating from the v3 app there's extra one-time setup work on top of that.

The problem wasn't that the work was slow; it was where it ran. It was happening on the same thread that draws the screen, so for as long as it took, the screen couldn't update at all. To the user, and to Android, that looks like a frozen app — the "app isn't responding" dialog is Android reacting to exactly that.

Doing this work in the background changes nothing about what the unlock does or how long it takes. The user waits the same amount of time, but now sees the loading animation instead of a frozen screen, and the app is never reported as unresponsive.

## Manual Tesing

- Setup a new install or old with PIN
- Login as expected. 

## Decisions

- **Plain `java.util.concurrent` executor, not coroutines.** `bcsc-core` declares no coroutines dependency today; relying on a transitive copy would be fragile, and this workload is a single hop with nothing to compose. A single dedicated background thread (`BcscCoreBackground`) also gives mutual exclusion over the keystore/PIN-storage work for free — relevant because the v3 migration path performs two separate secure-keystore round trips. This is the first background-threading construct in `BcscCoreModule`, so it sets a precedent for future work in this module; coroutines remain a reasonable choice for anything that later needs real composition.
- **`GuardedPromise` settle-once wrapper.** Running the callback body asynchronously widens the window in which more than one terminal path could try to settle the same React Native promise (RN throws if a promise is settled twice). Every settle path in `unlockWithDeviceSecurity` — including the early rejects, the executor-unavailable case, and the outer catch — now routes through this wrapper.
- **Reject on `RejectedExecutionException` rather than silently dropping.** If the background executor has already been shut down (e.g. a dev reload or activity teardown racing an open biometric prompt), we settle the promise with a reject instead of leaving the JS side waiting forever.
- **`invalidate()` uses `shutdown()`, not `shutdownNow()`.** So a migration write already in progress against the secure key store is never interrupted mid-write.
- **Only the `SUCCESS` branch moves.** `CANCELLED`/`ERROR` do no meaningful work, so they stay on the callback thread. The other biometric call site (`performDeviceAuthentication`, which only returns a boolean) is unaffected and out of scope per the issue.
- **No JS changes.** The existing `BCSCLoadingContext` overlay in `useAuthentication.tsx` already brackets the entire unlock call (`startLoading()` before, `stopLoading()` in a `finally` covering both resolve and reject) and renders an animated loading icon. It was rendering today but starved, because the blocked main thread was also the render thread — unblocking main is what makes it actually animate.
- **Android only.** iOS runs the equivalent post-auth work inside `Task { @MainActor in ... }`, which is a different threading model; that's out of scope for this issue.
- **Out of scope (per the issue):** reducing the number of repeated secure key store round trips (caching in `NativeCompatiblePinStorage.kt`/`NativeCompatibleStorage.kt`) and replacing the deprecated `MasterKeys` API — both larger, separate pieces of work.

## Tests

New `GuardedPromiseTest.kt` (JUnit4 + mockk) covers: resolve-then-resolve settles once, resolve-then-reject and reject-then-resolve don't cross-settle, both `reject` overloads (with/without throwable) dispatch correctly, and a concurrent race between two threads settling simultaneously results in exactly one delegate call.

No new JS tests — there is no JS-visible change; the existing `useAuthentication.test.tsx` and the full Jest/Android suites were re-verified and still pass.

## Verification

- `packages/bcsc-core`: `yarn build`, `yarn test:android`, `yarn lint:kotlin` all pass.
- Root `yarn check` (typecheck, lint, format, full Jest suite) passes.
- `app`: `yarn android:setup` produces no lockfile diff (no dependency was added).
- Confirmed in code (not just assumed) that the loading overlay covers both the resolve and reject paths of the unlock call — see Decisions above.
- This fix cannot be verified by unit tests; it needs a physical Android device with biometrics/device PIN enrolled. Manual on-device verification (by the requester, not part of this PR):
  - Unlock on a real device and confirm no ANR dialog appears.
  - Exercise the v3 migration path specifically (no stored PIN hash) — it does the heaviest work (210,000-iteration key derivation).
  - Confirm the loading animation is visible for the duration of the unlock.
  - Prefer testing on a lower-end device where the effect is most pronounced.
  - Confirm the app stays responsive on a device with existing account and issuer files on disk, so all three secure-keystore round trips are actually exercised.

Fixes #4440

### DIFF
--- a/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
+++ b/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
@@ -2812,15 +2812,8 @@ class BcscCoreModule(
         reason: String?,
         promise: Promise,
     ) {
-        // The biometric SUCCESS callback below fires on the main thread
-        // (DeviceAuthenticationServiceImpl uses runOnUiThread + getMainExecutor). Its work
-        // performs multiple secure key store round trips (issuer/account file decryption, PIN
-        // hash lookup, and — on the v3 migration path with no stored PIN hash — a 210,000
-        // iteration PBKDF2 derivation plus a second key store write) that can take long enough
-        // to trigger an ANR. That branch is moved onto backgroundExecutor so the main thread is
-        // never blocked; every settle path is routed through a GuardedPromise since running the
-        // work asynchronously widens the window in which more than one terminal callback path
-        // could attempt to settle the same promise.
+        // The biometric callback fires on the main thread, so the SUCCESS branch's keystore work
+        // runs on backgroundExecutor; async settling means every path must go through GuardedPromise.
         val guarded = GuardedPromise(promise)
 
         try {
@@ -2852,7 +2845,6 @@ class BcscCoreModule(
                         try {
                             backgroundExecutor.execute {
                                 try {
-                                    // Get the stored PIN hash
                                     val hashResult = pinService.getPINHash(accountID)
 
                                     if (hashResult != null) {
@@ -2887,10 +2879,7 @@ class BcscCoreModule(
                                 }
                             }
                         } catch (e: RejectedExecutionException) {
-                            // The executor has already been shut down (e.g. invalidate() ran
-                            // during a dev reload or activity teardown while the biometric
-                            // prompt was still open). Settle rather than leave the JS promise
-                            // pending forever.
+                            // Executor shut down while the prompt was open; settle so JS is not left pending.
                             Log.w(NAME, "unlockWithDeviceSecurity: background executor unavailable, rejecting", e)
                             guarded.reject(
                                 "E_UNLOCK_DEVICE_SECURITY_ERROR",
@@ -4147,11 +4136,8 @@ class BcscCoreModule(
         PinService(reactApplicationContext, nativeStorage)
     }
 
-    // Single-thread executor for offloading main-thread-unsafe work triggered from
-    // biometric/device-authentication callbacks (see unlockWithDeviceSecurity). Deliberately a
-    // plain java.util.concurrent executor rather than coroutines: bcsc-core declares no
-    // coroutines dependency, the workload here is a single hop with nothing to compose, and one
-    // named serialized thread gives mutual exclusion over keystore/PIN-storage work for free.
+    // Serialized off-main thread for keystore/PIN work reached from biometric callbacks.
+    // A plain executor, not coroutines: bcsc-core declares no coroutines dependency.
     private val backgroundExecutorDelegate =
         lazy {
             Executors.newSingleThreadExecutor { r -> Thread(r, "BcscCoreBackground") }
@@ -4159,10 +4145,8 @@ class BcscCoreModule(
     private val backgroundExecutor: ExecutorService by backgroundExecutorDelegate
 
     /**
-     * Shuts down [backgroundExecutor] when the module is torn down (dev reload, activity
-     * teardown). Checks [Lazy.isInitialized] first so tearing down a module that never used the
-     * executor doesn't create one just to shut it down. Uses shutdown() rather than
-     * shutdownNow() so an in-flight migration write is never interrupted mid-keystore-write.
+     * Shuts the background executor down on teardown, without initializing one that was never used.
+     * shutdown(), not shutdownNow(): an in-flight migration must never be interrupted mid-keystore-write.
      */
     override fun invalidate() {
         if (backgroundExecutorDelegate.isInitialized()) {
--- a/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
+++ b/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
@@ -157,6 +157,12 @@ class BcscCoreModule(
         // Notification channel constants
         private const val NOTIFICATION_CHANNEL_ID = "bcsc_foreground_notifications"
         private const val NOTIFICATION_CHANNEL_NAME = "BCSC Notifications"
+
+        // Auto-generated device-security PIN constants
+        private const val AUTO_GENERATED_PIN_LENGTH = 6
+
+        // Exclusive upper bound passed to SecureRandom.nextInt() to produce a single decimal digit (0-9)
+        private const val DECIMAL_DIGIT_BOUND = 10
     }
 
     override fun getName(): String = NAME
@@ -2777,11 +2783,11 @@ class BcscCoreModule(
 
             val accountID = account.uuid
 
-            // Generate a cryptographically secure random 6-digit PIN
+            // Generate a cryptographically secure random PIN
             val secureRandom = java.security.SecureRandom()
             val pinDigits = StringBuilder()
-            for (i in 0 until 6) {
-                val digit = secureRandom.nextInt(10)
+            for (i in 0 until AUTO_GENERATED_PIN_LENGTH) {
+                val digit = secureRandom.nextInt(DECIMAL_DIGIT_BOUND)
                 pinDigits.append(digit)
             }
             val pin = pinDigits.toString()
@@ -2857,8 +2863,8 @@ class BcscCoreModule(
                                         // User had device security but no random PIN. Generate one now.
                                         val secureRandom = java.security.SecureRandom()
                                         val pinDigits = StringBuilder()
-                                        for (i in 0 until 6) {
-                                            pinDigits.append(secureRandom.nextInt(10))
+                                        for (i in 0 until AUTO_GENERATED_PIN_LENGTH) {
+                                            pinDigits.append(secureRandom.nextInt(DECIMAL_DIGIT_BOUND))
                                         }
                                         val pin = pinDigits.toString()
 
--- a/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
+++ b/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
@@ -47,6 +47,9 @@ import java.util.Date
 import java.util.Locale
 import java.util.TimeZone
 import java.util.UUID
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.RejectedExecutionException
 import javax.crypto.SecretKey
 import javax.crypto.Mac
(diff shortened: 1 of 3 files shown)


bryce-mcmath (approved):
Tested biometric authentication, failing biometric auth, using PIN, changing app security method, all works fine.

PR #4480, opened 2026-08-14
fix: prevent crash from missing cached video

### LINKED ISSUE #4439: App crashes on Android when reading a file that is no longer on the device

Unrelated to the InCallManager crash also reported under #4435 — different library, different root cause. Tracked separately.

## Issue

Android users are experiencing an app crash when the app inspects a file that is no longer on the device. The most likely everyday trigger is a photo or video the app captured earlier being cleared from the device's cache (Android reclaims cache space on its own, and users can clear it manually), after which the app still holds the old location and tries to read it.

Instead of handling the missing file and showing the user something useful, the app terminates. This is a poor experience during evidence capture and upload in BC Services Card identity verification, where the user may have to redo work.

There is a second, related consequence worth fixing at the same time: because the app currently dies rather than reporting the failure, we have no graceful path for "the file we expected is gone." That case needs a real user-facing outcome, not just an absence of crashing.

**Affected version:** 4.0.2 (Android). Reported twice, on different Android OS versions — the same defect in both cases.

## Exception

```
Exception java.lang.NullPointerException: Parameter specified as non-null is null: method com.facebook.react.bridge.PromiseImpl.reject, parameter code
  at com.facebook.react.bridge.PromiseImpl.reject (PromiseImpl.kt:2)
  at com.rnfs.RNFSManager.reject (RNFSManager.java:978)
  at com.rnfs.RNFSManager.stat (RNFSManager.java:645)
  at java.lang.reflect.Method.invoke
  at com.facebook.react.bridge.JavaMethodWrapper.invoke (JavaMethodWrapper.kt:217)
  at com.facebook.react.bridge.JavaModuleWrapper.invoke (JavaModuleWrapper.kt:134)
  at com.facebook.jni.NativeRunnable.run
  at android.os.Handler.handleCallback (Handler.java:959)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at com.facebook.react.bridge.queue.MessageQueueThreadHandler.dispatchMessage (MessageQueueThreadHandler.kt:21)
  at android.os.Looper.loopOnce (Looper.java:257)
  at android.os.Looper.loop (Looper.java:342)
  at com.facebook.react.bridge.queue.MessageQueueThreadImpl$Companion.startNewBackgroundThread$lambda$1 (MessageQueueThreadImpl.kt:175)
  at com.facebook.react.bridge.queue.MessageQueueThreadImpl$Companion.$r8$lambda$ldnZnqelhYFctGaUKkOKYj5rxo4 (MessageQueueThreadImpl.kt)
  at com.facebook.react.bridge.queue.MessageQueueThreadImpl$Companion$$ExternalSyntheticLambda0.run (D8$$SyntheticClass)
  at java.lang.Thread.run (Thread.java:1572)
```

(The second report is identical apart from Android OS line numbers.)

## Root cause

An incompatibility between the `react-native-fs` library (version `~2.20.0`) and React Native 0.81.5, which the 4.0 line uses.

When any `react-native-fs` file operation on Android fails with a generic error, the library reports it through a shared code path that passes no error code:

```java
// RNFSManager.java:978
promise.reject(null, ex.getMessage());
```

That was valid against the React Native version we used previously, where the receiving method accepted a missing code and substituted a default. In React Native 0.81 that method was converted to Kotlin and the two-argument overload now declares the code as non-null (`reject(code: String, message: String?)` in `PromiseImpl.kt`), so the Kotlin runtime null-check throws a `NullPointerException` instead.

Two consequences make this worse than a normal error path:

1. The `NullPointerException` is thrown from *inside* the library's own error handler, so the library cannot catch it. It propagates up and terminates the process.
2. No result is ever returned to the JavaScript side — no success, no failure. Existing `try/catch` blocks around these calls in our code cannot help, because there is nothing to catch. The app is already gone.

Notably, `stat` raises a plain generic `Exception("File does not exist")` for a missing file — not a `FileNotFoundException` — so it bypasses the library's ENOENT branch and takes the faulty null-code path every time. In practice, checking a file that isn't there is a reliable crash on Android rather than a recoverable error. The same is true of `unlink` (`delete`), which throws the identical generic exception, so this is a crash class shared by every Android operation in the library, not a `stat`-specific bug.

Of our three `stat` call sites, only two can actually produce these reports: the photo metadata read (`file-info.ts`) and the video upload preparation (`useEvidenceUploadModel.tsx`); the capture-date read (`capture-date.ts`) sits inside a try/catch that — while unable to stop the crash — marks a different usage pattern. The video upload site matches the reported scenario best.

**How the path goes stale (corrected):** the stale-file window is *within a session*, not across app restarts. `videoPath` is held in in-memory state only — the `SAVE_VIDEO` reducer does not persist it, and state hydration on launch explicitly clears `photoPath`/`videoPath`/`videoThumbnailPath` (`container-imp.ts`). The failure mode is Android reclaiming the recorder's cache file (or the recorder cleaning up) between capture and upload while the app still holds the path in memory. Photos are less exposed because they are copied to permanent storage after capture; video has no analogous permanent copy.

This is not a regression from any recent feature work. The `react-native-fs` line involved is years old and unchanged; the behaviour changed when we upgraded React Native to 0.81.5 (#3177). We confirmed the newest file-inspection code added on `main` is **not** present in the 4.0 line, so it is not a candidate for these reports.

Upstream `react-native-fs` is no longer maintained (last release 2.20.0, May 2022), so no fix will arrive from the original project.

## Proposed fix

Two parts, to be delivered together.

**1. Supply an error code on the Android side, via our existing library patch.**

We already carry a `yarn` patch for `react-native-fs` (currently iOS-only). Extend it to provide a non-null error code in that shared error path, matching what the actively maintained community fork does:

```java
promise.reject("RNFS", ex.getMessage());
```

Why this is low risk:

- Today's behaviour on that line is *always* a process crash. Nothing observable to the app changes, because no result is currently delivered at all.
- It is the shared error path for every `react-native-fs` Android operation, so it also protects file delete, move, copy, and similar operations — not just the one in these reports. (Our `removeFileSafely` helper does an `exists()` check before `unlink`, but a file can vanish between the two calls, so it is exposed to the same crash today.)

The patch should be regenerated with `yarn patch react-native-fs` rather than hand-edited, and the existing resolution key left unchanged.

**2. Handle the "file is gone" case properly in the video upload flow.**

The video path can go stale within a session once Android reclaims the recorder's cache file (see root cause). Part 1 converts that from a crash into a reported failure — but the surrounding code would then quietly swallow it, and the user would see an upload that simply doesn't work with no explanation.

Validate the path before use and, when the file is no longer present, guide the user back to re-capture rather than failing silently. Without this, fixing the crash makes the underlying problem invisible instead of solving it.

Considered and rejected:

- **Checking whether the file exists before inspecting it.** Insufficient. It leaves a gap between the check and the use, and the library's "exists" check resolves paths differently than the operation that crashes, so it does not cover all the failing cases. It would give false confidence.
- **Migrating to the maintained fork (`@dr.pogodin/react-native-fs`).** It does contain this fix (its shared error path always supplies a code, and its `stat` correctly reports ENOENT for missing files), and is worth considering as separate future work, but it is not a drop-in replacement: `@bifold/core` imports `react-native-fs` directly, so it would need a dependency alias, the Android package name differs (the two cannot coexist), and compatibility with our React Native version is unconfirmed. Too large for a crash fix. (For clarity: `@bifold/core` itself never calls `stat` — its usage is path constants and read/write/exists/mkdir — so it is not implicated in these reports.)
- **Upgrading React Native.** Does not actually fix this: React Native 0.82 and 0.83 still declare the code parameter non-null; only unreleased `main` has relaxed it back to nullable. Far too large a change for this regardless.

## Verification

- On Android, exercise the evidence photo and video paths after clearing the app cache, and confirm the app reports a handled failure and offers re-capture rather than terminating.
- Confirm the regenerated patch applies and reaches the compiled Android source.
- Monitor Play Store crash reporting for recurrence of `RNFSManager.reject` after release.

## Open question

The reports do not tell us which of the file-inspection call sites in 4.0.2 produced them (photo metadata read or video upload preparation — the two unguarded `stat` sites). Occurrence counts per call site would help confirm the cache-eviction theory, but the fix above is not blocked on knowing.

MacDeluca:
# Summary of Changes

This PR fixes a bug on Android that caused a fatal crash when the cached video was missing from the device (cache cleared).

# Testing Instructions

1. Record video verification (dont submit / confirm)
2. Android -> settings -> BCSC app -> storage -> clear cache
3. Submit video

# Acceptance Criteria

Error modal renders NOT a fatal crash.

# Related Issues
issue: https://github.com/orgs/bcgov/projects/108/views/1?pane=issue&itemId=227179045&issue=bcgov%7Cbc-wallet-mobile%7C4439

new: https://github.com/orgs/bcgov/projects/108/views/1?pane=issue&itemId=228953740&issue=bcgov%7Cbc-wallet-mobile%7C4478

### DIFF
--- a/.yarn/patches/react-native-fs-npm-2.20.0-a38fe24051.patch
+++ b/.yarn/patches/react-native-fs-npm-2.20.0-a38fe24051.patch
@@ -2,7 +2,7 @@ diff --git a/RNFSManager.m b/RNFSManager.m
 index 5ddd9417de101a0b18a298434dbc57f46b130f21..7bca62aadd52e2aa1575887965541c6972bcebb1 100755
 --- a/RNFSManager.m
 +++ b/RNFSManager.m
-@@ -294,8 +294,8 @@ RCT_EXPORT_METHOD(readFile:(NSString *)filepath
+@@ -294,8 +294,8 @@ + (BOOL)requiresMainQueueSetup
  }
  
  RCT_EXPORT_METHOD(read:(NSString *)filepath
@@ -13,3 +13,16 @@ index 5ddd9417de101a0b18a298434dbc57f46b130f21..7bca62aadd52e2aa1575887965541c69
                    resolver:(RCTPromiseResolveBlock)resolve
                    rejecter:(RCTPromiseRejectBlock)reject)
  {
+diff --git a/android/src/main/java/com/rnfs/RNFSManager.java b/android/src/main/java/com/rnfs/RNFSManager.java
+index 351ac066dd7936bba7f35627a0730db1dd272b2c..c350ddcc91ab280be7aa772bf1144edb7e746b85 100755
+--- a/android/src/main/java/com/rnfs/RNFSManager.java
++++ b/android/src/main/java/com/rnfs/RNFSManager.java
+@@ -975,7 +975,7 @@ public class RNFSManager extends ReactContextBaseJavaModule {
+       return;
+     }
+ 
+-    promise.reject(null, ex.getMessage());
++    promise.reject("RNFS", ex.getMessage());
+   }
+ 
+   private void rejectFileNotFound(Promise promise, String filepath) {
--- a/app/package.json
+++ b/app/package.json
@@ -111,7 +111,7 @@
     "react-native-confirmation-code-field": "~7.3.2",
     "react-native-device-info": "~8.7.1",
     "react-native-drop-shadow": "^1.0.3",
-    "react-native-fs": "~2.20.0",
+    "react-native-fs": "patch:react-native-fs@npm%3A2.20.0#~/.yarn/patches/react-native-fs-npm-2.20.0-a38fe24051.patch",
     "react-native-gesture-handler": "~2.28.0",
     "react-native-get-random-values": "~1.11.0",
     "react-native-gifted-chat": "*",
--- a/yarn.lock
+++ b/yarn.lock
@@ -7816,7 +7816,7 @@ __metadata:
     react-native-confirmation-code-field: "npm:~7.3.2"
     react-native-device-info: "npm:~8.7.1"
     react-native-drop-shadow: "npm:^1.0.3"
-    react-native-fs: "npm:~2.20.0"
+    react-native-fs: "patch:react-native-fs@npm%3A2.20.0#~/.yarn/patches/react-native-fs-npm-2.20.0-a38fe24051.patch"
     react-native-gesture-handler: "npm:~2.28.0"
     react-native-get-random-values: "npm:~1.11.0"
     react-native-gifted-chat: "npm:*"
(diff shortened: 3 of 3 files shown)


jleach (approved):
Confirmed on Android. I see the report a problem with "File not found" rather than an app crash.

PR #4481, opened 2026-08-14
fix: enter birthdate navigation

### LINKED ISSUE #4463: if a citizen sets up the app with an expired card, and the app closes before they enter a new serial number, they are not able to correct it and can only edit the birthdate
### Description of problem
the app re-opens in a state that the user cannot edit the card serial, but is being asked for their birthdate.

### Expected behavior
either allow the user to re-enter a good serial number, or roll the state back to before a card was entered.

https://github.com/user-attachments/assets/2547759d-07cf-4d05-9ecb-776860c512e5

### Environment: SIT

Build #: 7888

Android Device Model: Pixel 8

iOS Device Model:

### Workaround
Settings - Remove Account

MacDeluca:
# Summary of Changes

This PR refactors the VerifyStack goBack navigation and updates the logic for EnterBirthdate screen back navigation.

# Testing Instructions

Test entering and backing out of the verify stack at different points ie:
- Enter a serial -> navigate back
- Enter a serial -> enter a birthdate -> navigate back
- Enter a serial -> navigate home
- Enter a serial -> enter a birthdate -> navigate home
etc...

# Acceptance Criteria

# Related Issues

https://github.com/orgs/bcgov/projects/108/views/1?pane=issue&itemId=228314957&issue=bcgov%7Cbc-wallet-mobile%7C4463


al-rosenthal (changes requested):
I tested this on android and noticed a few things:
- once I entered a CSN and a birthdate there is no more back button, it changes into the hamburger menu, so I'm unable to go back
- on android, once I enter a CSN number, the hardware back button closes the app, I think it should emulate the header back button when possible

MacDeluca:
> I tested this on android and noticed a few things:
> 
> * once I entered a CSN and a birthdate there is no more back button, it changes into the hamburger menu, so I'm unable to go back
- This is intentional and the same behaviour as before (we have registered the device with IAS at this point)

> * on android, once I enter a CSN number, the hardware back button closes the app, I think it should emulate the header back button when possible
- I am unable to replicate this. Can you add some repro steps? The navigation is slightly different depending if the navigation history exists or not.

al-rosenthal (approved):
retested locally and it all works as expected

PR #4482, opened 2026-08-14
fix: handle already approved evidence response

### LINKED ISSUE #4475: 409 evidence error when registration is already approved
### Description

Occasional 409 error response on evidence upload endpoints in app, server logs say, "Registration Request already approved"

See BCSC Teams channel for more details

bryce-mcmath:
# Summary of Changes

Fixes a bug where users could get into a state where they were attempting to submit evidence after they had already been approved. The bug is worse in 4.0.2, we've partially addressed it already in main, this addresses the remaining broken paths.

# Testing Instructions

Fresh install, don't enable notifications, do an in-person verification and approve yourself but don't press "Complete", instead navigate back to the "Choose verification method" screen and do a send video. On main, you should get a 409 error, on this branch you should get fast forwarded to approval.

# Acceptance Criteria

No longer get 409

N/A

# Related Issues

#4475

### DIFF
--- a/app/src/bcsc-theme/utils/axios-error-utils.ts
+++ b/app/src/bcsc-theme/utils/axios-error-utils.ts
@@ -198,6 +198,8 @@ const getClientErrorDefinitionFromStatus = (status?: number): ErrorDefinition =>
       return ErrorRegistry.FORBIDDEN
     case 404:
       return ErrorRegistry.NOT_FOUND
+    case 409:
+      return ErrorRegistry.ALREADY_VERIFIED
     case 429:
       return ErrorRegistry.RETRY_LATER
     default:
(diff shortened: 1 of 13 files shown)


app/src/bcsc-theme/utils/axios-error-utils.ts
```diff
@@ -198,6 +198,8 @@ const getClientErrorDefinitionFromStatus = (status?: number): ErrorDefinition =>
       return ErrorRegistry.FORBIDDEN
     case 404:
       return ErrorRegistry.NOT_FOUND
+    case 409:
+      return ErrorRegistry.ALREADY_VERIFIED
```
MacDeluca: suggestion: This is probably too specific to go here, the other error codes are generic.

ie: not all 409 errors are because the user is already verified.

MacDeluca:
Tested locally on iOS and works as expected!

thought: I did notice that the log show two errors for /photos and /videos. Maybe we can eject earlier to prevent having to hit both endpoints?

bryce-mcmath:
@MacDeluca refactored to make 409's a generic `CONFLICT` and instead used an `clientErrorPolicies.ts` addition to catch our specific type of 409. The reason I didn't do an `useEvidenceService` approach is because I found we weren't actually using that hook in most cases where this error was occurring and it required a pretty big refactor. In the interest of keeping the PR smaller I just added the single specific error policy instead.

MacDeluca (approved):
Works great!

PR #4517, opened 2026-08-22
feat: add native support for creating a new signing key

### LINKED ISSUE #3876: Implement client-side key rotation in v4
## Summary

v4 has no client-side key rotation. The `key-rotation.workflow.ts` file at `app/src/bcsc-theme/api/workflow/key-rotation.workflow.ts` contains only empty stubs and is not called from anywhere. v3 implemented a 180-day rotation policy (introduced in the 3.8.x–3.10.x window, May 2024 — commit `d28c9ebe1`), so v4 currently inherits whatever key v3 left behind and rides it indefinitely.

## Current State

`app/src/bcsc-theme/api/workflow/key-rotation.workflow.ts`:

```ts
export async function checkKeyStatus() {
  // keys can be rotated for the following reasons:
  // - 90 days have passed
  // - key is not valid
}
export async function rotateKeys() {
  // get current keys
  // get new keys
  // end date old keys
  // save new keys to device
  // save new keys to server
  // check new keys work
}
```

Both functions have empty bodies. Neither is called from anywhere in the app — verified by grep.

## What v3 Did

In v3, `BcscKeyPairRepo.getNewBcscKeyPair()` generated a new keypair under an incremented alias (`rsa1` → `rsa2` → `rsa3`...). `AttemptClientRegistrationUpdateUseCase` PUT the new public key to IAS via `/register`. `cleanUpBcscKeyPairs()` removed the oldest alias once three or more existed. The 180-day age check triggered the whole flow.

In v4, the native `BcscKeyPairSource.getNewBcscKeyPair()` exists in `packages/bcsc-core/android/.../BcscKeyPairRepo.java` (it's referenced by PR #3874's reconciliation), but no JS caller invokes it for actual rotation.

## Why This Matters

- IAS may have server-side expiry policies that v4 won't survive long-term without client-side rotation.
- Long-tenure v4 users are stuck with their inherited v3 key indefinitely.
- The recovery story (server rejection, network drop during PUT, orphaned aliases) needs to be designed alongside the rotation itself — see related issues / PRs for the failure modes we've already observed.

## Acceptance

- v4 detects when the current signing key needs rotation (age threshold, or signature-rejection signal from the server)
- Generates a new keypair via the existing native `getNewBcscKeyPair`
- PUTs the new public key to IAS via `/register` using the stored `registration_access_token`
- Persists the new alias as the active signing key (last-successful-alias pattern)
- Falls back gracefully on failure: retains the previous key, surfaces a recoverable error path, schedules retry rather than locking the user out

## Open Questions

- **Cadence**: the placeholder comment says 90 days; v3 used 180. Which is the intended target?
- **Trigger source**: time-based check only, or also reactive to server-side `invalid_client` / "Signature did not validate" responses?
- **Cleanup**: should v4 mirror v3's `cleanUpBcscKeyPairs` (deletes oldest when 3+ exist) or use a different retention policy?

## Related

- #3870 — v3-to-v4 update error 119; the keypair-on-upgrade recovery story overlaps directly with the rotation recovery story
- PR #3865 (merged) — clientID/issuer fallback for v3→v4 migration
- PR #3874 (draft) — keypair reconciliation between v3 keystore aliases and v4 SharedPreferences metadata

jleach:
Closes #3876

## What changed

Adds a native `createNewKeyPair()` to `react-native-bcsc-core` (iOS Swift, Android Kotlin/Java, TurboModule spec, JS wrapper). It generates a new device signing key and returns its public RSA components in the same shape as the existing `getAllKeysWithPublicInfo`.

Nothing calls it yet — this PR adds a capability, not a behaviour. Rotation itself arrives in later PRs in the series.

It also hardens the generation path, which *is* shared with existing key handling:

- Generating a key implicitly makes it the active signing key on both platforms. Every failure path after that point now best-effort deletes the new key before rejecting, so a partial failure can't leave the device signing with a key the server has never seen. Six such paths across the two platforms.
- Key aliases embed a stable per-device identifier, so they're redacted in log lines *and* error messages (the latter surface in `AppError.technicalMessage` and analytics).
- iOS RSA-extraction failure now rejects with `E_KEY_EXPORT_FAILED` instead of `E_120_KEYCHAIN_KEY_DOESNT_EXIST_ERROR`, which mapped to `KEYCHAIN_KEY_NOT_FOUND` and misdescribed the failure.

**The `private` → `protected` widening of `loadAndroidKeyStore`, `generateKeyPair`, and `getKeyPair` in `BcscKeyPairRepo.java`.** Robolectric can't drive the real `AndroidKeyStore`, so the new cleanup tests need a subclass with fakes. It's a deliberate encapsulation tradeoff — push back if you'd rather have the tests than the visibility.

Secondary: the cleanup ordering in `BcscKeyPairRepo.getNewBcscKeyPair()` — metadata is now persisted only *after* generation succeeds, and cleanup unwinds both the keystore entry and the metadata row.

No app code touched, so no UI or runtime behaviour change to check.

## How to test

Covered by unit tests: 4 new Robolectric cases in `BcscKeyPairRepoCleanupTest` pin the cleanup paths (native suite 116/116). Each was falsification-checked — the cleanup block was deliberately broken, the test confirmed failing, then restored.

CI is green, including Native Tests (iOS) and Native Tests (Android).

For end-to-end rotation behaviour, the complete implementation is on `3876-key-rotation-full-spike`.

### DIFF
--- a/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
+++ b/packages/bcsc-core/android/src/main/java/com/bcsccore/BcscCoreModule.kt
@@ -395,6 +395,91 @@ class BcscCoreModule(
         }
     }
 
+    /**
+     * Generate a new signing keypair under a freshly incremented alias and return its public
+     * RSA components, for the JS-side key-rotation flow (issue #3876).
+     *
+     * IMPORTANT side effect: [BcscKeyPairSource.getNewBcscKeyPair] stamps the new alias'
+     * `createdAt` to now, so [BcscKeyPairSource.getCurrentBcscKeyPair] (newest-by-`createdAt`)
+     * picks it up as the active signing key immediately — there is no separate "activate" step.
+     * The JS caller is responsible for registering the new key with the server or rolling back
+     * via [deleteKey].
+     *
+     * Reuses [BcscKeyPairSource.convertBcscKeyPairToJWK] so the n/e encoding matches what
+     * registration actually sends (Nimbus canonical unsigned base64url).
+     */
+    @ReactMethod
+    fun createNewKeyPair(promise: Promise) {
+        // Captured once getNewBcscKeyPair() succeeds so a later failure in this method can
+        // best-effort delete it — see cleanUpUnregisteredKeyAfterGenerationFailure below.
+        var bcscKeyPair: BcscKeyPair? = null
+        try {
+            if (!keyPairSource.isAvailable()) {
+                promise.reject("E_KEYSTORE_UNAVAILABLE", "Android KeyStore is not available on this device")
+                return
+            }
+            bcscKeyPair = keyPairSource.getNewBcscKeyPair()
+            val jwk = keyPairSource.convertBcscKeyPairToJWK(bcscKeyPair)
+            if (jwk !is RSAKey) {
+                cleanUpUnregisteredKeyAfterGenerationFailure(bcscKeyPair.getKeyInfo().getAlias(), "not an RSA key")
+                promise.reject(
+                    "E_KEYSTORE_ERROR",
+                    "Newly generated key '${bcscKeyPair.getKeyInfo().getAlias()}' is not an RSA key",
+                )
+                return
+            }
+            val entry: WritableMap = Arguments.createMap()
+            entry.putString("id", bcscKeyPair.getKeyInfo().getAlias())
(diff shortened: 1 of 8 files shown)


al-rosenthal (approved):

al-rosenthal (commented):
a few robot comments but otherwise looks good

PR #4540, opened 2026-08-25
chore: extract ship-to-play composite action

### LINKED ISSUE #4522: Managed UAT/QA builds (release candidates)
### What do you want to be able to do?

Not all builds should necessarily create a release candidate. We need a way to "bundle" multiple commits in one release candidate

### Why does this matter?

We need to provide more stable builds for UAT/QA team, so that they are not surprised with an unexpected new build during their UAT/QA process

### Notes
- Publish to Firebase app distribution during the build phase, but assign to ring 2 during the release candidate approval
### Acceptance criteria

- Upon merge a commit in the "main" or any release/hotfix branch, the GHA workflow should pause and ask for approval if we want to publish this build as release candidate; or
- A manual workflow is somehow started/triggered that will re-use (publish to app stores) the build artifacts previously generated

### Wireframes, screenshots, or links

N/A

jleach:
Closes #4522

## What changed

Google Play was the only publish target still inlined in a build job — extracted it to a `ship-to-play` composite action, matching `ship-to-itunes`.

Two side-effects: publisher credentials now land in `RUNNER_TEMP` instead of the repo checkout, and the unused `VERSION_CODE` env var is gone (gpublish reads `GITHUB_RUN_NUMBER`).

Both steps are gated `if: github.ref_name == 'main'`, so PR CI only proves the YAML parses. The real check is the first main build.

The credentials handoff: `op read` writes to `$RUNNER_TEMP`, the action reads `${{ runner.temp }}`, and both steps carry the same `if:`.

## How to test

`actionlint` reports no new findings versus main and `shellcheck` is clean on the extracted script. After merge, the first main build should show the same Play internal-track release as today.

### DIFF
--- a/.github/workflows/actions/ship-to-play/action.yml
+++ b/.github/workflows/actions/ship-to-play/action.yml
@@ -1,0 +1,53 @@
+name: Ship to Google Play
+description: |
+  Upload an already-built, signed AAB to the Google Play internal track via
+  @bcgov/gpublish. This action only uploads — it does NOT promote the release
+  to any other track.
+
+inputs:
+  service_account_file:
+    description: |
+      Path to the Google Play publisher service-account JSON
+      (`op://bcsc-mobile-app-cd/google-play-publisher/service-account.json`).
+    required: true
+  package_name:
+    description: 'Android application ID to publish under (e.g. ca.bc.gov.id.servicescard).'
+    required: true
+  version_name:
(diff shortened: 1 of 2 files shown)


jleach:
Heads up on the Sonar C, they're not new to this PR (pre-existing) and both fixed in a add-on PR #4541. They just triggered because I moved the logic.

bryce-mcmath (approved):

PR #4549, opened 2026-08-26
feat: wire automatic signing-key rotation into system checks

### LINKED ISSUE #3876: Implement client-side key rotation in v4

jleach:
Closes #3876

## What changed

This is 5 of 5 in the signing-key rotation stack, and the first one with a real runtime behaviour
change — it turns automatic key rotation **on**. It plugs the rotation check into the app's normal
startup checks and adds the launch marker rotation depends on to know when it's safe to run.

Marked draft with `status/do-not-merge`: PRs 1–4 (#4517, #4524, #4535, #4547) must land first, and
this one needs extra review before it goes live.

The gating and ordering: rotation only runs for verified BC Services Card users who already have a
registration, and it's always sequenced strictly after the existing device-registration update so
the two automatic server calls can never collide.

## How to test

1. Have a fully setup app installed
2. Change both KEY_ROTATION_MAX_AGE_DAYS and KEY_ROTATION_RETRY_BACKOFF_DAYS to 0.
3. Run the app, it will rotate keys. Wait a few minutes, run it again, it will rotate again on startup (you can see in metro logs)
4. Repeat Step 3 a few times. The app will work as expected.


jleach:
Blocked on two defects found during manual testing on Android:

- #4595 — the app opens server responses with its newest key instead of the one the response names, so the launch that rotates shows an error and a Home screen with no account details.
- #4596 — generating the new key freezes the screen for up to a few seconds right after unlock.

Neither changes the rotation design in this PR. Both only become reachable once this PR merges, since this is what turns rotation on — nothing to fix retroactively in the earlier PRs in the stack.

https://claude.ai/code/session_01YKii4NH5hcgrphTecdmNFx

jleach:
Third blocker added: #4601 — the previous signing key is deleted as soon as rotation succeeds, so a response already in flight and locked with that key cannot be opened when it lands. #4595 does not cover this one, because reading the key label off the response cannot help once the key is gone.

Current blockers on this PR:
- #4595 — wrong key used to open server responses (fixed, draft PR #4600, verified on both platforms)
- #4596 — key generation freezes the Android screen during rotation
- #4601 — old key deleted before in-flight responses arrive

Context for #4601: comparing against the v3 native apps, both platforms kept between two and five keys after rotation, but neither verified the server had actually accepted the new key — the extra keys were covering that uncertainty. This app does verify it (`confirmModulusRegistered`), so the retention is not needed; only the in-flight window is.

https://claude.ai/code/session_01YKii4NH5hcgrphTecdmNFx

KjartanE (approved):

PR #4579, opened 2026-08-31
fix: filter services properly

### LINKED ISSUE #4570: The list of Services on 4.1.0 (PROD beta) has duplicate names
### What happened?

The feedback from the Beta PROD for 4.1.0 discussed some confusion around duplicate names.  I looked at it in PROD also, and found 4.1.0 is listing same name for some services. Additionally, from @gvodden:

>If you compare v3.15 to v4.0.2 in PROD, the 3.13 app has a "services not listed" "some services are only available to specific people..." where v4.0.2 (and the latest v4.1.0) is listing all the private services like Caregiver Registry, City of Chilliwack/Kelowna, diagnostic services secure upload tool, etc... there must be some filter criteria we are missing.

IAS related ticket: https://citz-cdt.atlassian.net/browse/IASP-19049

Names of Services should be unique.

### Steps to reproduce

1. Have a verified ID in PROD for 4.1.0
2. Look at the Services list
3. Note any duplicates

### Where did you hit this?

Production

### How often does it happen?

Every time

### Device and OS version

Pixel 6 Android 16

### Impact

Confusing the Citizens as to what the services are.

### Screenshots or log output

Image

### Build number

4.1.0 7828

bryce-mcmath:
Closes #4570

## What changed

Up until now, we weren't filtering our services list by an omitted service_listing_sort_order. The undocumented rule is that if that field is omitted, the service shouldn't be shown in the services list *unless* it's bookmarked.

Fairly simple change code-wise, manual testing should make the change obvious though.

## How to test

Verify yourself, check the services list, see that it is a lot cleaner (even in SIT environment). Message me on Teams for a more in-depth way to check the list is correct. Check bookmarking still works as expected.

### DIFF
--- a/app/src/bcsc-theme/features/home/components/SavedServices.tsx
+++ b/app/src/bcsc-theme/features/home/components/SavedServices.tsx
@@ -23,6 +23,7 @@ const SavedServices: React.FC = () => {
 
   const { serviceClients } = useFilterServiceClients({
     serviceClientIdsFilter: store.bcscSecure.savedServices,
+    savedServiceClientIds: store.bcscSecure.savedServices,
   })
 
   const styles = StyleSheet.create({
--- a/app/src/bcsc-theme/api/hooks/useMetadataApi.tsx
+++ b/app/src/bcsc-theme/api/hooks/useMetadataApi.tsx
@@ -25,6 +25,7 @@ export interface ClientMetadata {
   initiate_login_uri?: string
   client_description?: string
   policy_uri?: string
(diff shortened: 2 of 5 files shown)


app/src/bcsc-theme/features/home/components/SavedServices.tsx
bryce-mcmath, line 1: This component is no longer used, it's just for reference

al-rosenthal (approved):
tested and works as expected

PR #4584, opened 2026-09-01
fix: system check wait before home screen

### LINKED ISSUE #4539: Visual/Status Transition for Home notifications
### What happened?

State:
The user has been approved for their request in ID Check, but they have not clicked the "Finish up" button on Home page notification tile.
There are two instances where this happens:
1. Mobile powering up: instead of the "Your identity has been verified" screen, the screen "You're not verified"
2. Force-closing the app and opening it again

Item 1. is more pronounced - the delay is 1 - 2 seconds enough to be noticed.
Item 2. transition is much quicker - the delay is more of a flash on the screen.

There should not be a screen transition from one screen to the next.  As discussed with the team, the status check is not completed, and as a result, there is a transition from one screen to the actual one that needs to be displayed.

### Steps to reproduce

1. Have a request approved, but the citizen has not clicked the "Finish up" button.
2. Observe the transitions when powering up the mobile and starting the app, or when force-closing the app and opening it again.

### Where did you hit this?

SIT

### How often does it happen?

Every time

### Device and OS version

14 Plus (26.4.2)

### Impact

Visual flashing and in the case of a fresh boot, delayed transition between screens that can be avoided.

### Screenshots or log output

Timestamp 00:12 to 00:13:

https://github.com/user-attachments/assets/1b3e4ca9-6d3c-408e-acba-4d3f5053aa19

### Build number

4.1.0 8043

bryce-mcmath:
Closes #4539

## What changed
Wait for system checks to complete before showing home screen, up to a maximum wait time. This is to prevent a jumpy UI.

No regressions introduced, only waiting as much as necessary up until the max wait time.

## How to test
Restart the app to the home screen in a variety of states, especially when verified but needing renewal (`CustomNotificationId.AccountRenewalAvailable`). Inspect that there is no UI jumpiness unless for some reason loading takes longer than the max weight time.

### DIFF
--- a/app/src/bcsc-theme/hooks/useCreateSystemChecks.tsx
+++ b/app/src/bcsc-theme/hooks/useCreateSystemChecks.tsx
@@ -51,6 +51,11 @@ type UseGetSystemChecksReturn = Record<
      * Indicates if the system checks for the scope are ready to be run
      */
     isReady: boolean
+    /**
+     * Indicates if the scope will ever run in the current app state. False means callers waiting on
+     * the scope (see the loading gate in MainStack) should stop waiting — `isReady` will never flip.
+     */
+    isApplicable: boolean
   }
 >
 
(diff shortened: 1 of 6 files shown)


app/src/bcsc-theme/hooks/useCreateSystemChecks.tsx
```diff
@@ -51,6 +51,11 @@ type UseGetSystemChecksReturn = Record<
      * Indicates if the system checks for the scope are ready to be run
      */
     isReady: boolean
+    /**
+     * Indicates if the scope will ever run in the current app state. False means callers waiting on
+     * the scope (see the loading gate in MainStack) should stop waiting — `isReady` will never flip.
+     */
+    isApplicable: boolean
```
jleach: I might have called that something more easy to understand the context. Maybe `skipOnCurrentAppState`. Not a blocker, just a FYI based on how I read the code comment.

al-rosenthal (approved):

app/src/bcsc-theme/hooks/useSystemChecks.tsx
```diff
@@ -35,9 +35,10 @@ export enum SystemCheckScope {
  *   - VERIFY: Checks that run within the verification flow (VerifyStack) for an unverified, authenticated user ie: expired verification session
  *
  * @param {SystemCheckScope} scope - The scope of the system checks to run
- * @returns {*} {void}
+ * @returns Whether the scope has settled, i.e. its checks have finished running (or will never run).
+ *   Callers that paint state the checks can change (see MainStack's loading gate) wait on this.
```
jleach: This line is a little hard to read.

PR #4597, opened 2026-09-01
feat: custom change log workflow

### LINKED ISSUE #4464: Changelog generation
### Description

Either with changesets or some other tool, we need a autogenerated changelog to more easily track what has changed between two different publishes. The changelog should be readable to non-technical internal folks

al-rosenthal:
Closes #4464 

## What changed
A change log system to generate user facing change logs. A new workflow was added to collect any change sets and adds them to the `CHANGELOG.md` tagged with the date

- `.changes/` directory added to store all change sets
- added `CHANGELOG.md` to track all changes
- added `finalize-changelog.yml` workflow added to collect any change sets in `.changes/` and adds them in a single section in `CHANGELOG.md` tagged with the date
- added `index.mjs` script to grab any existing change sets and add them to the `CHANGELOG.md`
- updated `quality.yml` to keep changelog PRs from running tests for the change log PR
- added `yarn changeset` script to generate change sets
- updated label `status/skip-changelog` to ignore change set check on PRs (for documentation or minor changes)
- added `yarn changelog:test` 

## What's left
- Changelog Entry: this needs to be added to the ruleset so a change set will be required for a PR
- Changelog Script Tests: this needs to be added to the rulesets for quality checks

- finalize-changelog.yml: this is the workflow that takes any change sets and adds them to the change log file
- `index.mjs`: this is the script that the workflow runs to do the work of collecting change sets 
- changes in `qaulity.yml`: this was updated to skip the required checks at a job level for the change log workflow PR

## How to test
- run `yarn changelog:test`, all tests should pass
- pull locally and try out `yarn changeset`


scripts/changelog/create.mjs
```diff
@@ -0,0 +1,182 @@
...
+  'lantern',
+  'compass',
+  'anchor',
+  'maple',
+  'cedar',
+  'willow',
+]
+
+function randomChoice(list) {
+  return list[Math.floor(Math.random() * list.length)]
```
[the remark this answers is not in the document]
al-rosenthal: this random number is grabbing strings from an array, not a real threat

.github/workflows/quality.yml
```diff
@@ -3,15 +3,85 @@ name: Code Quality
...
+            echo "should_skip=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "should_skip=false" >> "$GITHUB_OUTPUT"
+          fi
+
   typecheck-test:
     runs-on: macos-26
     name: Build, Typecheck, Test & Coverage
+    needs: detect-relevant-changes
+    if: needs.detect-relevant-changes.outputs.should_skip != 'true'
```
jleach, line 84: I think this will cause a failed / cancelled `detect-relevant-changes` to report as skipped here, and our branch protection rules will treat that as "green" (satisfied). So a flake in the new job could let a PR through with none of the four required checks actually running.

The bit that makes it happen is the implicit `success()` — a job `if` gets that prepended unless the expression already contains a status check function, so this currently evaluates as `success() && ...`, and a failed dependency means skipped rather than failed.

I think `!cancelled()` would address it — it suppresses the implicit `success()`, and a job that died never wrote its output, so `should_skip` is empty and the job runs:

```yaml
if: ${{ !cancelled() && needs.detect-relevant-changes.outputs.should_skip != 'true' }}
```

(`always()` does the same job, but the docs steer away from it because it also returns true on cancellation — we'd end up running the four macOS jobs on a superseded push.)

.github/workflows/quality.yml
```diff
@@ -88,6 +158,8 @@ jobs:
   lint-format:
     runs-on: macos-26
     name: Lint & Format
+    needs: detect-relevant-changes
+    if: needs.detect-relevant-changes.outputs.should_skip != 'true'
```
jleach, line 162: same as above re: `!cancelled()`

.github/workflows/quality.yml
```diff
@@ -121,6 +193,8 @@ jobs:
   test-ios:
     runs-on: macos-26
     name: Native Tests (iOS)
+    needs: detect-relevant-changes
+    if: needs.detect-relevant-changes.outputs.should_skip != 'true'
```
jleach, line 197: same as above re: `!cancelled()`

.github/workflows/quality.yml
```diff
@@ -151,6 +225,8 @@ jobs:
   test-android:
     runs-on: macos-26
     name: Native Tests (Android)
+    needs: detect-relevant-changes
+    if: needs.detect-relevant-changes.outputs.should_skip != 'true'
```
jleach, line 229: same as above re: `!cancelled()`

.github/workflows/quality.yml
```diff
@@ -173,3 +249,49 @@ jobs:
...
+            echo "::error::No .changes/*.md file added in this PR. Add one describing your change for a non-technical reader (see CONTRIBUTING.md), or apply the 'status/skip-changelog' label if this PR doesn't need one."
+            exit 1
+          fi
+          echo "Found changelog entry: $added"
+
+  changelog-tests:
+    name: Changelog Script Tests
+    runs-on: ubuntu-22.04
+    needs: detect-relevant-changes
+    if: needs.detect-relevant-changes.outputs.should_skip != 'true'
```
jleach, line 285: same as above re: `!cancelled()` — not a required check so lower stakes, but same outcome.

.github/workflows/finalize-changelog.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      # --allow-empty makes "nothing pending" a clean no-op instead of a
+      # failure: this workflow runs on a schedule/whim, not in response to
+      # a specific change, so an empty .changes/ is the expected common case.
+      - name: Assemble changelog
+        run: node scripts/changelog/index.mjs assemble --allow-empty
+
+      - name: Open the PR
+        env:
+          GH_TOKEN: ${{ github.token }}
```
jleach, line 47: I don't think this PR will get its checks run automatically, because of how GH treats things done with `github.token`. Events caused by `GITHUB_TOKEN` don't kick off new runs — the `push` and the `--add-label` on line 73 won't trigger anything at all. `pull_request: opened` is a partial exception: it does create runs, but in an approval-required state, so the PR sits with an "Approve workflows to run" banner until someone with write access clicks it. Until then the required checks are pending and branch protection blocks the merge.

So not permanently stuck, just not unattended — which reads a bit against the "nothing to review by hand" line in the generated PR body. If we want it to just run, `actions/create-github-app-token` (or a PAT) instead of `github.token` would do it. Otherwise maybe worth a note in the PR body that someone has to approve the workflows.

Docs: https://docs.github.com/en/actions/concepts/security/github_token

I think :)

EDIT: PATs are tied to a specific developer so note great, but not terrable either. Allwork done under a PAT appears as if it was done by that developer.

jleach (changes requested):

al-rosenthal:
Closing this pull request until requirements become more clear

PR #4606, opened 2026-09-03
fix: stop ledger lookup messages from flooding development logs

### LINKED ISSUE #4599: Ledger lookup messages flood the development logs
## Summary

Routine ledger lookups print three messages every time the app resolves a DID, once per configured ledger, and each one arrives with a full stack trace attached. They are the most verbose class of message the wallet framework produces, and no current setting turns them off without also turning off everything else useful.

## Why it matters

Developers lose real warnings in the scroll, and the stack traces make routine lookups look like failures. The messages also reach our remote log service during support sessions, and building a stack on every message costs a little time in shipped builds.

## Proposed fix

Three small changes in the shared logging package (`@bifold/remote-logs`): treat these as the lowest-detail messages they actually are so the existing log level setting hides them, print them without a stack trace, and keep support sessions at full detail so nothing is lost from remote logs. Error and crash reporting are untouched.

## Acceptance

- Ledger lookup messages no longer appear at the default development log level
- No message prints a stack trace unless it is an actual error
- Turning on remote logging still captures the same detail it does today
- Error and problem reporting are unchanged

## Notes

The fix belongs upstream in Bifold; a temporary patch in this repo can carry it until that lands. Related follow-up: `LOG_LEVEL=trace` and `LOG_LEVEL=test` are documented in the README but currently behave the same as `debug` (`app/src/utils/logger.ts`).

jleach:
Closes #4599

## What changed

Ledger lookups printed three stack-trace-laden messages per ledger on every DID resolution. A temporary yarn patch on `@bifold/remote-logs` gates `trace`/`test` on their own log levels, drops `console.trace`, and keeps remote logging at full detail. Also fixes `LOG_LEVEL=trace|test`, which the README documents but which both behaved as `debug`.

The one to check is remote logging: enabling it must still capture the same detail it does today, since that is what support sessions rely on. There is a test that asserts the trace line actually reaches the log service, not just the console.

The patch belongs upstream in Bifold. It is a bridge until that lands, and it removes itself: bumping `@bifold/remote-logs` past 3.0.21 fails the install until someone deletes the patch. The upstream change and the two Bifold tests that will need updating are written up in the issue.

## How to test

Covered by unit tests, including a new spec that exercises the real patched build rather than the usual mock.

On device (BC Wallet), with a credential offer or connection invitation to scan:

1. Set `LOG_LEVEL=trace` in `app/.env`. Scanning produces the ledger lookup lines,
   now tagged `[TRACE]` and with no stack trace attached:

   console.js:661 [TRACE] Get public did 'Ttmj1pEotg8FbKZZD81S7i' from ledger 'candy'
   console.js:661 [TRACE] Submitting get did request for did 'Ttmj1pEotg8FbKZZD81S7i' to ledger 'candy'

2. Set `LOG_LEVEL=debug` (the default). The same scan produces none of those lines,
while other debug messages still appear.

### DIFF
--- a/.yarn/patches/@bifold-remote-logs-npm-3.0.21-4ae200989a.patch
+++ b/.yarn/patches/@bifold-remote-logs-npm-3.0.21-4ae200989a.patch
@@ -1,0 +1,68 @@
+diff --git a/build/logger.js b/build/logger.js
+index 7e9c252ba4d35eba614d8ac5362dba79e514a510..0c58b0e8665bc20703fed6f57af545431b54da94 100644
+--- a/build/logger.js
++++ b/build/logger.js
+@@ -27,14 +27,14 @@ class RemoteLogger extends core_1.AbstractBifoldLogger {
+         // Standardized logging methods with consistent overloads
+         this.test = (message, dataOrError, error) => {
+             var _a, _b;
+-            if (!this.isEnabled(core_2.LogLevel.debug))
++            if (!this.isEnabled(core_2.LogLevel.test))
+                 return;
+             const { data, actualError } = this.parseLogArguments(dataOrError, error);
+             (_b = (_a = this._log) === null || _a === void 0 ? void 0 : _a.test) === null || _b === void 0 ? void 0 : _b.call(_a, { message, data, error: actualError });
+         };
+         this.trace = (message, dataOrError, error) => {
+             var _a, _b;
+-            if (!this.isEnabled(core_2.LogLevel.debug))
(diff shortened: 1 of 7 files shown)


jleach:
The patch in this PR is temporary. I'll create a PR in bifold and remove when we're ready for a package update.

bryce-mcmath (approved):
I think this is good, but I wonder if it's better for us to just upstream the bifold change now and go straight to the proper package update?

PR #4609, opened 2026-09-03
feat: publish workflow tags build

al-rosenthal:
## What changed

- job added to publish workflow to create a git tag for the build chosen to publish
- git tag is the following format: bcsc-v4.1.0-1234
  - v4.1.0: is the version grabbed from the variant.env during the run
  - 1234: is the build number selected to publish 

- the changes made in `publish.yml`

## How to test
CI/CD only change, nothing to test for this PR.


jleach (approved):
Docs note on tagging covers what the tag means and the retry behaviour. Approving.

PR #4614, opened 2026-09-04
feat(deps): upgrade vision camera to v5

MacDeluca:
Closes #


app/package.json
```diff
@@ -118,6 +118,8 @@
     "react-native-keychain": "~10.0.0",
     "react-native-localize": "~2.2.6",
     "react-native-logs": "~5.1.0",
+    "react-native-nitro-image": "^0.15.2",
```
bryce-mcmath: 🥰

app/app.json
```diff
@@ -1,4 +1,13 @@
 {
   "name": "BCWallet",
-  "displayName": "BC Wallet"
+  "displayName": "BC Wallet",
```
bryce-mcmath, line 3: Since the default variant is BCSC Dev, I think the display name might be better as "BC Services Card", and then we can patch it to "BC Wallet" when we do that variant script.

PR #4621, opened 2026-09-08
chore: shared component and navigator test-ids

### LINKED ISSUE #4593: testID single source of truth (app + e2e import one key registry)
### What do you want to be able to do?

As a developer changing either side of the app/e2e boundary:

- Rename or remove a screen element's testID with **one registry edit** that both the app and the e2e suite compile against — no hand-maintained mirror.
- Add a new testID by registering its key first (lint enforces it), so the e2e suite discovers ids by importing the same file the app does.
- Have a PR that strands an e2e reference (a key renamed or deleted while a descriptor still uses it) fail **at PR time** as a `tsc` error, instead of minutes into a Sauce run behind a build cycle.
- Compose dynamic ids (`EvidenceTypeListItem-…`, `ServiceButton-…`) from shared stem constants instead of duplicated prefixes.

No behavior change: the emitted id strings stay byte-identical (`com.ariesbifold:id/<key>`).

### Why does this matter?

- Today the app writes inline `testIdWithKey('Literal')` strings (324 non-test call sites, 248 distinct keys, ~42 dynamic) and e2e mirrors them by hand in `e2e/src/test-ids/registry.ts` (~277 keys). An app-side rename surfaces as a **runtime** e2e failure on Sauce — minutes late, behind a build cycle.
- Keeping the mirror honest has already cost two rounds of manual testID stabilization (one during the e2e suite rework, another after a full screen audit); the drift is recurring, not hypothetical.
- Nothing structural prevents recurrence: e2e's `tsc` runs nowhere in PR CI (`quality.yml` typechecks app + bcsc-core only), so nothing catches a stranded reference before merge.
- The groundwork is done and proven: the registry was built dependency-free precisely so an RN bundle can import it, every e2e descriptor module already consumes it, and its header documents this migration as "step 3" — deliberately deferred as a follow-up when the registry pilot landed.

### Acceptance criteria

- Every static app-owned `testIdWithKey` call site references a registry key; eslint fails on a new inline string literal.
- App and e2e compile against the same file; renaming or deleting a key an e2e descriptor uses fails PR CI (`tsc`), not a Sauce run.
- Emitted ids unchanged: the app jest snapshot diff is **empty** (any snapshot change = a mistyped key), and one Sauce smoke is green on a migrated build.
- A guard test pins `TESTID_PREFIX` to bifold's `testIdPrefix`.
- `tsc` / eslint / jest clean on both sides.

### Wireframes, screenshots, or links

- `e2e/src/test-ids/registry.ts` — the pilot registry; its header documents this exact migration ("step 3") and the guard test.
- `e2e/src/screens/core/appId.ts` — how e2e consumes keys today (the `bcsc()` prefix helper).
- `chore/e2e_testid` branch — the most recent manual stabilization pass, the kind of catch-up work this makes unnecessary.

KjartanE:
Closes #4593

## What changed

Migrates the shared BCSC components and navigators — 58 call sites across 21 files, 25 of them dynamic. Refactor-only.

**Stacked on `chore/e2e_testID_main`.** Review only this branch's diff.

- New top-level `shared` registry area for ids that components **compose** rather than own: the `MaskedCamera` trio, `AppBanner`'s four `<part>-<type>` prefixes, the eight `InputWithValidation`/`DropdownWithValidation` field suffixes, `PINInput`'s visibility suffix, and the `CardButton-`/`SectionButton-` title prefixes.
- New top-level `reportProblem` area for the help-menu modal — distinct from `ErrorInfoCard`'s own report controls, which are different ids and land in the next PR.
- The eslint glob widens to `['src/test-ids/**', 'src/bcsc-theme/**']`, so the whole BCSC tree is enforced.
- Six sites intentionally unchanged: two passthroughs (`TileButton`, `PINInput`) and four label-derived ids.

- **Blast radius.** These components feed nearly every screen, so this is the PR most likely to move an emitted id. It doesn't: 160/160 snapshots pass with no `.snap` written.
- The `shared` area deliberately duplicates ids that also appear spelled out under their screen (`verify.manualSerial.serialInput` is `'serial'` + `field.input`). The doc comment says why — specs select the composed form, call sites need the pieces.
- `ActionScreenLayout` and `BulletedInstructionsScreen` still derive ids from translated labels (`testIdWithKey(primaryActionText)`), so the emitted id differs by locale. The rule can't see them — they're identifiers, not literals. Left alone here on purpose; fixing them means a `testIDKey` prop across 11 callers, which isn't refactor-only.

## How to test

```
cd app && yarn typecheck && yarn lint && yarn test
cd e2e && yarn typecheck
```

Snapshots are the contract here more than anywhere: 160/160 with no `.snap` written.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/app/eslint.config.mjs
+++ b/app/eslint.config.mjs
@@ -101,7 +101,7 @@ export default [
   // so unmigrated areas stay green while migrated ones cannot regress. Tests are exempt on purpose:
   // asserting the literal id string is what proves the emitted ids never moved.
   {
-    files: ['src/test-ids/**', 'src/bcsc-theme/features/**', 'src/bcsc-theme/contexts/**'],
+    files: ['src/test-ids/**', 'src/bcsc-theme/**'],
     ignores: ['**/*.test.{ts,tsx}'],
     rules: {
       'no-restricted-syntax': [
--- a/app/src/bcsc-theme/components/AppBanner.tsx
+++ b/app/src/bcsc-theme/components/AppBanner.tsx
@@ -1,3 +1,4 @@
+import { TestIds } from '@/test-ids/registry'
 import { ThemedText, testIdWithKey, useTheme } from '@bifold/core'
 import React, { useState } from 'react'
 import { StyleSheet, TouchableOpacity, View } from 'react-native'
@@ -120,7 +121,7 @@ export const AppBannerSection: React.FC<AppBannerSectionProps> = ({
   return (
     <TouchableOpacity
       style={[{ ...styles.container, backgroundColor: bannerColor(type) }]}
-      testID={testIdWithKey(`button-${type}`)}
+      testID={testIdWithKey(`${TestIds.shared.appBanner.buttonPrefix}${type}`)}
       accessibilityLabel={a11yLabel(title || description || '')}
       accessibilityRole="button"
       onPress={() => {
@@ -135,7 +136,7 @@ export const AppBannerSection: React.FC<AppBannerSectionProps> = ({
         size={24}
         color={type === 'warning' ? ColorPalette.brand.secondaryBackground : ColorPalette.grayscale.white}
         style={styles.icon}
-        testID={testIdWithKey(`icon-${type}`)}
+        testID={testIdWithKey(`${TestIds.shared.appBanner.iconPrefix}${type}`)}
       />
       <View style={styles.textContainer}>
         {title ? (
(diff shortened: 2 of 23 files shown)


jleach (dismissed):
Formatting tidied on MainStack and PINInput (pushed directly). Rest looks good.

jleach (approved):
Re-approving after reverting the MainStack alias; PINInput tidy-up stands.

PR #4628, opened 2026-09-09
fix: banner colour palette and severity order

### LINKED ISSUE #4200: Device Management banner - Contrast does not conform WCAG standards

jleach:
Closes #4200

## What changed

Warning banners showed white text and a white icon on yellow, failing WCAG contrast for every user. They now use the off-black from the BCSC style guide. Where several banners appear at once, they display most-serious-first: error, warning, informative, success. All other banner colours and the layout are unchanged.

Screenshot_20260910-182822

- The four banner colour pairs now live in one named `BannerColors` constant in the shared theme file, so light and dark can't drift apart on them again. The four background colours moved across unchanged; the only value that actually changes is the warning foreground (`#2D2D2D`).
- The Live Call screen's two warning banners ("video will resume", "volume low") pick up the same change — their text and icon go from white to off-black.

## How to test

Comment below on some test code you can use. I recommend having an account already setup so you land on the home screen ready for the demo code to kick in.

The severity ordering came from a comment in Figma. May not have been in the original ticket. 

Test code adds banners out of order so you can see sorting in action.


jleach:
### Quick way to see this on the Home screen

Banners only show up when something real triggers them, so here's a temporary snippet that pushes all four onto the Home screen at once. **It's throwaway — delete it after.**

In `app/src/bcsc-theme/features/home/Home.tsx`, add this above `type HomeProps = ...`:

```tsx
const useBannerOrderDemo = () => {
  const [, dispatch] = useStore<BCState>()

  useEffect(() => {
    // Deliberately the reverse of the order they should appear in.
    const demo: BCSCBannerMessage[] = [
      { id: BCSCBanner.ACCOUNT_EXPIRED, title: '4 - Success (expect LAST, green)', type: 'success' },
      { id: BCSCBanner.IAS_SERVER_NOTIFICATION, title: '3 - Info (expect THIRD, blue)', type: 'info' },
      { id: BCSCBanner.DEVICE_LIMIT_EXCEEDED, title: '2 - Warning (expect SECOND, yellow + OFF-BLACK text)', type: 'warning' },
      { id: BCSCBanner.APP_UPDATE_AVAILABLE, title: '1 - Error (expect FIRST, red)', type: 'error' },
    ]

    const timers = demo.map((banner, i) =>
      setTimeout(() => dispatch({ type: BCDispatchAction.ADD_BANNER_MESSAGE, payload: [banner] }), 2000 + i * 1500)
    )

    return () => {
      timers.forEach(clearTimeout)
      demo.forEach((banner) => dispatch({ type: BCDispatchAction.REMOVE_BANNER_MESSAGE, payload: [banner.id] }))
    }
  }, [dispatch])
}
```

Call it inside `Home`, next to the other hooks:

```tsx
useBannerOrderDemo() // TEMPORARY - delete before committing
```

And extend the imports already at the top of the file:

```tsx
import { BCSCBanner, BCSCBannerMessage } from '@/bcsc-theme/components/AppBanner'
import { BCDispatchAction, BCState } from '@/store'
import React, { useCallback, useEffect } from 'react'
```

Run with `BUILD_TARGET=bcsc`.

**What you should see:** the banners arrive worst-last, one every 1.5s, but the list always reads top to bottom **red → yellow → blue → green**. The titles are numbered, so a wrong order is obvious.

The main thing to check is the **yellow** banner: its text and warning icon should be near-black and easy to read. Before this change they were white on yellow, which is what the ticket reported.

Two gotchas:

- Banners are saved to storage, so if you force-quit the app mid-test they'll still be there next launch. Tap to dismiss, or navigate away from Home and let the cleanup run.
- Tapping the yellow banner opens the device-management modal rather than dismissing — it reuses the real device-limit banner id.

bryce-mcmath (approved):

PR #4644, opened 2026-09-11
fix: focus first invalid field on non-BCSC form submit

### LINKED ISSUE #4611: Usability - When non-BCSC form fails field validation the users screen should focus on the first instance of it
### What happened?

We had a ticket originally for it, #3220, but it looks like something broke what was fixed then.
Creating this ticket to track the new fix.

When form fails validation on a field, and the user clicks the Continue button, the user's screen is scrolled to the first instance of the issue.

### Steps to reproduce

1. Create a non-BCSC flow
2. For any of the Evidence Documents that have long form, leave the first field blank and fill the rest.
3. Click the Continue button at the bottom of the form.
4. Observe the screen not scrolling to the issue.

### Where did you hit this?

QA

### How often does it happen?

Every time

### Device and OS version

14 Plus (26.4.2)

### Impact

_No response_

### Screenshots or log output

_No response_

### Build number

4.1.0 8333

jleach:
Closes #4611

## What changed

Tapping Continue with a missing or invalid field now jumps the screen to the first field needing attention and puts the cursor in it, with the error in view. Covers both non-BCSC forms: the evidence ID form and Address Entry. Province is a picker, so the screen scrolls to it without opening it — a modal would hide the error.

The jump is instant, not animated, and runs before the field takes focus; the keyboard listener only acts if its field is still focused. Reversing either one breaks the behaviour, and tests fail if they're changed. Shared logic now lives in `useFocusFirstInvalidField` — the evidence form's test file is untouched by the extraction, which is the proof nothing shifted.

## How to test

Covered by unit tests (`useFocusFirstInvalidField.test.tsx`, `EvidenceIDCollectionScreen.test.tsx`, `ResidentialAddressScreen.test.tsx`, `DateInput.test.tsx`).

Manual, in a non-BCSC flow (bcsc target):

**Evidence ID form (one of the longer ones)**
1. Leave document number blank, fill the rest, tap Continue. The screen jumps to document number, the cursor lands in it, and the keyboard opens.
2. Dismiss the keyboard, fill everything except birth date, tap Continue. Birth date and its error should be fully visible above the keyboard, not cut off.
3. Enter middle names over 30 characters plus an invalid birth date, tap Continue. Middle names should be the target, not birth date.
4. With a field already focused and showing an error, tap Continue, then tap a different field. The screen should stay on the field you tapped.

**Address Entry**
5. Empty form, tap Continue. Jumps to Street Address, keyboard opens, error visible.
6. Street, city and postal code filled, province empty. Jumps to Province, keyboard unchanged, picker stays closed.
7. Clear postal code with the keyboard open on City, tap Continue. Jumps to Postal Code with its error above the keyboard.
8. Pick a province while postal code is empty. The cursor moves to Postal Code (unchanged from #4626).

### DIFF
--- a/app/jestSetup.js
+++ b/app/jestSetup.js
@@ -117,6 +117,9 @@ jest.mock('react-native-keyboard-controller', () => {
   const { ScrollView } = jest.requireActual('react-native')
   return {
     KeyboardAwareScrollView: ScrollView,
+    KeyboardEvents: {
+      addListener: () => ({ remove: () => {} }),
+    },
   }
 })
 
--- a/app/src/bcsc-theme/components/DateInput.test.tsx
+++ b/app/src/bcsc-theme/components/DateInput.test.tsx
@@ -1,7 +1,8 @@
 import { testIdWithKey } from '@bifold/core'
 import { BasicAppContext } from '@mocks/helpers/app'
 import { fireEvent, render, screen } from '@testing-library/react-native'
-import React, { useState } from 'react'
+import React, { createRef, useState } from 'react'
+import { TextInput } from 'react-native'
 import DateInput from './DateInput'
 
 const defaultProps = {
@@ -72,6 +73,12 @@ describe('DateInput Component', () => {
     test('does not show the error element when no error provided', () => {
       renderDefault()
       expect(screen.queryByTestId(testIdWithKey('birthdate-error'))).toBeNull()
+    })
+
+    test('exposes the underlying TextInput through ref', () => {
+      const ref = createRef<TextInput>()
+      renderDefault({ ref })
+      expect(ref.current?.props.testID).toBe(testIdWithKey('birthdate-input'))
     })
   })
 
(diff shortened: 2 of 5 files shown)


MacDeluca (changes requested):
suggestion: The ticket is referring to all forms during the non-bcsc flow. We also need this logic in the `Residential Address` form.

thought: If there are enough usecases we could probably wrap the meat of the scrollTo logic in `EvidenceIDCollectionScreen` into a shareable util or hook.

jleach:
Agreed on both. The Residential Address had no scroll-to-error logic at all, so this is net new there.

MacDeluca:
Postal code input is still not auto-focusing after the province selection. (Tested on iOS)

Error focusing and auto-scrolling is working great!

jleach:
Fixed, ios jumps to postal when a prov is selected

MacDeluca (approved):
Works great!

PR #4646, opened 2026-09-12
fix: banner colour palette and severity order

### LINKED ISSUE #4200: Device Management banner - Contrast does not conform WCAG standards

jleach:
Closes #4200

Cherry-pick of #4628 onto the 4.1.1 release line. #4628 stays open against `main`.

## What changed

Warning banners showed white text and a white icon on yellow, failing WCAG contrast for every user. They now use the off-black from the BCSC style guide. Where several banners appear at once, they display most-serious-first: error, warning, informative, success. All other banner colours and the layout are unchanged.

- Same three commits as #4628, with one conflict resolution: the test-id registry (`TestIds.shared.appBanner.*`, added in #4621) does not exist on the 4.1.0 line, so the four `testID` values keep their original string literals. Those literals are the same strings the registry prefixes resolve to, and the PR's own tests assert them, so behaviour is unchanged.
- The four banner colour pairs now live in one named `BannerColors` constant in the shared theme file, so light and dark can't drift apart on them again. The only value that actually changes is the warning foreground (`#2D2D2D`).
- The Live Call screen's two warning banners pick up the same change — text and icon go from white to off-black.

## How to test

As per #4628. Have an account already set up so you land on the home screen.

Note: the `Validate APP_VERSION matches release branch` check fails until the 4.1.1 version bump lands (see the keychain PR on this branch).


bryce-mcmath (dismissed):

jleach:
Problem with the banner was that our test IDs are different now between main and when v4.1.0 build was created.

bryce-mcmath (approved):

PR #4669, opened 2026-09-15
fix: handle torch availability

### LINKED ISSUE #4664: Error 2000 — Camera hardware unavailable or failed to initialize
### What happened?

Reported in release 4.1.0-8333

No camera errors 

### Steps to reproduce

Unknown at this point

### Where did you hit this?

Production

### How often does it happen?

Occasionally

### Device and OS version

iOS

### Impact

Some users can't use the camera features

### Screenshots or log output

_No response_

### Build number

4.1.0-8333

KjartanE:
Closes #4664

## What changed

Fixes the 4.1.0 `camera.add_card_camera_broken.2000` incidents on the Scan Card screen: an iPad torch tap that broke the camera, and a transient iOS AVFoundation runtime error that was treated as fatal.

- `CodeScanningCamera`: torch is only sent to the native camera when `device.hasTorch`; built-in torch button hidden otherwise; new `onTorchAvailabilityChange` prop.
- `ScanSerialScreen`: torch button rendered only when the camera reports a torch; camera no longer remounted on the initial focus (only on re-entry and Try again).
- `utils/camera`: `isRecoverableCameraRuntimeError` — iOS `unknown/unknown` only.
- `handleCameraError`: a first recoverable error logs a warning and lets VisionCamera restart the session; a repeat within 10 s, any other code, or Android fails over as before.
- Tests for all of the above.

UI: the torch button disappears on devices without a torch (non-Pro iPads) — attach a screenshot from one.

- `handleCameraError` in `CodeScanningCamera.tsx` — the strike window: check nothing else can starve escalation if the restart fails.
- The first-focus skip in `ScanSerialScreen.tsx` — a previous scan's lock/frozen frame must still reset on re-entry.
- Recoverable-error path is iOS only; Android error handling is untouched.

## How to test

```sh
cd app && yarn test src/bcsc-theme/components/utils/camera.test.ts src/bcsc-theme/components/CodeScanningCamera.test.tsx src/bcsc-theme/features/verify/ScanSerialScreen.test.tsx
```

- Non-Pro iPad: open the Scan Card screen — no torch button, scanning still works (before: tapping the torch showed "Camera Error").
- iPhone Pro: open the Scan Card screen, back out, re-enter — camera comes up each time without a Camera Error state.


al-rosenthal (approved):
tested this locally with my old iPad and scanner and cameras work as expected, no torch button visible

PR #4671, opened 2026-09-15
fix: device code recovery process

### LINKED ISSUE #4650: Error 2111 on Address screen — Invalid registration request
### What happened?

This comes from prod reporting of the 4.1.0 release, example report ID: 4PHQ-REY5

Not sure what the source of the bug is yet but testing back navigation and upgrade mid verification is a good place to start.

Successful registration on address screen in non-BCSC flow.

### Steps to reproduce

Unknown

### Where did you hit this?

Production

### How often does it happen?

Occasionally

### Device and OS version

iPhone 12 thru 18, possibly Android but none reported yet

### Impact

Blocks BCSC flow presumably

### Screenshots or log output

_No response_

### Build number

4.1.0-8333

al-rosenthal:
Closes #4650 

## What changed

2 major things have changed:
1. Users backing out of a QR transfer and starting a "normal" verification flow have the old device code/ registration cleaned up, locally and remotely (IAS)
2. A new recovery hook is used when a verification attempt fails in 3 particular spots due to an old device code being used. The recovery will capture the error, attempt recovery (get a new device code) and continue on. 

- updated error code in registry to be inline with actual error code (matched v3.15)
- added new hook `useDeviceAuthorizationRecovery`
- added new `cycleRegistration` function to remove and reregister device with IAS
- added new hook int `EnterBirthdateScreen`, `useResidentialAddressViewModel` and `useCardScanner`

the majority of the complicated work is done in these files:
- `useDeviceAuthorizationRecovery`
- `useRegistrationService`
- `AccountSetupScreen`

## How to test
- comment out the lines 65-69 in `AccountSetupScreen` 
- Start the qr transfer flow until you see the camera
- navigate back to the setup selection (this sets up the incorrect state)
- BCSC Card scan 
  - Scan a 2D bar code 
  - Auto scan will take a bit longer while it recovers and work as expected
  - Check logs for warn: [DeviceAuthorizationRecovery]
- Enter CSN manually
  - When entering birthdate see metro error appear
  - After some time the recovery completes workflow continues as expected
  - Check logs for warn: [DeviceAuthorizationRecovery]
- Begin Non BCSC flow
  - When submitting data on address screen, see metro error
  - Submission will take a bit longer than normal as recovery completes, workflow continues as expected
  - Check logs for warn: [DeviceAuthorizationRecovery]
- Uncomment out lines of code 
- Restart verification process
- Begin QR transfer process
- Navigate back to setup screen, there is a slight delay as device code clean up happens


app/src/bcsc-theme/services/hooks/useRegistrationService.tsx
```diff
@@ -128,13 +134,36 @@ export const useRegistrationService = () => {
...
+      const nativeToken = await getToken(TokenType.Registration)
+      registrationAccessToken = nativeToken?.token
+    }
+
+    if (registrationAccessToken && account?.clientID) {
+      await registrationApi.deleteRegistration(registrationAccessToken, account.clientID)
+    }
+
+    const securityMethod = await getAccountSecurityMethod()
+    await registrationApi.createRegistration(securityMethod)
```
jleach: I'm not sure if this is an overly edge case or not, but I though I'd point it out.

 If deletion succeeds but creating the replacement fails (for example, connectivity drops) the old client ID and registration token remain stored despite the server registration being gone. ensureRegistered() then skips registration because that token still exists. Another recovery attempt deletes the same missing registration first; a rejection prevents it from reaching creation again.

jleach (changes requested):
Removed my previous comment. I think that is an overly edge case.

jleach (approved):
Tested on Android. Started with transfer, backed out, continued with card setup, submitted a video. Working as expected.

PR #4672, opened 2026-09-15
fix: expired refresh token routes to account renewal instead of a generic error

### LINKED ISSUE #4654: Error 2500 — Problem with account, token failed validation
### What happened?

Reported in release 4.1.0-8333, no reproduction steps yet, but could be a race condition between unlock and read of unlocked data

No token validation issues

### Steps to reproduce

Unknown

### Where did you hit this?

Production

### How often does it happen?

Occasionally

### Device and OS version

iOS only so far, range of devices

### Impact

Affects a portion of users, unknown if it is blocking or can be dismissed with out issue

### Screenshots or log output

N/A

### Build number

4.1.0-8333

jleach:
Closes #4654

## What changed

Accounts whose 5-year device credential has expired now go straight to the existing "account expired, set it up again" screen instead of a repeated "Problem with Account" error. The app checks the stored expiry locally and skips the failing network call; if the server rejects a credential the device thought was valid, it no longer retries recovery.

The `isAppError(error, AppEventCode.INVALID_TOKEN)` gate in `useSecureActions.hydrateSecureState`. It must only stop retries for the server's explicit `invalid_token` response and leave the signing-key-mismatch recovery path (#4166) untouched.

## How to test

Unit tests cover the expiry util, API client, hydration, `useCardStatus`, and the new system check.

I'm not sure how you would test this other than making sure no regressions are introduced. The issue is triggered by the server rejecting an expired refresh token. I tested with:

 app/src/bcsc-theme/api/client.ts:370:

`this.tokensPromise = this.fetchTokens(refreshToken)`
to
`this.tokensPromise = Promise.reject(AppError.fromErrorDefinition(ErrorRegistry.INVALID_TOKEN)) // TEMP #4654`


KjartanE (approved):
Tested locally, with hotwire. works as intended

PR #4681, opened 2026-09-16
fix: page unavailable error 2835

### LINKED ISSUE #4663: Error 2835 — Page Unavailable
### What would you like to raise?

In the error report, we need to add the URL that the WebView is trying to load

KjartanE:
Closes #4663

## What changed

The in-app WebView no longer sends whatever access token happens to be cached. IAS account pages (Manage devices) wait for a valid token before loading and retry once with a forced refresh on a 401; public help and privacy pages get no bearer at all. `tel:` and `mailto:` links inside those pages now open the phone or mail app instead of failing in the WebView (error 2834).

- `BCSCApiClient.getAccessToken({ forceRefresh })`: public accessor over the existing refresh-before-expiry and forced-refresh paths.
- `WebViewContent`: token gating via `useDataLoader`, one-shot 401 refresh and reload, `onShouldStartLoadWithRequest` hand-off for non-web schemes.
- 2834 and 2835 problem reports now carry `context.url` (query stripped) and `context.statusCode`.
- HTTP error message drops the dangling separator when there is no description (`401`, not `401: `).

`WebViewContent.tsx`: `needsBearer` comes from `client.endpoints.account` / `accountDevices` prefixes, so a discovery change to those endpoints changes which pages get the bearer. The spinner render also covers a failed refresh (`accessTokenError`), so a rejected token is never resent. `originWhitelist` stays `['*']` on purpose: the library's own hand-off is gated on `canOpenURL`, which returns false for `tel`/`mailto` because Info.plist declares neither. No native changes.

## How to test

```bash
cd app
yarn test src/bcsc-theme/features/webview src/bcsc-theme/api/client.test.ts src/bcsc-theme/features/onboarding/TermsOfUseContent
```

On a physical device (simulators have no dialer or Mail), with a verified account: Settings → Contact Us → tap a phone number → the Call sheet appears and no "Page unavailable" modal. Settings → My devices → the page loads. For the stale-token path, leave the app in the background past the access-token lifetime (or raise `TOKEN_EXPIRY_BUFFER_MS` in a debug build) before opening My devices; a proxy (no TLS pinning) shows a `/device/token` refresh before the devices request and no `Authorization` header on help pages.


al-rosenthal (approved):

PR #4682, opened 2026-09-16
fix: guard account loads on verification, add load diagnostics

### LINKED ISSUE #4675: Error 2502 on 4.1.0: stored sign-in not found, report lacks the detail to diagnose it
## What happened

Users on 4.1.0 get "Something went wrong" (error 2502) because the app has an account but can't find the stored sign-in credential on the phone. It's new with 4.1.0: about 18 reports in the first two days, none on 4.0.2 over the same period with the same report volume.

The report doesn't say enough to tell what went wrong. No screen is recorded, and the one detail that did reach the logs says the app was in the background at the time.

## Two candidate causes

1. The 4.1.0 update lost the stored credential for some users (a real regression).
2. The phone was locked, and the app misread "locked" as "missing" — the phone reported "not found" (-25300), which our code only retries on "locked" (-25308).

Only (1) needs a "set up again" outcome. (2) should be a silent retry.

## What to do

Add the facts the app already knows at the moment of failure to the report: current screen, app state (foreground/background), and whether the phone's secure storage was available. They're logged locally today but that line almost never reaches remote logs (1 of ~19). This is a small change and will settle which cause it is within days of shipping.

While there, pass the server's reason through on the "Problem with Account" alerts (2500, 2501) — it's dropped today, so the non-expiry 2500s left after #4672 ships will be just as opaque.

## Evidence

Error code: `token.err_119_token_unexpectedly_null.2502`, "Token cache empty and no stored refresh token to recover from (nativeDiagnostic=-25300)". iOS only (15.x–27.0), 4.1.0-8333, Sep 14–16 2026. Sample install `b6902c5f…` hit it three times in 20 minutes.

Code: `app/src/bcsc-theme/api/client.ts` (`recoverTokens`), `packages/bcsc-core/ios/TokenService.swift` (`get`), `app/src/bcsc-theme/api/clientErrorPolicies.ts` (`invalidTokenReturnedPolicy`).

Related: #4645 (2515, keychain write failure), #4654 (2500, 5-year expiry).

jleach:
Closes #4675

## What changed

People who paused verification and returned Home could hit "Something went wrong" (error 2502) when their connection came back. The app no longer loads account details until sign-in is complete, and if a load still fails the report now records the trigger, screen, app state, connectivity and whether a sign-in token existed. The startup loading screen also overlays the app instead of replacing it, which is why the screen name was missing from reports. The "Problem with Account" alerts (2500, 2501) now keep the server's reason in the report too.

The single guard in `BCSCAccountContext.tsx` that every account load goes through, and the overlay change in `MainStack.tsx`. The `clientErrorPolicies.ts` change is a two-line pass-through.

## How to test

On the BCSC build: start verification, stop at "choose how to verify", go Help > Back to home, then toggle wifi off and on. Expect no error 2502 and a "Skipping account load for unverified user" log line. Unit tests cover unverified, verified and transfer-in-progress accounts, and the 2500/2501 pass-through.


bryce-mcmath (approved):
tested both main and your branch and was able to reproduce both the error and the lack of the error

PR #4688, opened 2026-09-17
fix: report keychain save failures and stop them blocking unlock

### LINKED ISSUE #4645: Unlock fails with Device Authentication Failed when a login token can't be saved to the keychain
## What happened

Unlock fails with **"Device Authentication Failed"** even though the biometric check succeeded. The failure happens straight after, when the app saves a login token back to the phone's secure storage.

Remote logs show this is widespread, not a one-off.

## Scale (last 30 days, remote logs)

- **300 reports**, all on 4.0.2-6887, roughly 10 a day
- Estimated **~165 distinct installs** (a one-week sample had 47 installs across 85 reports)
- **91% of all "Device Authentication Failed" reports** (300 of 328) are actually this, not a biometric problem
- **iOS and iPadOS only.** No Android equivalent in the logs
- Spread evenly across iOS 15.4 through 27.0 and every iPhone model — not tied to a particular OS or handset
- Hits both the Account Select and Device Authentication Prep screens

These counts are only users who hit the error *and* tapped "Report this problem", so the real number is higher.

## It doesn't always clear on its own

Most users retry within seconds and give up. But some come back much later and still can't get in: one install reported it, returned **4 days and 20 hours later**, and failed again. Others span 3 to 5 hours. For those users the app is effectively unusable until they remove and re-add their account.

## Leading theory — not confirmed

The app keeps login tokens in the iOS keychain. It looks like the app can't read one stored item, concludes it isn't there, tries to save a fresh copy, and the phone refuses because something is already in that slot. The failed save aborts the unlock.

**Confidence: moderate.** The pattern fits — it's persistent for some users, and it's independent of OS version and device. But the native code discards the underlying status code, so we can't yet tell "slot already taken" from "storage temporarily unavailable". We don't have a device to inspect.

Whether this is new in 4.0.2 is **unknown**: every occurrence is on 4.0.2-6887, but 4.0.1 has only 82 reports total in the retention window, far too few to compare.

## The message is misleading

Three layers relabel the error on the way up:

keychain save fails → storage **write** error (correct) → rewrapped as storage **read** error → rewrapped as **Device Authentication Failed**

So a storage problem reaches the user as a biometric problem, and support sees the wrong thing.

## Suggested work

1. **Log the real status code.** The iOS keychain call returns one; it's dropped and never reaches remote logging. Without it we can't confirm the theory. `main` already has a half-built helper for this.
2. **Recover automatically** — if the save fails because the slot is taken, clear it and write again.
3. **Don't fail unlock on this write.** The app re-saves tokens it just read on every unlock; a hiccup there shouldn't block a successful sign-in. This alone would clear most of the 300.
4. **Stop the relabelling** so a storage error doesn't surface as an authentication error.

Items 1 and 3 are small and independently useful. Item 2 is the actual fix if the theory holds.

## Evidence

Error chain: `device.device_authentication_error.2707` → `storage.err_101_failed_to_read_local_storage.2601` → `storage.err_100_failed_to_write_local_storage.2600` → `E_TOKEN_SAVE_FAILED` "Failed to save token to keychain"

Sample report: ID `2b6def3c-b07f-431f-8e60-4edb5c690f10`, 2026-09-11 21:48 UTC, iPhone 14 Pro Max (`iPhone15,3`), iOS 26.2, build 4.0.2-6887, screen Account Select, two failures 11 seconds apart.

LogQL used (via port-forwarded Loki read path):
```
sum by (system) (count_over_time({application=~".+"} |= "E_TOKEN_SAVE_FAILED" [30d]))
```

Relevant code (unchanged on `main`): `packages/bcsc-core/ios/TokenService.swift` (`save`, `get`), `packages/bcsc-core/ios/BcscCore.swift:526`, `app/src/bcsc-theme/hooks/useSecureActions.tsx` (`persistTokens`, `hydrateSecureState`), `app/src/bcsc-theme/hooks/useAuthentication.tsx:64`

jleach:
Closes #4645

Forward-port of #4647 to `main`, so 4.1.2 doesn't regress. Same four commits, cherry-picked with `-x`; no new work.

## What changed

1. **Keychain saves now report why they failed.** `save()` returns the `OSStatus` instead of a bare `Bool` and passes it to JS via `NSError.userInfo`.
2. **A failed token write no longer fails sign-in.** `hydrateSecureState` re-persists tokens it just read on every unlock; that write erroring used to abort an already-successful biometric unlock.
3. **Refreshed tokens survive that failure** — review on #4647 caught the store being re-populated from the original keychain values, dropping a rotated registration token.

Not carried over: the 4.1.1 version bump (N/A on `main`) and the swiftformat back-port (already here via #4608).

- `BcscCore.swift` auto-merged around the decrypt-key work that landed after the 4.1.0 tag. Worth confirming the `setToken` block matches the release branch.
- Still open in #4645: the write-back is redundant when nothing changed, and `performDeviceAuth` titles the modal "Device Authentication Failed" for any error escaping hydration.

## How to test

`useSecureActions.test.tsx` covers the JS half — 66/66 locally. The Swift half has no automated coverage (keychain tests skip under SPM for want of an entitlement); verified the package builds for the iOS simulator and lint/format are clean.

Real confirmation comes from the field: once this ships, `E_TOKEN_SAVE_FAILED` reports should carry an `osStatus` field — the number that turns the `errSecDuplicateItem` theory into a finding.


jleach:
@KjartanE This is a forward-port of #4647, which you reviewed on `release/v4.1.1` — same four commits onto `main`, so assigning to you.

KjartanE (approved):

bryce-mcmath (approved):

PR #4703, opened 2026-09-17
fix: add back more helpful image and text to card not found screen

### LINKED ISSUE #4701: Add back helpful Card not found text and image
### What would you like to raise?

If a user scans a non-BCSC that still has a serial and birthdate, 3.15, 4.0.2 and 4.1.0 all take them to the Card not found screen, but 4.1.0 is missing a helpful image and text from earlier versions that says: 

"Check your card is a BC Services Card by looking at the front of your card:" <image showing correct and incorrect cards>

We should add that image and text back to help prevent confusion for users

bryce-mcmath:
Closes #4701

## What changed

Before and after:
before
after

Just the content and small amount of code changes

## How to test

Enter an incorrect birthdate + serial combo


al-rosenthal (approved):

PR #4707, opened 2026-09-18
fix: expired refresh token routes to account renewal instead of a gen…

### LINKED ISSUE #4654: Error 2500 — Problem with account, token failed validation

bryce-mcmath:
Closes #4654

## What changed

Cherry-pick of https://github.com/bcgov/bc-wallet-mobile/pull/4672

Same reviewer, same change

## How to test

Same test strategy


jleach (approved):
Fast tracking since it's a back-port from main.

PR #4723, opened 2026-09-18
feat(bcsc): validate service responses and persisted state

### LINKED ISSUE #3581: Test.
x

jleach:
Closes #3581

## What changed

The app now checks the shape of three BC Services Card service responses — registration, sign-in tokens, and user info — before using or saving them. It also checks its own saved state on the phone at startup: a field in the wrong shape is reset to a safe default for just that field, and the app always starts.

When a registration or user-info response is shaped wrong, the user sees the existing "something went wrong" message instead of the app quietly continuing with bad data. This includes the automatic registration refresh that runs at app launch. A wrong-shaped sign-in token response is logged and counted but shows no message in this stage.

Logs record only which field was wrong and why — never the value — so no personal data or tokens are exposed.

- The user-info error path shows the message only for user-info problems; a token refresh failing the same check during that call stays silent by design.
- At startup, a corrupted saved field is reset after the one-time migrations run, so a genuinely missing field still migrates while a corrupted one is not misread as a legacy value.

## How to test

`cd app && TZ=GMT yarn test response-schemas persisted-bcsc-state container-imp client useUserService useRegistrationService useSecureActions`

On-device smoke test of a fresh launch is a manual follow-up.


jleach:
Just an experament.

PR #4724, opened 2026-09-18
feat: release notes workflow

### LINKED ISSUE #4464: Changelog generation

al-rosenthal:
Closes #4464

## What changed

- added a new `release-notes.yml` workflow
- workflow takes two tags and finds all the PRs merged in between the tags
- it then collects any linked tickets (via `Closes ###` convention)
- it will organize those tickets based on label (Feature, Bug, etc.)
- mark down tables are created and posted to a discussion under the "Release notes" label

- `release-notes.yml`, this contains all the work you'd need to focus on

## How to test

The PR will need to be put in before we can see it working, an example of the output can be found [here](https://github.com/bcgov/bc-wallet-mobile/discussions/4717)


.github/workflows/release-notes.yml
MacDeluca, line 1: suggestion: The inline scripts in this action are fairly large. Could we move them into .github/scripts so we get IntelliSense while editing them?

